PR #1, opened 2021-02-08
add ts-web

pro-wh:
Moved from https://github.com/oasisprotocol/oasis-bridge/pull/3, which had thus far been non-bridge-specific material


pro-wh:
I'm going to move the secp256k1 stuff to #3

pro-wh:
reviewers please review

client-sdk/ts-web-ledger/package.json
```diff
@@ -0,0 +1,23 @@
+{
+    "name": "@oasisprotocol/client-ledger",
```
kostko, line 2: Sorry for nitpicking over the name, but I think it may be more clear if it was named `@oasisprotocol/client-signer-ledger` or `@oasisprotocol/signer-ledger`.

client-sdk/ts-web-ledger/playground/dist/index.html
```diff
@@ -0,0 +1,4 @@
+<!doctype html>
```
kostko, line 1: Should this playground be integrated as some kind of a test case? Probably not in this PR, as we need to add tests all around the SDK anyway.

client-sdk/ts-web/playground/sample-envoy.yaml
```diff
@@ -0,0 +1,74 @@
...
...Scheduler/(GetValidators|GetCommittees|StateToGenesis|WatchCommittees)|Registry/(GetEntity|GetEntities|GetNode|GetNodeByConsensusAddress|GetNodeStatus|GetNodes|GetRuntime|GetRuntimes|StateToGenesis|GetEvents|WatchEntities|WatchNodes|WatchNodeList|WatchRuntimes)|Staking/(TokenSymbol|TokenValueExponent|TotalSupply|CommonPool|LastBlockFees|Threshold|Addresses|Account|Delegations|DebondingDelegations|StateToGenesis|ConsensusParameters|GetEvents|WatchEvents)|KeyManager/(GetStatus|GetStatuses)|Storage/(SyncGet|SyncGetPrefixes|SyncIterate|Apply|ApplyBatch|GetCheckpoints|GetDiff|GetCheckpointChunk)|RuntimeClient/(SubmitTx|GetGenesisBlock|GetBlock|GetBlockByHash|GetTx|GetTxByBlockHash|GetTxs|QueryTx|QueryTxs|WaitBlockIndexed|WatchBlocks)|EnclaveRPC/(CallEnclave)|Consensus/(SubmitTx|StateToGenesis|EstimateGas|GetSignerNonce|GetEpoch|WaitEpoch|GetBlock|GetTransactions|GetTransactionsWithResults|GetUnconfirmedTransactions|GetGenesisDocument|GetStatus|WatchBlocks)|ConsensusLight/(GetLightBlock|GetParameters|StateSyncGet|StateSyncGetPrefixes|StateSyncIterate|SubmitTxNoWait|SubmitEvidence)|NodeController/(RequestShutdown|WaitSync|IsSynced|WaitReady|IsReady|UpgradeBinary|CancelUpgrade|GetStatus))'
```
kostko, line 25: Ouch :facepalm:

client-sdk/ts-web/src/signature.ts
```diff
@@ -0,0 +1,134 @@
...
+
+export function combineChainContext(context: string, chainContext: string) {
+    return `${context} for chain ${chainContext}`;
+}
+
+export async function prepareSignerMessage(context: string, message: Uint8Array) {
+    return await hash.hash(misc.concat(misc.fromString(context), message));
+}
+
+export interface Signer {
```
kostko, line 15: Why do we have a signer without domain separation contexts? We should make sure that domain separation contexts are used everywhere. Just rename `ContextSigner` to `Signer`.

kostko (commented):
Just some quick comments, will resume with review later.

client-sdk/ts-web-ledger/package.json
pro-wh: sure let's go with client-signer-ledger. it'll be good for making sure the scope doesn't creep to include other non-signing functionality

client-sdk/ts-web-ledger/playground/dist/index.html
pro-wh: should be. it'll be more feasible after I switch over to oasis-net-runner instead of running on testnet

client-sdk/ts-web/playground/sample-envoy.yaml
pro-wh: it was this or replicate the whole `{match: ..., route: ...}` for each method :facepalm: 
or perhaps per module, if we're willing to whitelist at that granularity

client-sdk/ts-web/src/signature.ts
pro-wh: blind no-context signing might be an appealing to devs. I've seen another interface design, Rosetta, expose signing this way https://www.rosetta-api.org/docs/construction_api_introduction.html#detached-key-generation-and-transaction-signing. it might be valuable to have this and the BlindSigner helper for the time being that our context signing construction is only one idea in a broader tooling ecosystem.

pro-wh:
cross reference: https://github.com/grpc/grpc-web/issues/1043

let's get this out and start working on updates for the next version of oasis-core

client-sdk/ts-web/src/common.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export const CODE_NO_ERROR = 0;
+
+/*
+Regular_expression('User defined','Err\\w+\\s*=\\s*errors\\.New\\(\\w+, \\d+, ".*"\\)',true,true,false,false,false,false,'List matches')
+Find_/_Replace({'option':'Regex','string':'Err(\\w+)\\s*=\\s*errors\\.New\\(\\w+, (\\d+), ".*"\\)'},'$1 = $2;',true,false,true,false)
+Find_/_Replace({'option':'Regex','string':'[A-Z]'},'_$&',true,false,true,false)
+To_Upper_case('All')
+Find_/_Replace({'option':'Regex','string':'^_'},'export const CODE_',true,false,true,false)
+*/
```
kostko, line 28: Remove?

kostko (approved):
Some general comments:
* Could we move all the ts-web stuff under `client-sdk/ts-web/` instead of having multiple of these directly under `client-sdk`?
* Could we add some kind of documentation comments for all the things? Not sure what is the preferred way.

Otherwise let's merge this initial version and we can improve incrementally.

client-sdk/ts-web/src/common.ts
pro-wh: I'll move these scripts to a separate file. They're really without context here.

pro-wh:
>     * Could we move all the ts-web stuff under `client-sdk/ts-web/` instead of having multiple of these directly under `client-sdk`?

ok I'll try. client-sdk/ts-web/{core,signer-ledger}

>     * Could we add some kind of documentation comments for all the things? Not sure what is the preferred way.

manually :laughing: I could try transferring the docstrings over from go

PR #3, opened 2021-02-11
add ts-web/rt

pro-wh:
utilities for this runtime sdk

### DIFF
--- a/client-sdk/ts-web/core/src/runtime.ts
+++ b/client-sdk/ts-web/core/src/runtime.ts
@@ -1,4 +1,9 @@
 /**
+ * RoundLatest is a special round number always referring to the latest round.
+ */
+export const ROUND_LATEST = 0xffffffffffffffffn;
+
+/**
  * ModuleName is the runtime client module name.
  */
 export const CLIENT_MODULE_NAME = 'runtime/client';
--- a/client-sdk/ts-web/core/src/runtime.ts
+++ b/client-sdk/ts-web/core/src/runtime.ts
@@ -1,40 +1,42 @@
 /**
  * RoundLatest is a special round number always referring to the latest round.
  */
-export const ROUND_LATEST = 0xffffffffffffffffn;
+export const CLIENT_ROUND_LATEST = 0xffffffffffffffffn;
 
 /**
  * ModuleName is the runtime client module name.
  */
 export const CLIENT_MODULE_NAME = 'runtime/client';
+
 /**
  * ErrNotFound is an error returned when the item is not found.
  */
-export const CODE_NOT_FOUND = 1;
+export const CLIENT_ERR_NOT_FOUND_CODE = 1;
 /**
  * ErrInternal is an error returned when an unspecified internal error occurs.
  */
-export const CODE_INTERNAL = 2;
+export const CLIENT_ERR_INTERNAL_CODE = 2;
 /**
  * ErrTransactionExpired is an error returned when transaction expired.
  */
-export const CODE_TRANSACTION_EXPIRED = 3;
+export const CLIENT_ERR_TRANSACTION_EXPIRED_CODE = 3;
 /**
  * ErrNotSynced is an error returned if transaction is submitted before node has finished
  * initial syncing.
  */
-export const CODE_NOT_SYNCED = 4;
+export const CLIENT_ERR_NOT_SYNCED_CODE = 4;
 /**
  * ErrCheckTxFailed is an error returned if the local transaction check fails.
  */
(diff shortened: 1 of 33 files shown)


client-sdk/ts-web-rt/package.json
```diff
@@ -0,0 +1,24 @@
...
+    "devDependencies": {
+        "@types/elliptic": "^6.4.12",
+        "ts-loader": "^8.0.16",
+        "typescript": "^4.1.3",
+        "webpack": "^5.21.2",
+        "webpack-cli": "^4.5.0",
+        "webpack-dev-server": "^3.11.1"
+    }
+  }
+
```
pro-wh, line 24: what in the

pro-wh:
ready for review

pro-wh:
back to draft, in spirit. I'll be adding helpers for the built-in modules Core and Accounts. and making changes to the signing API

client-sdk/ts-web/rt/src/accounts.ts
```diff
@@ -0,0 +1,9 @@
+export const MODULE_NAME = 'acccounts';
```
kostko, line 1:
```suggestion
export const MODULE_NAME = 'accounts';
```

client-sdk/ts-web/rt/src/accounts.ts
```diff
@@ -0,0 +1,9 @@
...
+
+export const CODE_INVALID_ARGUMENT = 1;
+export const CODE_INSUFFICIENT_BALANCE = 2;
+export const CODE_FORBIDDEN = 3;
+
+export const METHOD_TRANSFER = 'accounts.Transfer';
+export const METHOD_NONCE = 'accounts.Nonce';
+export const METHOD_BALANCES = 'accounts.Balances';
```
kostko, line 9: This should probably differentiate between what are callable methods (e.g. that can be invoked via transactions) and what are query methods.

pro-wh: yeah, maybe we can add this along with the request/response typed wrappers

pro-wh:
I got this to work using the playground page in oasis-bridge today, so back to review.

so far no typed call/query wrappers yet though

client-sdk/ts-web/rt/src/accounts.ts
pro-wh: also added separator comments for now :shrug:

client-sdk/ts-web/rt/playground/src/index.js
```diff
@@ -0,0 +1,22 @@
+// @ts-check
+
+import * as oasisRT from './../..';
+
+(async function () {
+    try {
```
pro-wh, line 6: There ends up not being much here, because we went to focus on getting the bridge SDK working. Maybe we could put up a simple-keyvalue client here someday.

client-sdk/ts-web/core/src/runtime.ts
```diff
@@ -1,3 +1,5 @@
+export const ROUND_LATEST = 0xffffffffffffffffn;
```
kostko, line 1: No documentation comment for this one?

client-sdk/ts-web/rt/playground/src/index.js
kostko: Yeah once we have an example "simple-keyvalue" runtime that uses the runtime-sdk.

client-sdk/ts-web/rt/src/accounts.ts
```diff
@@ -0,0 +1,18 @@
...
+
+// Callable methods.
+export const METHOD_TRANSFER = 'accounts.Transfer';
+// Queries.
+export const METHOD_NONCE = 'accounts.Nonce';
+export const METHOD_BALANCES = 'accounts.Balances';
+
+export const CODE_TRANSFER_HEX = '00000001';
+export const CODE_BURN_HEX = '00000002';
+export const CODE_MINT_HEX = '00000003';
```
kostko, line 18: What are these, events? Maybe call them `EVENT_*`? Also, why in hex, they could be integers?

client-sdk/ts-web/core/src/runtime.ts
pro-wh: oops, I'll transfer it

client-sdk/ts-web/rt/src/accounts.ts
pro-wh: they're events. ok to the suggestions

kostko (approved):

pro-wh:
thanks, proceeding toward merge!

PR #4, opened 2021-02-19
ts-web/core: transfer docstrings

pro-wh:
transferring docstrings from go codebase

also some corrections and minor additions that I noticed when I went through to do that

### DIFF
--- a/client-sdk/ts-web/core/src/common.ts
+++ b/client-sdk/ts-web/core/src/common.ts
@@ -2,27 +2,72 @@ import * as misc from './misc';
 import * as signature from './signature';
 import * as types from './types';
 
+/**
+ * RoleComputeWorker is the compute worker role.
+ */
 export const ROLE_COMPUTE_WORKER = 1 << 0;
+/**
+ * RoleStorageWorker is the storage worker role.
+ */
 export const ROLE_STORAGE_WORKER = 1 << 1;
+/**
+ * RoleKeyManager is the the key manager role.
+ */
 export const ROLE_KEY_MANAGER = 1 << 2;
+/**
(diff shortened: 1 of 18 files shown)


client-sdk/ts-web/core/src/common.ts
```diff
@@ -2,27 +2,72 @@ import * as misc from './misc';
...
 export const CODE_UNKNOWN_ERROR = 1;
 
 export const IDENTITY_MODULE_NAME = 'identity';
+/**
+ * ErrCertificateRotationForbidden is returned by RotateCertificates if
+ * TLS certificate rotation is forbidden.  This happens when rotation is
+ * enabled and an existing TLS certificate was successfully loaded
+ * (or a new one was generated and persisted to disk).
+ */
 export const CODE_CERTIFICATE_ROTATION_FORBIDDEN = 1;
```
kostko, line 71: The identity module is not exposed to clients so it probably doesn't make much sense to transfer it over?

kostko (approved):
I see this is based on 20.12? Can we update it for latest master (in a separate PR) where some things are quite different?

client-sdk/ts-web/core/src/common.ts
pro-wh: oh I had searched for all calls to `errors.New`. let me open an issue for tracking this and related removals

pro-wh: #8

pro-wh:
> I see this is based on 20.12? Can we update it for latest master (in a separate PR) where some things are quite different?

yup. I've opened #9 to track this

PR #6, opened 2021-02-20
add license

pro-wh:
note that the license includes a template at the bottom. don't fill in the fields in that copy of the template, because it's meant to be a template. we have incorrectly done so in our other repos, but don't do it here.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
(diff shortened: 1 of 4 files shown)


kostko (approved):
Can we just rename it to `LICENSE` (without the extension)? Just to be consistent with other repositories.

kostko:
> note that the license includes a template at the bottom. don't fill in the fields in that copy of the template, because it's meant to be a template. we have incorrectly done so in our other repos, but don't do it here.

Ouch, we should fix that in other repositories.

pro-wh:
oh and I hadn't added the license metadata to the other packages in here. I'll add those too

PR #10, opened 2021-02-23
pull in changes from oasis-core master

### LINKED ISSUE #9: Update ts-web for oasis-core master branch
tasks breakdown forthcoming

but we'll be pursuing this mainly for its expanded runtime interfaces

pro-wh:
fixes #9 

This brings ts-web/core to oasis-core rev e000dfaf06a1e3a58f1eb3ca99c24922fac7aea5

### DIFF
--- a/client-sdk/ts-web/core/src/beacon.ts
+++ b/client-sdk/ts-web/core/src/beacon.ts
@@ -1,4 +1,25 @@
 /**
+ * MethodPVSSCommit is the method name for a PVSS commitment.
+ */
+export const METHOD_PVSS_COMMIT = 'beacon.PVSSCommit';
+/**
+ * MethodPVSSReveal is the method name for a PVSS reveal.
+ */
+export const METHOD_PVSS_REVEAL = 'beacon.PVSSReveal';
+/**
+ * MethodSetEpoch is the method name for setting epochs.
+ */
+export const METHOD_SET_EPOCH = '000_beacon.SetEpoch';
+
+/**
+ * BackendInsecure is the name of the insecure backend.
+ */
+export const BACKEND_INSECURE = 'insecure';
+/**
+ * BackendPVSS is the name of the PVSS backend.
+ */
+export const BACKEND_PVSS = 'pvss';
+/**
  * ModuleName is a unique module name for the beacon module.
  */
 export const MODULE_NAME = 'beacon';
--- a/client-sdk/ts-web/core/src/consensus.ts
+++ b/client-sdk/ts-web/core/src/consensus.ts
@@ -59,6 +59,10 @@ export const CODE_UNSUPPORTED = 4;
  * ErrDuplicateTx is the error returned when the transaction already exists in the mempool.
  */
 export const CODE_DUPLICATE_TX = 5;
+/**
+ * ErrInvalidArgument is the error returned when the request contains an invalid argument.
+ */
+export const CODE_INVALID_ARGUMENT = 6;
 
 /**
  * moduleName is the module name used for error definitions.
--- a/client-sdk/ts-web/core/src/types.ts
+++ b/client-sdk/ts-web/core/src/types.ts
@@ -88,263 +88,9 @@ export interface BeaconPVSSReveal {
 }
 
 /**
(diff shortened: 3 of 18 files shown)


client-sdk/ts-web/core/src/beacon.ts
```diff
@@ -1,3 +1,24 @@
...
+ */
+export const METHOD_PVSS_COMMIT = 'beacon.PVSSCommit';
+/**
+ * MethodPVSSReveal is the method name for a PVSS reveal.
+ */
+export const METHOD_PVSS_REVEAL = 'beacon.PVSSReveal';
+/**
+ * MethodSetEpoch is the method name for setting epochs.
+ */
+export const METHOD_SET_EPOCH = '000_beacon.SetEpoch';
```
kostko, line 12: Not sure if these need to be exposed in the client SDK as the first two can only be called by a validator and the last one can only be used in debug mode.

client-sdk/ts-web/core/src/consensus.ts
```diff
@@ -78,6 +82,8 @@ export const CODE_INSUFFICIENT_FEE_BALANCE = 2;
  */
 export const CODE_GAS_PRICE_TOO_LOW = 3;
 
+// %%% method priority?
```
kostko, line 85: I don't think method priority needs to be exposed in the client SDK as there is nothing there that the clients can do specifically.

client-sdk/ts-web/core/src/types.ts
```diff
@@ -1208,6 +1478,12 @@ export interface RegistryUnfreezeNode {
     node_id: Uint8Array;
 }
 
+/**
+ * ValidatorSetConstraint specifies that the entity must have a node that is part of the validator
+ * set. No other options can currently be specified.
+ */
+export type RegistryValidatorSetConstraint = Map<never, never>;
```
kostko, line 1485: The constraints types should probably be grouped together? Now it seems that some are 300 lines higher.

kostko: Also why does it have this `Registry` prefix that the original type does not have?

client-sdk/ts-web/core/src/beacon.ts
pro-wh: I'm fine with exposing methods used in debug mode.

client-sdk/ts-web/core/src/consensus.ts
pro-wh: yeah agreed

client-sdk/ts-web/core/src/types.ts
pro-wh: I went with alphabetical sorting in types.ts ):

these types almost all have a prefix for what Go package they're from

pro-wh:
oh right I was gonna check how the kyber stuff ended up serializing

client-sdk/ts-web/core/src/types.ts
```diff
@@ -96,6 +160,16 @@ export interface CommonConsensusInfo {
     addresses: CommonConsensusAddress[];
 }
 
+/**
+ * Constraints are the Intel SGX TEE constraints.
+ */
+export interface CommonConstraints {
```
kostko, line 166: This is originally under `sgx.Constraints` so maybe `SGXConstraints`?

pro-wh: ya there's definitely some weirdness with this being a flat namespace and the Go codebase being hierarchical. currently there's a `Common` type name prefix and a `common.ts` with almost everything from there mixed together: node, entity, sgx, etc. these are ones that are mostly _type definitions_ and _constants_. the exceptions are modules that are mostly _helpers_: common/cbor -> `misc.ts` and common/crypto/... -> their own file and prefix, such as `signature.ts` and `hash.ts`. PVSS is on the fence.

pro-wh: let's change `CommonConstraints` to `SGXConstraints` and similar. `CommonNode` will be `Node`. `SignatureSignature` will become `Signature`. but keep helpers and constants in `common.ts`.

kostko (approved):

PR #14, opened 2021-02-24
runtime-sdk: Add fee payment and disbursement

kostko:
TODO
* [x] Use last round committee metadata to inform fee disbursement (requires support in Oasis Core)

### DIFF
--- a/runtime-sdk/src/module.rs
+++ b/runtime-sdk/src/module.rs
@@ -151,6 +151,22 @@ impl MigrationHandler for Tuple {
     }
 }
 
+/// Block handler.
+#[impl_for_tuples(30)]
+pub trait BlockHandler {
+    /// Perform any common actions at the start of the block (before any transactions have been
+    /// executed).
+    fn begin_block(_ctx: &mut DispatchContext) {
+        // Default implementation doesn't do anything.
+    }
+
+    /// Perform any common actions at the end of the block (after all transactions have been
+    /// executed).
+    fn end_block(_ctx: &mut DispatchContext) {
+        // Default implementation doesn't do anything.
+    }
+}
+
 /// A runtime module.
 pub trait Module {
     /// Module name.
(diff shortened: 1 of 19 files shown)


runtime-sdk/src/modules/accounts/mod.rs
```diff
@@ -481,18 +561,102 @@ impl module::AuthHandler for Module {
...
+        if !addrs.is_empty() {
+            let amounts: Vec<_> = previous_fees
+                .iter()
+                .map(|(denom, fee)| {
+                    token::BaseUnits::new(
+                        fee.checked_div(&token::Quantity::from(addrs.len() as u64))
+                            .expect("addrs is non-empty"),
+                        denom.clone(),
+                    )
+                })
```
ptrus, line 623: Could we also filter out zero amounts here, to avoid unnecessary adds/subs bellow?

kostko: Good idea, will change it.

kostko: Done.

runtime-sdk/src/module.rs
```diff
@@ -151,6 +151,22 @@ impl MigrationHandler for Tuple {
     }
 }
 
+/// Block handler.
+#[impl_for_tuples(30)]
```
ptrus, line 155: I see we use this on most traits that support automatic implementations (methods with no return values, no associated types). Do we actually need/use any of these in tuples anywhere (or plan to in future)? Or are we just doing this as a "convention" where we can?

kostko: Oh yes, this is a neat feature that allows us to do things like:
https://github.com/oasisprotocol/oasis-bridge/blob/91db3f59a73611f482b86d12890ed8efde4d75b2/runtime/src/lib.rs#L14-L17

And then call all these methods on the `Runtime::Modules` type (which basically combines multiple modules).

ptrus: Ah yeah that's nice :+1: i was actually looking for a runtime implementation but didn't found it in this repo - didn't think about looking over at bridge repo :grin:

ptrus (approved):

PR #16, opened 2021-02-25
explode common

pro-wh:
this is a draft to see what it would be like.

separate them to have their own prefixes/modules based on the Go codebase's detailed breakdown of the 'common' directory.
I kinda don't like it. but it may be more sustainable if we end up having types with the same name in different 'common' subpackages.

also tries out removing duplicate parts of names, e.g. `CommonNode` -> `NodeNode` -> `Node`.
I kinda don't like that either. seems like we had done this in the Go codebase with things like `consensus.Parameters` instead of `consensus.ConsensusParameters`. we had a lint for it, formerly known as 'stutter' https://github.com/rust-lang/rust-clippy/issues/3521. I personally would have preferred rule simplicity and consistency across packages and longer name (`consensus.ConsensusParameters`) over the lint rule's more complex rule, less consistency across packages, and shorter name (`consensus.Parameters`). viva `HTMLHtmlElement` too

oh and this causes us to have a `cbor` module while our actual cbor helpers are in `misc`. hm

### DIFF
--- a/client-sdk/ts-web/core/src/common.ts
+++ b/client-sdk/ts-web/core/src/common.ts
@@ -71,17 +71,17 @@ export const IDENTITY_MODULE_NAME = 'identity';
 export const CODE_CERTIFICATE_ROTATION_FORBIDDEN = 1;
 
 export async function openSignedEntity(context: string, signed: types.SignatureSigned) {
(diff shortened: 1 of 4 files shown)


pro-wh:
let's go ahead with the type name changes, but not the module changes.

PR #19, opened 2021-03-03
ts-web/core: pull in changes from oasis-core master

pro-wh:
this brings us up to https://github.com/oasisprotocol/oasis-core/commit/d186644faee8ab99cc1e94df4bdcf04a92873e00

on top of #40


pro-wh:
uh oh this is rotting a lot

pro-wh:
updated, rolled it back a little to match the rust dependency. will put the rest into a separate PR for after we get the tests passing. ready for review

PR #26, opened 2021-03-06
Remove assert-matches dependency

nhynes:


kostko (approved):
Oh that's nice, was looking for it but didn't find it :-)

PR #27, opened 2021-03-06
Use WebCrypto for hashing

nhynes:
`crypto.subtle` is both faster and better maintained

### DIFF
--- a/client-sdk/ts-web/core/src/hash.ts
+++ b/client-sdk/ts-web/core/src/hash.ts
@@ -1,5 +1,9 @@
-import {sha512_256} from 'js-sha512';
-
+/** Returns the SHA-512/256 hash of `data`. */
 export async function hash(data: Uint8Array) {
-    return new Uint8Array(sha512_256.arrayBuffer(data));
+    const dataBuf = data.buffer.slice(
+        data.byteOffset,
+        data.byteOffset + data.byteLength
+    );
+    const hashBuf = await crypto.subtle.digest("SHA-512", dataBuf);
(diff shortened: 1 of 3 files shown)


client-sdk/ts-web/core/src/hash.ts
```diff
@@ -1,5 +1,9 @@
-import {sha512_256} from 'js-sha512';
-
+/** Returns the SHA-512/256 hash of `data`. */
 export async function hash(data: Uint8Array) {
-    return new Uint8Array(sha512_256.arrayBuffer(data));
+    const dataBuf = data.buffer.slice(
+        data.byteOffset,
+        data.byteOffset + data.byteLength
+    );
+    const hashBuf = await crypto.subtle.digest("SHA-512", dataBuf);
```
kostko, line 7: This is not right, it should be SHA-512/256 not SHA-512.

kostko (changes requested):

client-sdk/ts-web/core/src/hash.ts
nhynes: SHA-512/256 is SHA-512 truncated to 256 bits, used because 64-bit 512 is more efficient on modern processors [[ref](https://eprint.iacr.org/2010/548.pdf))

nhynes: gets me every time

pro-wh: ya, I wish we could use built in stuff. all this stuff is declared `async` in hope that one day we can

PR #28, opened 2021-03-06
Lint ts-web client using prettier

nhynes:
The prettierrc is intended to codify the pre-existing style


kostko (commented):
Would be cool if you could update CI to enforce this.

kostko (approved):

Issue #22, opened 2021-03-04
Add derive macro for easier error derivation


kostko:
In addition to what the procedural macro for `thiserror::Error` does, it should also implement the SDK's `Error` trait (`module` and `code`), like this (where module name and error codes would be specified via struct/variant attributes):
```rust
impl error::Error for Error {
    fn module(&self) -> &str {
        MODULE_NAME
    }

    fn code(&self) -> u32 {
        match self {
            Error::InvalidArgument => 1,
            Error::InsufficientBalance => 2,
            Error::Forbidden => 3,
        }
    }
}
```

And a conversion to `error::RuntimeError` like this:
```rust
impl From<Error> for error::RuntimeError {
    fn from(err: Error) -> error::RuntimeError {
        error::RuntimeError::new(err.module(), err.code(), &err.msg())
    }
}
```

PR #29, opened 2021-03-08
Add version_from_cargo proc macro

nhynes:
this PR adds `version_from_cargo` proc macro because in the one in `oasis_core::runtime::common` uses `parse`, which isn't a `const fn`.

### DIFF
--- a/runtime-sdk-macros/Cargo.toml
+++ b/runtime-sdk-macros/Cargo.toml
@@ -16,5 +16,4 @@ syn = "1.0.62"
 
 [dev-dependencies]
 diff = "0.1.12"
-rustfmt = "0.10"
 syn = { version = "1.0.62", features = ["extra-traits"] }
--- a/runtime-sdk-macros/Cargo.toml
+++ b/runtime-sdk-macros/Cargo.toml
@@ -9,8 +9,12 @@ license = "Apache-2.0"
 proc-macro = true
 
 [dependencies]
+darling = "0.12.2"
 proc-macro2 = "1.0.24"
 quote = "1.0.9"
+syn = "1.0.62"
 
 [dev-dependencies]
+diff = "0.1.12"
+rustfmt = "0.10"
 syn = { version = "1.0.62", features = ["extra-traits"] }
--- a/runtime-sdk-macros/Cargo.toml
+++ b/runtime-sdk-macros/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "oasis-runtime-sdk-macros"
+version = "0.1.0"
+authors = ["Oasis Protocol Foundation <info@oasisprotocol.org>"]
+edition = "2018"
+license = "Apache-2.0"
+
+[lib]
+proc-macro = true
+
+[dependencies]
+proc-macro2 = "1.0.24"
+quote = "1.0.9"
+
+[dev-dependencies]
+syn = { version = "1.0.62", features = ["extra-traits"] }
--- a/runtime-sdk-macros/src/event_derive.rs
+++ b/runtime-sdk-macros/src/event_derive.rs
@@ -1,0 +1,154 @@
+use darling::{util::Flag, FromDeriveInput, FromVariant};
+use proc_macro2::TokenStream;
+use quote::{format_ident, quote};
+use syn::{DeriveInput, Ident};
+
+#[derive(FromDeriveInput)]
+#[darling(supports(enum_any), attributes(event))]
+struct Event {
+    ident: Ident,
+
+    data: darling::ast::Data<EventVariant, darling::util::Ignored>,
+
+    /// The type ident of the event enum.
+    module: syn::Path,
+
+    /// Whether to sequentially autonumber the event codes.
+    /// This option exists as a convenience for runtimes that
+    /// only append events or release only breaking changes.
+    #[darling(default)]
+    autonumber: Flag,
+}
+
+#[derive(FromVariant)]
+#[darling(attributes(event))]
+struct EventVariant {
+    ident: Ident,
+
+    /// The explicit ID of the event code. Overrides any autonumber set on the event enum.
+    #[darling(default)]
+    id: Option<u32>,
+}
+
+pub fn derive_event(input: DeriveInput) -> TokenStream {
+    let event = match Event::from_derive_input(&input) {
+        Ok(event) => event,
+        Err(e) => return e.write_errors(),
+    };
+
+    let event_ty_ident = &event.ident;
+    let wrapper_ident = format_ident!("_IMPL_EVENT_FOR_{}", event_ty_ident);
+    let module_path = &event.module;
+
+    let variants = event.data.as_ref().take_enum().unwrap();
+    let mut next_autonumber = 0u32;
+    let mut reserved_numbers = std::collections::BTreeSet::new();
+    let code_match_arms = variants.iter().map(|variant| {
+        let event_id = match variant.id {
+            Some(id) => {
+                if reserved_numbers.contains(&id) {
+                    variant
+                        .ident
+                        .span()
+                        .unwrap()
+                        .error(format!("id {} already used", id))
+                        .emit();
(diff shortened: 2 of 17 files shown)


runtime-sdk-macros/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "runtime-sdk-macros"
```
kostko, line 2:
```suggestion
name = "oasis-runtime-sdk-macros"
```

runtime-sdk/Cargo.toml
```diff
@@ -8,6 +8,7 @@ license = "Apache-2.0"
 [dependencies]
 # TODO: Use specific tag.
 oasis-core-runtime = { git = "https://github.com/oasisprotocol/oasis-core", branch = "master" }
+runtime-sdk-macros = { path = "../runtime-sdk-macros", optional = true }
```
kostko, line 11: This should probably be on by default?

kostko (commented):
Nice! Left some minor comments.

runtime-sdk/Cargo.toml
nhynes: That makes sense. I was thinking that maybe people don't want to pull in `syn` for a little macro, but I guess they already are!

runtime-sdk-macros/src/event_derive.rs
```diff
@@ -0,0 +1,154 @@
...
+
+#[derive(FromDeriveInput)]
+#[darling(supports(enum_any), attributes(event))]
+struct Event {
+    ident: Ident,
+
+    data: darling::ast::Data<EventVariant, darling::util::Ignored>,
+
+    /// The type ident of the event enum.
+    module: syn::Path,
```
nhynes, line 14: it'd be nice to remove this using type information. not a huge fan of getting it, though

kostko (approved):

kostko (commented):
Oh, maybe you could also change the existing `modules/accounts` to use the new event macro?

nhynes:
also resolve https://github.com/oasisprotocol/oasis-sdk/issues/22 now

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -11,38 +11,31 @@ pub mod types;
 /// Unique module name.
 pub const MODULE_NAME: &str = "core";
 
-// TODO: Add a custom derive macro for easier error derivation (module/error codes).
 /// Errors emitted by the core module.
-#[derive(Error, Debug)]
+#[derive(Error, Debug, oasis_runtime_sdk_macros::Error)]
+#[runtime_error(module_name_path = "MODULE_NAME")]
```
kostko, line 16: Could it be `sdk::error` or something? Also this seems slightly inconsistent with how it's done for events? Could we use the same syntax for both.

nhynes: Unfortunately it needs to be an `Ident`. `runtime_event` sounds sad, though. Let me mess with it a bid.

runtime-sdk-macros/src/test_utils.rs
```diff
@@ -0,0 +1,64 @@
+pub fn rustfmt(code: &str) -> String {
+    use std::{
+        io::Write,
+        process::{Command, Stdio},
+    };
+
+    let mut cp = Command::new("rustfmt")
```
kostko, line 7: Using rustfmt directly (as was done before) doesn't work?

runtime-sdk/src/event.rs
```diff
@@ -2,6 +2,24 @@
...
 
 /// An event emitted by the runtime.
+///
+/// This trait can be derived:
+/// ```ignore
+/// # #[cfg(feature = "oasis-runtime-sdk-macros")]
+/// # mod example {
+/// # use serde::{Serialize, Deserialize};
+/// #[derive(Clone, Debug, Serialize, Deserialize, oasis_runtime_sdk::Event)]
+/// #[event(module = "path::to::MyModule", autonumber)] // `module` is required
```
kostko, line 12: This example has the `event` attribute, actual code has `sdk` :)

runtime-sdk/src/error.rs
```diff
@@ -7,6 +7,19 @@ use crate::types::transaction::CallResult;
 ///
 /// It extends `std::error::Error` with module name and error code so that errors can be easily
 /// serialized and transferred between different processes.
+///
+/// This trait can be derived:
+/// ```ignore
+/// # #[cfg(feature = "oasis-runtime-sdk-macros")]
+/// # mod example {
+/// #[derive(Clone, Debug, thiserror::Error, oasis_runtime_sdk::Error)]
+/// #[event(module = "path::to::MyModule", autonumber)] // `module` is required
```
kostko, line 16: This needs to be updated.

runtime-sdk/src/error.rs
```diff
@@ -7,6 +7,19 @@ use crate::types::transaction::CallResult;
...
+/// enum Error {
+///    InvalidArgument,      // autonumbered to 0
+///    #[event(code = 401)]  // manually numbered to 403 (`code` is required if not autonumbering)
```
kostko, line 19: Update.

runtime-sdk-macros/src/generators.rs
```diff
@@ -0,0 +1,116 @@
...
+
+pub trait CodedVariant {
+    /// The field in the helper attribute that yields the value provided by `code`.
+    /// For instance, in `#[sdk(event_code = 0)]`, the `FIELD_NAME` would be `event_code`.
+    const FIELD_NAME: &'static str;
+
+    /// The variant ident.
+    fn ident(&self) -> &Ident;
+
+    /// The code to which the variant should be converted..
```
kostko, line 34:
```suggestion
    /// The code to which the variant should be converted.
```

runtime-sdk-macros/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+
+/// Derives the `Event` trait on an enum.
+#[proc_macro_derive(Event, attributes(sdk))]
+pub fn event_derive(input: TokenStream) -> TokenStream {
+    let input = syn::parse_macro_input!(input as syn::DeriveInput);
+    event_derive::derive_event(input).into()
+}
+
+/// Derives the `Error` trait on an enum.
+// The helper attribute is `runtime_error` to avoid conflict with `thiserror::Error`.
```
kostko, line 21: Hm?

kostko (commented):
Instead of using `sdk` everywhere, maybe it would be better to use `sdk_event` and `sdk_error`? Not sure though.

runtime-sdk-macros/src/test_utils.rs
nhynes: It's an old, deprecated one; the crate says so. Including the nightly one as a lib pulls in all sorts of auto-published rust internals.

runtime-sdk/src/event.rs
nhynes: would you mind me adding [mockall](https://crates.io/crates/mockall) so that I can easily create a mock `Module` and un-`ignore` the test?

runtime-sdk-macros/src/lib.rs
nhynes: 🤦 but in my defense, I was waiting for feedback on the final helper attribute name

runtime-sdk/src/event.rs
kostko: Yes, that would be great so we could test these examples :-)

runtime-sdk-macros/src/lib.rs
kostko: He he, so as I mentioned above I think it could be `sdk_error` and `sdk_event`, what do you think?

kostko:
@nhynes can you rebase? I think there are only some minor things left to address and then we can merge this.

nhynes:
okay this should be ready to go now

PR #36, opened 2021-03-10
ts-web: multi-step typed transaction wrappers

### LINKED ISSUE #31: ts-web/rt: retrying a transaction etc?
There are some drawbacks to having a single helper that does the signing and submission, one of which is that it reduces the flexibility if you want to have the user sign it once and then send it either later or multiple times, etc. if, for example, there are network issues.

For our records, the reason we have a draft doing it this way is to carry method type information between (i) constructing the transaction with the request body type, which happens before signing, and (ii) parsing the response data, which happens after submitting.

Maybe we should look into making some clearly container structs like `UnverifiedTransaction` and `SignatureSigned` have a type parameter instead of `unknown` content. It takes us farther away from the possibility of translating type definitions automatically, but seems worth it.

pro-wh:
this is now authored on top of some loud style changes #28 #34 , which apparently we're doing today. good luck

the workflow has several steps, so the wrapper is a class with multiple functions.

this PR has expanded to change the runtime transaction and query wrappers to be similar. fixes #31


client-sdk/ts-web/core/src/client.ts
```diff
@@ -0,0 +1,1341 @@
...
+    constructor(base: string) {
+        super(base);
+    }
+
+    // beacon
+
+    /**
+     * GetBaseEpoch returns the base epoch.
+     */
+    beaconGetBaseEpoch() {
```
kostko, line 522: We could also split these into multiple classes so instead of `nic.beaconGetBaseEpoch()` you would do `nic.beacon.getBaseEpoch()`?

kostko (approved):

client-sdk/ts-web/core/src/client.ts
pro-wh: yeah I'll file this

pro-wh: #39

pro-wh:
that's weird, the only commits in `pro-wh/feature/wrapper..main` were merge commits, and I had no conflicts when I tried to merge main into pro-wh/feature/wrapper locally. maybe git works harder to resolve conflicts automatically than github.

Issue #21, opened 2021-03-03
Add E2E tests


kostko:
We should add some simple E2E tests (using the E2E test framework from Oasis Core) that make sure that a trivial runtime based on the runtime-sdk works correctly with the specified version of Oasis Core and the various client-sdk implementations.

pro-wh:
ts side is in with #51

PR #37, opened 2021-03-12
Add test runtime

abukosek:
Part of #21.  This is a port of the simple-keyvalue runtime from oasis-core using the new oasis-sdk.
It exports three methods: `keyvalue.Insert`, `keyvalue.Get`, and `keyvalue.Remove`.


tests/runtimes/simple-keyvalue/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+}
+
+impl sdk::module::AuthHandler for Module {}
+impl sdk::module::BlockHandler for Module {}
+
+impl sdk::module::MethodRegistrationHandler for Module {
+    /// Register all supported methods.
+    fn register_methods(methods: &mut sdk::module::MethodRegistry) {
+        methods.register_callable(sdk::module::CallableMethodInfo {
+            name: "keyvalue.insert",
```
kostko, line 132: The convention is to have capitalized method names, e.g. `keyvalue.Insert` :-)

abukosek: OK, I'll fix that, thanks :)

tests/runtimes/simple-keyvalue/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+
+#[derive(Error, Debug)]
+pub enum ParameterValidationError {
+    // XXX: Same as above -- what if we don't have any parameters?
+    #[error("dummy error")]
+    DummyError,
+}
+
+impl sdk::module::Parameters for Parameters {
+    type Error = ParameterValidationError;
```
kostko, line 99: If there's no errors you can use the `()` type.

tests/runtimes/simple-keyvalue/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+
+    fn validate_basic(&self) -> Result<(), Self::Error> {
```
kostko, line 101: No need to override default implementation if it is the same.

tests/runtimes/simple-keyvalue/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+            return true;
+        }
+
+        // Migrations are not supported.
+        false
+    }
+}
+
+/// Simple keyvalue runtime.
+pub struct Runtime;
```
kostko, line 258: Would be nice if the `Module` was separate from the runtime (doesn't need to be its own crate just in its own Rust module).

tests/runtimes/simple-keyvalue/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+#[serde(deny_unknown_fields)]
+pub struct KeyValue {
+    #[serde(rename = "key")]
+    #[serde(with = "serde_bytes")]
+    key: Vec<u8>,
+
+    #[serde(rename = "value")]
+    #[serde(with = "serde_bytes")]
+    value: Vec<u8>,
+}
```
kostko, line 35: Would be nice if these were defined in a separate module like `types`.

tests/runtimes/simple-keyvalue/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "simple-keyvalue-runtime"
+version = "0.1.0"
+authors = ["Oasis Protocol Foundation <info@oasisprotocol.org>"]
+edition = "2018"
+license = "Apache-2.0"
+
+[dependencies]
+oasis-runtime-sdk = { git = "https://github.com/oasisprotocol/oasis-sdk", branch = "main" }
```
kostko, line 9: This should reference the crate by relative path instead of Git. Otherwise this will not use the correct version.

tests/runtimes/simple-keyvalue/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+    fn from(err: Error) -> sdk::error::RuntimeError {
+        sdk::error::RuntimeError::new(err.module(), err.code(), &err.msg())
+    }
+}
+
+/// Events emitted by the keyvalue module (none so far).
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum Event {
+    // XXX: What if we don't want to have any events?  The SDK seems to require an implementation anyways :/
```
kostko, line 66: Make an issue or PR against the SDK to support modules without events (e.g., where the event type is `()`). Also you probably want to test events as well.

pro-wh:
I'm interested in adding e2e tests around this runtime.

I have a set of declarations here https://github.com/oasisprotocol/oasis-bridge/pull/17/files, if you want to see what it will probably look like

tests/e2e/runtime.go
```diff
@@ -0,0 +1,465 @@
...
+	return pk.String() == other.String()
+}
+
+func (pk *pkWrapper) Verify(context, message, signature []byte) bool {
+	return pk.pk.Verify(coreSig.Context(context), message, signature)
+}
+
+// Also really annoying:
+//
+// cannot use signer (type "github.com/oasisprotocol/oasis-core/go/common/crypto/signature".Signer) as type "github.com/oasisprotocol/oasis-sdk/client-sdk/go/crypto/signature".Signer in argument to kvInsert:
```
pro-wh, line 347: https://github.com/oasisprotocol/oasis-sdk/blob/fdda3859da1fde8b8b9aab28ec9d3e55315f62c9/client-sdk/go/crypto/signature/ed25519/signer.go#L30 we have this though

tests/e2e/runtime.go
```diff
@@ -0,0 +1,465 @@
...
+		return err
+	}
+	return nil
+}
+
+func kvGet(rtc client.RuntimeClient, key string) (string, error) {
+	ctx := context.Background()
+
+	var resp kvKeyValue
+	if err := rtc.Query(ctx, client.RoundLatest, "Get", kvKey{Key: key}, &resp); err != nil {
```
pro-wh, line 421: but this is registered as a callable in the runtime. does this work?

tests/e2e/runtime.go
```diff
@@ -0,0 +1,465 @@
...
+
+	val, err := kvGet(rtc, "test_key")
+	if err != nil {
+		return err
+	}
+	if val != "test_value" {
+		return fmt.Errorf("fetched value does not match inserted value")
+	}
+
+	if err := kvRemove(rtc, signer, 1, "test_key"); err != nil {
```
pro-wh, line 444: this runtime doesn't include the accounts module, so there's no nonce check

tests/e2e/runtime.go
abukosek: Oh, nice!  Don't know how I could have missed that :man_facepalming:  I'll fix this, thanks :)

tests/e2e/runtime.go
abukosek: Probably not, it's still a WIP :)

tests/e2e/runtime.go
kostko: Yes, you should add the accounts module to the test runtime as we would like to test that as well.

tests/e2e/runtime.go
```diff
@@ -0,0 +1,465 @@
...
+}
+
+func SimpleKVTest(net *oasis.Network, rtc client.RuntimeClient) error {
+	signer := newSignerWrapper(net.Entities()[1].Signer())
+
+	if err := kvInsert(rtc, signer, 0, "test_key", "test_value"); err != nil {
+		return err
+	}
+
+	// TODO: Wait for next epoch between all of these...
```
kostko, line 434: Why?

tests/e2e/runtime.go
```diff
@@ -0,0 +1,465 @@
...
+
+	var resp kvKeyValue
+	if err := rtc.Query(ctx, client.RoundLatest, "Get", kvKey{Key: key}, &resp); err != nil {
+		return "", err
+	}
+	return resp.Value, nil
+}
+
+func SimpleKVTest(net *oasis.Network, rtc client.RuntimeClient) error {
+	signer := newSignerWrapper(net.Entities()[1].Signer())
```
kostko, line 428: You can use the predefined SDK test keys:
https://github.com/oasisprotocol/oasis-sdk/blob/main/client-sdk/go/testing/testing.go

tests/e2e/runtime.go
kostko: Yeah, this won't work, either define `Get` as a query (this actually makes sense so you also test queries) or change this to issue a transaction.

tests/e2e/runtime.go
```diff
@@ -0,0 +1,465 @@
...
+
+	if err := kvRemove(rtc, signer, 1, "test_key"); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// RegisterScenarios registers all oasis-sdk end-to-end runtime tests.
+func RegisterScenarios() error {
```
kostko, line 452: Consider moving this into a separate file as this will become messy when we have multiple scenarios.

tests/e2e/runtime.go
```diff
@@ -0,0 +1,465 @@
...
+type kvKey struct {
+	Key string `json:"key"`
+}
+
+type kvKeyValue struct {
+	Key   string `json:"key"`
+	Value string `json:"value"`
+}
+
+func kvInsert(rtc client.RuntimeClient, signer signature.Signer, nonce uint64, key string, value string) error {
```
kostko, line 384: Consider moving these key-value specific APIs into a separate package that you then use in tests.

tests/e2e/runtime.go
```diff
@@ -0,0 +1,465 @@
...
+			)
+			return testErr
+		}
+		sc.Logger.Info("test passed", "test", testName)
+	}
+
+	return sc.Net.CheckLogWatchers()
+}
+
+// Ugh, this is really annoying :/
```
kostko, line 318: Why are you doing it like that? There are provisions for doing this in the SDK. See an example here (that also uses predefined test keys):
https://github.com/oasisprotocol/oasis-bridge/blob/master/examples/user-witness-flow/main.go

tests/e2e/runtime.go
```diff
@@ -0,0 +1,465 @@
...
+	cfgIasMock                 = "ias.mock"
+	cfgEpochInterval           = "epoch.interval"
+)
+
+var (
+	// RuntimeParamsDummy is a dummy instance of runtimeImpl used to register global e2e/runtime flags.
+	RuntimeParamsDummy *runtimeImpl = newRuntimeImpl("", []RunTestFunction{})
+
+	// SimpleKVRuntime is the basic network + client test case with runtime support.
+	SimpleKVRuntime scenario.Scenario = newRuntimeImpl("simple-keyvalue-runtime", []RunTestFunction{SimpleKVTest})
```
kostko, line 46: You should really split these things up, so you define the common runtime runner in one file and then the key-value runtime scenario in a separate file etc.

abukosek: Yeah, this is an early draft, I'll do the cosmetic fixes after I get it working :)

tests/e2e/runtime.go
abukosek: Since the SDK doesn't have any docs yet, it's not entirely obvious (at least to me) how things should be done.
Thanks for the link, I'll update the code to use the test keys :)

tests/runtimes/simple-keyvalue/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "simple-keyvalue-runtime"
```
kostko, line 2:
```suggestion
name = "test-runtime-simple-keyvalue"
```
I think this would make it explicit that this is something only for tests.

Cargo.toml
```diff
@@ -1,4 +1,5 @@
 [workspace]
 members = [
     "runtime-sdk",
+    "tests/runtimes/simple-keyvalue",
```
kostko, line 4:
```suggestion
    
    # Test runtimes.
    "tests/runtimes/simple-keyvalue",
```

tests/e2e/go.mod
```diff
@@ -0,0 +1,19 @@
+module github.com/oasisprotocol/oasis-sdk/tests/e2e
+
+go 1.15
+
+// Keep these in sync with oasis-core!
```
kostko, line 5:
```suggestion
// Should be synced with Oasis Core as replace directives are not propagated.
```

tests/e2e/go.mod
```diff
@@ -0,0 +1,19 @@
...
+	github.com/gorilla/websocket => github.com/gorilla/websocket v1.4.2
+
+	github.com/tendermint/tendermint => github.com/oasisprotocol/tendermint v0.34.8-oasis1
+	golang.org/x/crypto/curve25519 => github.com/oasisprotocol/ed25519/extra/x25519 v0.0.0-20210127160119-f7017427c1ea
+	golang.org/x/crypto/ed25519 => github.com/oasisprotocol/ed25519 v0.0.0-20210127160119-f7017427c1ea
+)
+
+require (
+	github.com/oasisprotocol/oasis-core/go v0.2100.1
+	github.com/oasisprotocol/oasis-sdk/client-sdk/go v0.0.0-20210319174144-fdda3859da1f
```
kostko, line 17: Can this depend on `../../client-sdk/go`? Maybe using a replace directive if it doesn't work directly? As we want to test the correct version.

abukosek: Done :)

tests/runtimes/simple-keyvalue/src/lib.rs
abukosek: Filed as #61.

tests/e2e/runtime.go
abukosek: This will be done in a separate PR.

kostko (approved):

PR #38, opened 2021-03-12
Add chain domain separation context

### LINKED ISSUE #17: Configure domain separation context
The runtime SDK should configure the domain separation context based on:
* The runtime ID.
* The consensus layer genesis document hash.

kostko:
Fixes #17


kostko:
(Coverage decreased because this now also includes `client-sdk/go` and we need to add tests there.)

pro-wh (changes requested):
placeholder comment, github is having issues

client-sdk/go/crypto/signature/secp256k1/secp256k1.go
```diff
@@ -60,6 +60,5 @@ func (pk PublicKey) Equal(other signature.PublicKey) bool {
 
 // Verify returns true iff the signature is valid for the public key over the context and message.
 func (pk PublicKey) Verify(context, message, signature []byte) bool {
-	// TODO
```
pro-wh, line 63: ?

kostko: This is obviously still a todo, there were some issues with a linter that were resolved differently so this can be added back.

pro-wh: oh thanks

pro-wh:
cross reference: getting a clean test will depend on https://github.com/oasisprotocol/oasis-bridge/pull/17 and the additional work of making the test pass.

let's bring things up to https://github.com/oasisprotocol/oasis-core/commit/d186644faee8ab99cc1e94df4bdcf04a92873e00, as used in #14, get the tests to pass, then bring things up to https://github.com/oasisprotocol/oasis-core/commit/1ca440ee0f78be5381447fe655b32cc82d59d561, as used in this PR

client-sdk/go/.golangci.yml
```diff
@@ -0,0 +1,73 @@
+# golangci-lint configuration.
```
ptrus, line 1: This seems more detailed than the one in oasis-core. Should we update the one in oasis-core as well?

client-sdk/go/go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/oasisprotocol/oasis-sdk/client-sdk/go
 go 1.15
 
 require (
-	github.com/oasisprotocol/oasis-core/go v0.2011.1-0.20210219132338-9998deaee02f
-	google.golang.org/grpc v1.35.0
+	github.com/oasisprotocol/oasis-core/go v0.2011.1-0.20210312112829-1ca440ee0f78
```
ptrus, line 6: Should we have the oasis-core replace directives here? (https://github.com/oasisprotocol/oasis-core/blob/master/go/go.mod#L4-L16)
 
Specially the `crypto/curve25519` & `crypto/ed25519` ones? Since afaik code here currently uses the non-replaced libs?

pro-wh:
tests now pass https://github.com/oasisprotocol/oasis-bridge/pull/20

pro-wh:
ts sdk updates are up, in #46

client-sdk/go/go.mod
kostko: We are already using the correct import paths in our code. The replace directives are for any external code.

I guess the biggest issue would be our Tendermint fork and to change that we would basically need to change all of the import paths in it which seems suboptimal.

client-sdk/go/.golangci.yml
kostko: I actually took this one from oasis-core-ledger, but already needed to remove some lints. Yeah we should probably sync all of these across repositories.

client-sdk/go/go.mod
ptrus: > The replace directives are for any external code.

Oh yeah right that's true, i guess it's less work keeping the directive then.
However regardless we should still manually propagate all the replace directives in all projects that use oasis-core right? (since the directives don't get propagated automatically)

kostko: Yeah we should, I'll update `go.mod`.

kostko: Done.

ptrus (approved):

client-sdk/go/client/client.go
```diff
@@ -10,16 +10,22 @@ import (
...
 
+// RoundLatest is a special round number always referring to the latest round.
 const RoundLatest = coreClient.RoundLatest
 
 // RuntimeClient is a client interface for runtimes based on the Oasis Runtime SDK.
 type RuntimeClient interface {
+	// GetInfo returns information about the runtime.
+	GetInfo(ctx context.Context) (*types.RuntimeInfo, error)
+
 	SubmitTx(ctx context.Context, tx *types.UnverifiedTransaction) (cbor.RawMessage, error)
```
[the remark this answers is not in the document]
kostko: Added.

PR #40, opened 2021-03-13
ts-web/core: run playground in cypress

pro-wh:
transferring these into broken down tests isn't going well. in the meantime, here's a simple wrapper around the existing playground file.

~~current issues: errors don't get propagated up to cypress. it just hangs until timeout~~ filed #41 

~~next steps will be setting it up for CI, with all the steps leading up to this where you start the testnet, start the proxy, and start the webpack dev server~~ added here

for #21


pro-wh:
added the CI setup to this PR

client-sdk/ts-web/core/playground/sample-envoy.yaml
```diff
@@ -56,7 +56,7 @@ static_resources:
         - endpoint:
             address:
               pipe:
-                path: node/internal.sock
+                path: /tmp/oasis-net-runner-sdk-core/net-runner/network/validator-0/internal.sock
```
kostko, line 59: Shouldn't this point to a client node?

client-sdk/ts-web/core/playground/src/index.js
```diff
@@ -96,16 +64,48 @@ const nic = new oasis.client.NodeInternal('http://localhost:42280');
             console.log('sent');
         }
 
+        // Try verifying transaction signatures.
+        {
+            // TODO: Make sure this is the block with the transaction we sent above.
+            const genesis = await nic.consensusGetGenesisDocument();
```
kostko, line 70: You could use the new `GetChainContext` method instead of fetching the whole genesis document.

client-sdk/ts-web/rt/cypress/integration/playground.spec.js
```diff
@@ -0,0 +1,8 @@
+/// <reference types="cypress" />
+
+describe('playground', () => {
+    it('finishes', () => {
+        cy.visit('http://localhost:8080/');
+        cy.contains('barolitopsis', {timeout: 60_000});
```
kostko, line 6: Lol, could we put these into some (shared) constant?

client-sdk/ts-web/core/playground/sample-envoy.yaml
pro-wh: It should. I found out though that by not provisioning a runtime, it doesn't create a client node. So at this time I've switched it over to connect directly to the validator \\:

client-sdk/ts-web/core/playground/src/index.js
pro-wh: that's a good idea. I'll file that. I also noticed that we have a new GetAccountNonce thing that could replace a part where we get the entire account to read the nonce.

pro-wh: #48

client-sdk/ts-web/rt/cypress/integration/playground.spec.js
pro-wh: oh yeah, I'll see if we can

pro-wh: we can! I'll do this in oasis-bridge too

client-sdk/ts-web/core/playground/sample-envoy.yaml
pro-wh: opened https://github.com/oasisprotocol/oasis-core/issues/3794 to ask about this

kostko: Oh, so there will be a separate test for runtimes, I see.

client-sdk/ts-web/core/playground/src/index.js
kostko: Yeah the `GetSignerNonce` has been there for a while though.

kostko (approved):

client-sdk/ts-web/core/playground/consts.sh
```diff
@@ -0,0 +1,3 @@
+BUILD_NUMBER=3935
```
kostko, line 1: Once we release 21.0 it would be better if these were fetched from a stable release. Maybe open an issue to change this.

pro-wh: let's see if we can do that in #46. I'll file if we don't

pro-wh: switched over in #46

PR #46, opened 2021-03-17
changes from oasis-core master, separate batch

pro-wh:
up to https://github.com/oasisprotocol/oasis-core/commit/1ca440ee0f78be5381447fe655b32cc82d59d561

on top of #19


pro-wh:
this is now up to oasis-core@1ca440ee0f78be5381447fe655b32cc82d59d561. it has only the typescript sdk changes. bumps to the go and rust dependencies are in #38

pro-wh:
I've added two more commits to take us up to v21.0

kostko (approved):

Issue #7, opened 2021-02-20
ts-web/core release roadmap


pro-wh:
## prereqs
- [x] "documentation" :hand_over_mouth: #4 
- [x] license #6 

## intermediate release: share with our developer partners
- [x] ~~add read permission on github~~

## middle phase
- [ ] (add further discovered tasks here)

## public release
- [ ] bump version to 1.0.0
- [x] flip this repo public
- [x] create npm repos
- [x] build and upload to npm
- [x] ~~clean up allowed readers~~

pro-wh:
all that's left is the 1.0.0. but let's track that in #69

PR #47, opened 2021-03-17
ci: upload ts-web packages

pro-wh:
we'll need these for oasis-bridge CI. this might be convenient for our developer partners too.

cross reference: #7

on top of #19


pro-wh:
update: visitors can't access these artifacts. it won't be useful for any of those proposed use cases.

PR #51, opened 2021-03-20
ts-web/rt tests with keyvalue runtime

pro-wh:
uses #37, cc @abukosek

### DIFF
--- a/.github/workflows/ci-test.yaml
+++ b/.github/workflows/ci-test.yaml
@@ -121,7 +121,7 @@ jobs:
       - name: 'proxy: Install Envoy'
         # https://www.getenvoy.io/install/envoy/ubuntu/
         run: |
-          curl -sL 'https://getenvoy.io/gpg' | sudo apt-key add -
+          curl -sL 'https://www.getenvoy.io/gpg' | sudo apt-key add -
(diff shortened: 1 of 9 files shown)


pro-wh:
can't get the testnet to work. pls help

pro-wh:
this now runs locally. something's getting tied up in initialization or something, with the first `insert` call taking ~25 seconds.

might have some issue running on CI, because when I run it locally, it fails for a while with `client: not finished initial sync`. will need to add a wait for that

kostko:
> this now runs locally. something's getting tied up in initialization or something, with the first insert call taking ~1 minute.

That's normal as it requires some initial epoch transitions to set up committees.

pro-wh:
getting stream reset from client-0's internal.sock on the first WaitReady call, according to envoy. but client-0's logs say it never even received a request. can't reproduce this locally.

pro-wh:
seemed to be a CORS thing. the cypress test was pointed at the wrong port. dunno why it worked locally though

kostko (approved):
The test key value runtime needs some fixes as indicated in the other PR. But let's merge this so we have some E2E tests.

pro-wh:
thanks, I'll rebase

PR #55, opened 2021-03-22
ts-web/core: use client node for playground grpc

pro-wh:
depends #54 

This seems to need a little something more, with our test hanging when directed at the client node instead of the validator

### DIFF
--- a/.github/workflows/ci-test.yaml
+++ b/.github/workflows/ci-test.yaml
@@ -139,7 +139,7 @@ jobs:
 
       - name: 'testnet: Wait for node socket'
         run: |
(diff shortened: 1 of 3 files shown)


pro-wh:
it works locally, but admittedly I don't start the test as fast as CI does. maybe there's something else we need to wait for

pro-wh:
I tried launching it fast manually, and the test fails with `consensus: no committed blocks`. Maybe we can add a step to wait for the first block

pro-wh:
after some more experimentation, I've found that it takes ~30 seconds on my computer for client-0 to connect to the testnet. maybe it's taking longer on CI, where there's a 60 second timeout

pro-wh:
I added a `WaitReady` call, and it works now.

kostko (approved):

PR #59, opened 2021-03-24
ts-web/rt: add chain domain separation context

pro-wh:
on top of #38


kostko:
I've cherry-picked your commit to #38 and merged that.

PR #63, opened 2021-03-26
ts-web/core: revert to tweetnacl

Esya:
Replace the EllipticSigner with NaclSigner

Elliptic is not able to sign using the 64 bytes private key that we will need to use once we support key derivation.
For reference the go package does it right : https://golang.org/pkg/crypto/ed25519/ and allows you to instanciate a keypair from the seed or from the 64 bytes private key.

The whole process is described here : https://blog.mozilla.org/warner/2011/11/29/ed25519-keys/

This is required in order to eventually implement [BIP32-Ed25519](https://github.com/oasisprotocol/oasis-core/pull/3656/)

### DIFF
--- a/client-sdk/ts-web/core/package.json
+++ b/client-sdk/ts-web/core/package.json
@@ -14,12 +14,11 @@
     "dependencies": {
         "bech32": "^1.1.4",
         "cborg": "^1.1.2",
-        "elliptic": "^6.5.3",
         "grpc-web": "^1.2.1",
-        "js-sha512": "^0.8.0"
+        "js-sha512": "^0.8.0",
+        "tweetnacl": "^1.0.3"
     },
     "devDependencies": {
-        "@types/elliptic": "^6.4.12",
         "cypress": "^6.6.0",
         "typescript": "^4.1.3",
         "webpack": "^5.21.2",
--- a/client-sdk/ts-web/rt/playground/src/index.js
+++ b/client-sdk/ts-web/rt/playground/src/index.js
@@ -98,7 +98,7 @@ const keyvalueWrapper = new Wrapper(KEYVALUE_RUNTIME_ID);
             const THE_KEY = oasis.misc.fromString('greeting-js');
             const THE_VALUE = oasis.misc.fromString('Hi from JavaScript');
 
-            const alice = oasis.signature.EllipticSigner.fromSecret(await oasis.hash.hash(oasis.misc.fromString('oasis-runtime-sdk/test-keys: alice')), 'this key is not important');
+            const alice = oasis.signature.NaclSigner.fromSeed(await oasis.hash.hash(oasis.misc.fromString('oasis-runtime-sdk/test-keys: alice')), 'this key is not important');
             const csAlice = new oasis.signature.BlindContextSigner(alice);
             // The keyvalue runtime does not use the accounts module, so there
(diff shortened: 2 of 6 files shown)


client-sdk/ts-web/core/src/misc.ts
```diff
@@ -1,4 +1,3 @@
-// @ts-expect-error missing declaration
```
Esya, line 1: This was preventing npm i from generating a package-lock properly and ts-web/core had no package-lock file

pro-wh: This was recently removed from main. Could you rebase?

client-sdk/ts-web/core/package-lock.json
```diff
@@ -0,0 +1,5752 @@
+{
```
pro-wh, line 1: We're on the npm 7 boat, where there's a shared package-lock.json one directory level out. Would you be able to use npm 7 to run `npm install` out in `client-sdk/ts-web` and update that package-lock.json instead of adding this file?

client-sdk/ts-web/rt/playground/src/index.js
```diff
@@ -98,7 +98,7 @@ const keyvalueWrapper = new Wrapper(KEYVALUE_RUNTIME_ID);
             const THE_KEY = oasis.misc.fromString('greeting-js');
             const THE_VALUE = oasis.misc.fromString('Hi from JavaScript');
 
-            const alice = oasis.signature.EllipticSigner.fromSecret(await oasis.hash.hash(oasis.misc.fromString('oasis-runtime-sdk/test-keys: alice')), 'this key is not important');
+            const alice = oasis.signature.NaclSigner.fromSeed(await oasis.hash.hash(oasis.misc.fromString('oasis-runtime-sdk/test-keys: alice')), 'this key is not important');
```
Esya, line 101: @pro-wh This is imported from `core` - not sure if that breaks anything as I haven't played with `rt` yet

client-sdk/ts-web/core/src/misc.ts
Esya: Done

client-sdk/ts-web/core/package-lock.json
Esya: Done

client-sdk/ts-web/rt/playground/src/index.js
pro-wh: seems like it should result in the same key, from what we discussed earlier. maybe we could add a check against a reference public key

pro-wh (approved):
looks good, thanks

client-sdk/ts-web/core/package.json
```diff
@@ -13,13 +13,12 @@
     },
     "dependencies": {
         "bech32": "^1.1.4",
-        "cborg": "^1.1.2",
-        "elliptic": "^6.5.3",
+        "cborg": "^1.0.4",
```
kostko, line 16: Why the downgrade?

Esya: Ah, that's a typo from switching between branches, my bad. Will fix that right now.

kostko:
Rebase and then it's good to go. Thanks for your contribution!

PR #64, opened 2021-03-27
ci: propagate playground errors to cypress

pro-wh:
the overall design of Cypress is very adverse to this idea, that you should be able to see what went wrong

so after a day's experimentation, here's roughly how we might be able to do it:

1. export the great big async code into a global promise
  a. we have to configure webpack to do that
  b. but webpack-dev-server breaks that, making all exports undefined. thanks https://github.com/webpack/webpack-dev-server/issues/2484
  c. but webpack-dev-server 4 beta fixes that, so upgrade to that
  d. but it also changes the configuration schema in ways that aren't documented yet
  e. so fix those or move things into defaults and get rid of the configuration altogether. leaning towards the latter
2. reach in from cypress and grab that promise

so we won't need to do that hacky 'cypress-done-core' stuff anymore either


pro-wh:
currently have to figure out all the timeout systems

- `.playground` should be present really quickly
- it could take a long time to fulfill
- if it rejects, we should fail right away. but unfortunately cypress can't fathom the idea that the test makes progress other than towards success

client-sdk/ts-web/core/playground/src/index.js
```diff
@@ -1,147 +1,141 @@
 // @ts-check
 
 import * as oasis from './../..';
-import * as shared from './shared';
 
 const nic = new oasis.client.NodeInternal('http://localhost:42280');
 
-(async function () {
-    try {
```
pro-wh, line 9: this is best viewed with hidden whitespace changes. we're outdenting the whole script a level due to taking out the `try`

pro-wh:
ready for review

sample failure, which actually happened when I resolved a merge conflict wrong: https://github.com/oasisprotocol/oasis-sdk/runs/2223415566?check_suite_focus=true

```
  1) playground
       finishes:
     Error: bad secret key size
      at Function.nacl.sign.keyPair.fromSecretKey (webpack://playground/../../core/node_modules/tweetnacl/nacl-fast.js?:2320:11)
      at Function.fromSecret (webpack://playground/../../core/dist/signature.js?:101:39)
      at eval (webpack://playground/./src/index.js?:105:95)
```

kostko (approved):
Nice!

PR #65, opened 2021-03-27
ts-web/core: Adds support for hierarchical deterministic keys

Esya:
# Summary

Implements [ADR 0008: Entity Key Generation Process](https://github.com/oasisprotocol/oasis-core/pull/3656/) in ts-web/core, standardizing the key generation process using [BIP-0039] to generate the seed, and [BIP32-Ed25519] for hierarchical deterministic key generation. 

The value proposition mentioned in [ADR 0008] is as follows : 

> Adopting these standards will allow the Oasis ecosystem to:
> 
> - Make key derivation the same across different applications (i.e. wallets).
> - Allow users to hold keys in hardware wallets.
> - Allow users to hold keys in cold storage more reliably (i.e. using the
  familiar 24 word mnemonics).
> - Define how users can generate multiple keys from a single seed (i.e.
  the 24 or 12 word mnemonic).

# Usage

## Mnemonic and derivation
We add a new class `HdKey` inside `ts-web/core` that allows hierarchical deterministic keys to be generated from [BIP-0039] mnemonics and derived using [BIP32-Ed25519], such as : 

```typescript
const HdKey = oasis.hdkey.HdKey;

// From existing mnemonic
const root = HdKey.fromMnemonic('practice acoustic taxi');
const key = root.derivePath("44'/474'/0'/0'/0'");

// Generating a secure mnemonic
const mnemonic = await HdKey.generateValidMnemonic();
const root = HdKey.fromMnemonic(mnemonic);
const key = root.derivePath("44'/474'/0'/0'/0'");
```

## Signing with HdKey

`HdKey` implements `Signer` from `core/src/signature.ts` and can be used with `TransactionWrapper`  and `BlindContextSigner` : 

```typescript
const root = HdKey.fromMnemonic('practice acoustic taxi');
const key = root.derivePath("44'/474'/0'/0'/0'");

const tw = oasis.staking.transferWrapper();

// ... Build transaction ...

await tw.sign(new oasis.signature.BlindContextSigner(key), chainContext);
```

# Other changes to ts-web
## Jest support

This pull request adds `jest` as a framework for unit and integration tests, since the only existing tool was cypress that is more suited for e2e tests. While we should keep cypress for when testing browser integration, I propose this becomes standard for unit & integration tests and added a sane and simple `jest.config.js` in `ts-web` that can then be extended or reused by the underlying packages. 

I chose to use `@babel/preset-typescript` instead of `ts-jest` as it is faster and we do not need type-checking when running the tests as we already run tsc at build time/during CI. Just like jest, a sane babel config file lives in `ts-web` and can be extended by the children packages.

This PR also adds a github action job to run the jest tests for `core`

## Unified webpack configurations

The webpack configuration files from the various playgrounds (`core`, `rt` and `signer-ledger`) were changed to use a main one in `ts-web` .

Webpack 5 does not automatically include polyfills anymore and they must now be explicitly listed - so we polyfill `stream` (needed for crypto/cipher-base part of bip39) for the browser.

# To be discussed

- Currently, [BIP32-Ed25519] states : " If the third highest bit of the last byte of k<sub>L</sub> is not zero, discard the master secret". This means that some [BIP-0039] keys are "better than other", and while when we generate a mnemonic, we should only keep mnemonics satisfying this condition, what should we do about existing mnemonics that some users might rightfully be using and might want to use for Oasis ? I'm not sure about the security implications. For now, `HdKey` will refuse those mnemonics, we can make it more lenient if it is proven that this is not a security issue.

[ADR 0008]: https://github.com/oasisprotocol/oasis-core/pull/3656/
[BIP-0032]: https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki
[BIP-0039]: https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki
[BIP32-Ed25519]:
  https://raw.githubusercontent.com/LedgerHQ/orakolo/master/papers/Ed25519_BIP%20Final.pdf

### DIFF
--- a/client-sdk/ts-web/core/babel.config.js
+++ b/client-sdk/ts-web/core/babel.config.js
@@ -1,0 +1,1 @@
+module.exports = require('../babel.config');
--- a/client-sdk/ts-web/core/src/hdkey.ts
+++ b/client-sdk/ts-web/core/src/hdkey.ts
@@ -1,0 +1,234 @@
+import {eddsa} from 'elliptic';
+import {sha512} from 'js-sha512';
+import {sha256} from 'js-sha256';
+import * as BN from 'bn.js';
+import {concat} from './misc';
+import {Signer} from './signature';
+import {generateMnemonic, mnemonicToSeed} from 'bip39';
+
+// Shorthand for Bignumber instanciation
+const bn = (i: number | string | Uint8Array, endianness?: 'le' | 'be') => new BN.BN(i, endianness);
+
+/**
+ * Hierarchical deterministic keys using Ed25519
+ *
+ * Implements BIP32-Ed25519 to securely derive keys from a parent
+ * 32-bytes seed, itself usually generated from a BIP39 mnemonic.
+ */
+export class HDKey implements Signer {
+    public readonly secret: Uint8Array;
+
+    // Curve Parameters
+    protected static eddsa = new eddsa('ed25519');
+    protected static bp = HDKey.eddsa.curve.g;
+    protected static n = HDKey.eddsa.curve.n;
+
+    constructor(
+        public readonly kL: Uint8Array,
+        public readonly kR: Uint8Array,
+        public readonly A: Uint8Array,
+        public readonly c: Uint8Array,
+    ) {
+        this.secret = concat(kL, kR);
+    }
+
+    /**
+     * Returns the root node for the given mnemonic
+     * @param mnemonic Bip39 Mnemonic
+     * @returns Root node
+     */
+    public static async fromMnemonic(mnemonic: string): Promise<HDKey> {
+        const seed = await mnemonicToSeed(mnemonic);
+        return HDKey.fromSeed(seed.slice(0, 32));
+    }
+
+    /**
+     * Generates a safe mnemonic
+     * @param strength Length in bits of the generated mnemonic
+     * @returns Generated BIP39 Mnemonic
+     */
+    public static async generateValidMnemonic(strength = 256): Promise<string> {
+        let mnemonic = generateMnemonic(strength);
+        while (!(await HDKey.isValidMnemonic(mnemonic))) {
+            mnemonic = generateMnemonic(strength);
+        }
+
+        return mnemonic;
+    }
+
+    /**
+     * Checks if the given bip39 mnemonic is valid for Ed25519 derivation
+     * If it is not, you must generate another one.
+     * @param mnemonic
+     */
+    public static async isValidMnemonic(mnemonic: string): Promise<boolean> {
+        const seed = await mnemonicToSeed(mnemonic);
+        const hash = new Uint8Array(sha512.arrayBuffer(seed.slice(0, 32)));
+
+        // If the third highest bit of the lasst byte of kL is not zero, discard k
+        return (hash[31] & 0b00100000) === 0;
+    }
+
+    /**
+     * Returns the root node for the given 32 bytes seed
+     * @param seed 32 bytes seed
+     * @returns Root node
+     */
+    public static fromSeed(seed: Uint8Array): HDKey {
+        if (seed.length != 32) throw 'Seed must be 32 bytes (a Uint8Array of size 32)';
+
+        const k = new Uint8Array(sha512.arrayBuffer(seed));
+        const kL = k.slice(0, 32),
+            kR = k.slice(32);
+
+        //  If the third highest bit of the last byte of kL is not zero, discard k
+        if (kL[31] & 0b00100000) {
+            return null;
+        }
+
+        // clear lowest three bits of the first byte
+        // clear highest bit of the last byte
+        // set second highest bit of the last byte
+        kL[0] &= 248;
+        kL[31] &= 127;
+        kL[31] |= 64;
+
+        // Get the public key
+        const A = new Uint8Array(HDKey.eddsa.encodePoint(HDKey.bp.mul(bn(kL, 'le'))));
+
+        const chaincodePayload = concat(new Uint8Array([1]), seed);
+        const c = new Uint8Array(sha256.arrayBuffer(chaincodePayload));
+
+        return new HDKey(kL, kR, A, c);
+    }
+
+    /**
+     * Derive a new key at the specified derivation path from the current key
+     * @param path Derivation path, using ' for hardened indexes
+     * @returns Child node at path
+     */
+    public derivePath(path: string): HDKey {
+        let node: HDKey = this;
+        const chain = path.split('/');
+
+        for (let i = 0; i < chain.length; i++) {
+            let index;
+            if (!chain[i]) {
+                continue;
+            } else if (chain[i].endsWith("'")) {
+                // Hardened offset ; @todo 0x8...
+                index = bn(chain[i].slice(0, -1)).add(bn(2).pow(bn(31)));
+            } else {
+                // Non hardened path
+                index = bn(chain[i]);
+            }
+
+            node = node.deriveChild(index);
+        }
+        return node;
+    }
+
+    /**
+     * Derive a new key from the current key
+     * Use index > 2^31 for a hardened key,
+     * or (index + 0x80000000).
+     *
+     * @param index Index to use for derivation
+     * @returns Child node at given index
+     */
+    public deriveChild(index: BN): HDKey {
+        const kLP = this.kL,
+            kRP = this.kR,
+            AP = this.A,
+            cP = this.c;
+
+        if (!(index.gte(bn(0)) && index.lt(bn(2).pow(bn(32))))) {
+            throw 'Index i must be between 0 and 2^32 - 1, inclusive';
+        }
+
+        const i_bytes = index.toBuffer('le', 4);
+
+        let Z: Uint8Array, c: Uint8Array;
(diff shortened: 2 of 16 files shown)


client-sdk/ts-web/core/src/hdkey.ts
```diff
@@ -0,0 +1,234 @@
+import {eddsa} from 'elliptic';
```
kostko, line 1: Wasn't the idea to use nacl?

client-sdk/ts-web/core/src/hdkey.ts
```diff
@@ -0,0 +1,234 @@
...
+     * @returns Child node at path
+     */
+    public derivePath(path: string): HdKey {
+        let node: HdKey = this;
+        const chain = path.split('/');
+
+        for (let i = 0; i < chain.length; i++) {
+            let index;
+            if (!chain[i]) continue;
+            if (chain[i].endsWith("'"))
```
kostko, line 120: I suggest using explicit braces everywhere even if the body is just a single line.

Is there a lint that we can enable for this?

client-sdk/ts-web/core/tests/hdkey.test.ts
```diff
@@ -0,0 +1,124 @@
...
+
+            expect(tooShort).toThrowError(/must be 32 bytes/);
+            expect(tooLong).toThrowError(/must be 32 bytes/);
+        });
+
+        it('Should return node', () => {
+            const node = HdKey.fromSeed(new Uint8Array(32));
+
+            expect(toHex(node.A)).toBe(
+                '3b6a27bcceb6a42d62a3a8d02a6f0d73653215771de243a63ac048a18b59da29',
```
kostko, line 61: Where are these test cases from? Would be great if we could use some standard test vectors in these tests.

pro-wh (commented):
I just saw Jernej commented. Gonna dump this current batch and see if I can discuss some of the things, since we had talked earlier.

haven't really looked at the crypto yet

client-sdk/ts-web/core/src/hdkey.ts
pro-wh: some of the hdwallet derivation steps needed direct access to some curve operations. we briefly talked about the idea of seeing if doing the same with tweetnacl internals would be possible. @Esya, have you looked into that?

client-sdk/ts-web/core/src/hdkey.ts
pro-wh: https://github.com/prettier/prettier/issues/7659#issuecomment-590004436 ): I just searched, and it looks like prettier doesn't want to include this kind of thing in the scope of their tool. we should probably go down the road of eslint then

client-sdk/ts-web/core/src/hdkey.ts
```diff
@@ -0,0 +1,234 @@
...
+    public async sign(message: Uint8Array): Promise<Uint8Array> {
+        const kL = this.kL;
+        const kR = this.kR;
+        const A = new Uint8Array(this.A);
+        const M = message;
+        const n = HdKey.n; // Base order
+
+        const r_hash = new Uint8Array(sha512.arrayBuffer(concat(kR, M)));
+        const r = bn(r_hash, 10, 'le').mod(n);
+        const R = HdKey.eddsa.encodePoint(HdKey.bp.mul(r));
```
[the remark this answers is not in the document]
kostko: I was thinking about this as well. I think it would make sense to separate key derivation from the actual signer.

client-sdk/ts-web/core/src/hdkey.ts
Esya: Elliptic only supports signing from 32-bytes seeds - called `k` or `private key` in [RFC 8032](https://tools.ietf.org/html/rfc8032) - but with BIP32-Ed25519, we have the post-hash private key (64 bytes)  - Elliptic is still needed for access to and manipulation of the 25519 Curve. 

I initially wanted to sign using `tweetnacl`, but it seems that the js implementation of NaCl does not allow to sign using the post-hash 64-bytes either, as their 64 bytes that they call *private keys* are actually 32-bytes seed/32-bytes public key to save time on computing the public key which is expensive. 

The paper mentions : 

> Due to the way child Ed25519 keys are derived, they will not have the 32-byte master secret key. We note that the NaCl library [3]provides the necessary interface to the extended private key.

[This article describes the differences between major libs](https://blog.mozilla.org/warner/2011/11/29/ed25519-keys/)

This [java implementation for instance](https://github.com/str4d/ed25519-java/blob/master/src/net/i2p/crypto/eddsa/spec/EdDSAPrivateKeySpec.java#L77) gets it right and allows instanciating directly from `h` (the post-hash 64-bytes signing key) 

But that does not seem to be the case for `tweetnacl` - I'd need further testing and comparison between the C implementation and the JS implementation, but I'm not a crypto expert and that'd be time consuming. It's unlikely but possible that this was overlooked when porting - so I've implemented the signing function myself, which I've tested verifies with all libs and with oasis-core. I'm fine reverting my previous PR if that makes more sense in the meantime.

client-sdk/ts-web/core/src/hdkey.ts
Esya: Prettier does not support this rule, I'll change it right away though

client-sdk/ts-web/core/tests/hdkey.test.ts
Esya: Saddly I could not find any test vectors for bip32-ed25519 derivation - so I made my own and checked them against a running oasis-node and verified the signed messages/pub keys using various libraries to cross-check my implementation

client-sdk/ts-web/core/src/hdkey.ts
```diff
@@ -0,0 +1,234 @@
...
+ */
+export class HdKey implements Signer {
+    public readonly secret: Uint8Array;
+
+    // Curve Parameters
+    protected static eddsa = new eddsa('ed25519');
+    protected static q = bn(2).pow(bn(255)).sub(bn(19));
+    protected static by = bn(5).mod(HdKey.q).invm(HdKey.q).mul(bn(4)).mod(HdKey.q);
+    protected static bp = HdKey.eddsa.curve.pointFromY(HdKey.by);
+    protected static n = bn(2).pow(bn(252)).add(bn('27742317777372353535851937790883648493'));
```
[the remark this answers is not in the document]
Esya: You're right, I didn't think with their convention that the basepoint would be called `g` and would not be under `curve` - fixing this.

client-sdk/ts-web/core/src/hdkey.ts
```diff
@@ -0,0 +1,234 @@
...
+        } else {
+            // hardened child
+            Z = this.hmacSha512(concat(new Uint8Array([0]), kLP, kRP, i_bytes), cP);
+            c = this.hmacSha512(concat(new Uint8Array([1]), kLP, kRP, i_bytes), cP).slice(32);
+        }
+
+        const ZL = Z.slice(0, 28),
+            ZR = Z.slice(32);
+
+        const kLn = bn(ZL, 10, 'le').mul(bn(8)).add(bn(kLP, 10, 'le'));
```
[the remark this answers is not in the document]
Esya: I did not realize that `BN`  had an additional constructor signature that did not need `base` and just took the other two params (the input and the endianness), I've changed it to reflect it

client-sdk/ts-web/core/src/hdkey.ts
```diff
@@ -0,0 +1,234 @@
+import {eddsa} from 'elliptic';
+import {sha512} from 'js-sha512';
+import {sha256} from 'js-sha256';
```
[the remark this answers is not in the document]
Esya: Thought about this too but saw js-sha512 being used in core/src/hash.ts

client-sdk/ts-web/core/babel.config.js
```diff
@@ -0,0 +1 @@
+module.exports = require('../babel.config')
```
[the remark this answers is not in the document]
Esya: Should it be done in this PR or in another one?

We could do this in `package.json`  - this would also lint and fix the playground : 

```json
    "scripts": {
        "lint": "prettier --check '**/src/**.ts' '*.js' '**/src/*.js'",
        "fmt": "prettier --write '**/src/**.ts' '*.js' '**/src/*.js'"
    },
```

client-sdk/ts-web/core/tests/hdkey.test.ts
kostko: Ok, then just provide a comment on how these were generated.

client-sdk/ts-web/webpack.config.js
```diff
@@ -0,0 +1,15 @@
+const webpack = require('webpack');
+
+module.exports = {
+    mode: 'development',
+    devServer: {
+        contentBase: 'dist',
```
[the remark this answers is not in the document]
Esya: Thanks for letting me know - I'll have a look and fix the conflicts later once we figure out what to do with nacl !

Esya:
Closing this outdated PR - will open a new one from another branch shortly with slip-0010 support

PR #68, opened 2021-03-30
runtime-sdk/crypto: Derive PartialEq and Eq for public keys

kostko:


ptrus (approved):

pro-wh (approved):
it is as described

these aren't secret, so I hope there's nothing special about how these are implemented

will we be using this soon?

PR #72, opened 2021-03-30
runtime-sdk: allow macros to work with generic modules

nhynes:
After attempting to use the macros in the bridge, it became evident that the macros would not easily work with generic `Module` types. I can't really think of a good way to get type information in that wouldn't also degrade developer UX or maintainability of this codebase, so this just makes the module name a const parameter.


kostko (commented):
Looks good.

Could `module_name` in the macros default to `MODULE_NAME`? Slightly more magic but it would avoid constantly having to specify it in case a default is used.

nhynes:
actually, scratch this PR. In order to autogen the RPC handlers from a trait, the bound will need to exist since it's required for the impl. I'll just add another field that contains the trait bound.

PR #73, opened 2021-03-30
ts-web: bump versions for alpha release

pro-wh:
We can put a prerelease version on npm. We discussed this internally, and we didn't come up with anything that should block this.

for #7


pro-wh:
@ptrus I've thus far manually bumped all the versions up to the same `0.1.0-alpha1`. shall we go with that for now?

kostko (approved):

PR #75, opened 2021-03-31
CI: Run e2e test harness

### LINKED ISSUE #74: tests/e2e/simplekvtest.go needs update for signature context
from #38, there's a new parameter in `AppendSign`.

Over in https://github.com/oasisprotocol/oasis-bridge/pull/31, I'm adding a call to `GetInfo` to get the right context.

cc @abukosek 

I think there's ongoing work to hook this up to CI, but I think we don't have a PR open for it, so I'm filing a standalone issue

abukosek:
Part of #21, also fixes #74.

### DIFF
--- a/tests/e2e/go.mod
+++ b/tests/e2e/go.mod
@@ -17,6 +17,6 @@ replace (
 
 require (
 	github.com/oasisprotocol/oasis-core/go v0.2100.1
-	github.com/oasisprotocol/oasis-sdk/client-sdk/go v0.0.0-20210319174144-fdda3859da1f
+	github.com/oasisprotocol/oasis-sdk/client-sdk/go v0.0.0-20210328195842-4de788c1c6f7
 	google.golang.org/grpc v1.36.0
 )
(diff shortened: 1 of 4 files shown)


tests/e2e/go.mod
```diff
@@ -17,6 +17,6 @@ replace (
 
 require (
 	github.com/oasisprotocol/oasis-core/go v0.2100.1
-	github.com/oasisprotocol/oasis-sdk/client-sdk/go v0.0.0-20210319174144-fdda3859da1f
+	github.com/oasisprotocol/oasis-sdk/client-sdk/go v0.0.0-20210328195842-4de788c1c6f7
```
pro-wh, line 20: how do we manage this?

abukosek: There's a replace directive in the block above, so it always uses the current files:
```
	github.com/oasisprotocol/oasis-sdk/client-sdk/go => ../../client-sdk/go
```
I just updated this because it seems nice to have a more recent hash there, even if it's unused :)

wh0: oh ok

wh0 (approved):

kostko (approved):

PR #77, opened 2021-03-31
Remove StorageKey traits

### LINKED ISSUE #76: Consider replacing `StoreKey` with `AsRef<[u8]>`
`AsRef<[u8]>` seems about what the `StoreKey` trait is going for. It also comes with a shiny new const generic impl. Whether this issue is relevant depends on whether `StoreKey` will ever not be just `as_store_key()`.

nhynes:
resolves #76 

this PR is an RFC. I won't be sad if it's rejected.

### DIFF
--- a/runtime-sdk/src/storage/typed.rs
+++ b/runtime-sdk/src/storage/typed.rs
@@ -1,10 +1,10 @@
-use std::marker::PhantomData;
+use std::{convert::TryFrom, marker::PhantomData};
 
 use serde::{de::DeserializeOwned, Serialize};
 
 use oasis_core_runtime::{common::cbor, storage::mkvs};
 
-use super::{DecodableStoreKey, Store, StoreKey};
+use super::Store;
 
 /// A key-value store that transparently handles serialization/deserialization.
 pub struct TypedStore<S: Store> {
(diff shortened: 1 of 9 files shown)


runtime-sdk/src/types/token.rs
```diff
@@ -18,14 +16,18 @@ impl Denomination {
     /// Denomination in native token.
     pub const NATIVE: Denomination = Denomination(Vec::new());
 
-    // TODO: Enforce maximum length during deserialization.
```
kostko, line 21: So you add `TryFrom` below, but the serialization I meant here was also CBOR serialization (currently provided by the derived `Deserialize` trait). Do we need a custom implementation that forwards to `Denomination::try_from`?

runtime-sdk/src/types/address.rs
```diff
@@ -191,24 +194,6 @@ impl<'de> serde::Deserialize<'de> for Address {
...
-    }
-}
-
-impl StoreKey for &Address {
-    fn as_store_key(&self) -> &[u8] {
-        &self.0
-    }
-}
-
-impl DecodableStoreKey for Address {
```
kostko, line 206: Where's the replacement `TryFrom` implementation?

runtime-sdk/src/types/token.rs
```diff
@@ -37,31 +39,32 @@ impl fmt::Display for Denomination {
...
+    fn from_str(v: &str) -> Result<Self, Self::Err> {
+        Self::try_from(v.as_bytes())
     }
 }
 
-impl StoreKey for &Denomination {
-    fn as_store_key(&self) -> &[u8] {
-        &self.0
+impl TryFrom<&[u8]> for Denomination {
+    type Error = Error;
```
kostko, line 50: Empty line below this line.

runtime-sdk/src/types/token.rs
```diff
@@ -37,31 +39,32 @@ impl fmt::Display for Denomination {
...
-    fn from(v: &str) -> Denomination {
-        Denomination(v.as_bytes().to_vec())
-    }
-}
-
-impl StoreKey for Denomination {
-    fn as_store_key(&self) -> &[u8] {
-        &self.0
+impl std::str::FromStr for Denomination {
+    type Err = Error;
```
kostko, line 43: Empty line below this line.

runtime-sdk/src/storage/typed.rs
```diff
@@ -57,17 +69,15 @@ impl<'store, K: DecodableStoreKey, V: DeserializeOwned> TypedStoreIterator<'stor
...
 
-impl<'store, K: DecodableStoreKey, V: DeserializeOwned> Iterator
-    for TypedStoreIterator<'store, K, V>
+impl<'store, K, V> Iterator for TypedStoreIterator<'store, K, V>
+where
+    K: for<'k> TryFrom<&'k [u8]>,
+    V: DeserializeOwned,
 {
-    type Item = (K, V);
+    type Item = (K, V); // TODO: Should this return `Result`?
```
kostko, line 77: This would make it more awkward for the caller to use. And if deserialization fails, this means that the stored state is somehow corrupted/incompatible, so I'm not sure if we should let the caller handle it? Are there any nice ways of handling such iterators in the caller, by e.g. propagating an error there if needed without requiring too much boilerplate? In case its not too bad we could change it to `Result` here.

Currently it seems you will just return `None` which will silently hide the corruption/incompatibility which is definitely not what we want. If the type is not `Result` this should panic instead (which will crash the runtime and result in the compute node voting "fail" for the batch).

kostko (commented):
Please make sure that https://github.com/oasisprotocol/oasis-bridge works with the proposed changes or that there is a draft PR which makes any necessary adjustments there.

nhynes:
bridge looks good https://github.com/oasisprotocol/oasis-bridge/pull/33

kostko (approved):

PR #78, opened 2021-04-01
Use once-cell

nhynes:
This PR replaces `lazy_static` with `once_cell::_::Lazy` because the latter is [cooler](https://github.com/rust-lang/rust/issues/74465) and the docs now recommend wrapping big genesis state in a `Lazy` because making all genesis state lazy [looks weird](https://github.com/oasisprotocol/oasis-sdk/commit/0670b7af26f0dfee8d8bae6536a850a23f1430da).


runtime-sdk/src/module.rs
```diff
@@ -123,27 +123,32 @@ impl AuthHandler for Tuple {
 /// Migration handler.
 pub trait MigrationHandler {
     /// Genesis state type.
+    ///
+    /// If this state is expensive to compute and not often updated, prefer
+    /// to make the genesis type something like `once_cell::unsync::Lazy<T>`.
```
kostko, line 128: This comment doesn't seem right as lazy is used in any case?

kostko (commented):
Nice!

kostko (commented):
Yeah it seems nicer without the automatic Lazy wrapping.

kostko (approved):

PR #87, opened 2021-04-07
ts-web/rt: add typed event visitor

pro-wh:
I saw a loose TODO in the bridge Go example client that we desired wrappers for working with events. Here's one for the typescript.


kostko (approved):

kostko:
Nice!

PR #88, opened 2021-04-07
ts-web/rt: v0.2.0-alpha1

pro-wh:
for use in https://github.com/oasisprotocol/oasis-bridge/pull/40


kostko (approved):

pro-wh:
I intend to go to 1.0.0 when we release, so chewing up these middle digits should be okay. This helps mark it as a breaking change.

pro-wh:
gonna select the event.Visitor addition as the version's spotlight change

PR #92, opened 2021-04-12
runtime-sdk: Consensus module

ptrus:
This PR:
- adds generalized consensus message handling support
  - Since consensus emitted messages get executed after the round finishes the results need to be handled in the next round. Modules can now register message hooks that get called at the start of the next round. When emitting a message one now needs to specify a handler to be called after the message is processed (also supports passing arbitrary handler context).
- adds low level `consensus`, and higher level `consensus-accounts` modules to the SDK
- add a general `Context` trait so that module methods can support both `TxContext` and `DispatchCtx`  
- adds e2e test to test the consensus modules

TODO:
- [x] clean the added message handler (just focused on getting it working) 
- [x] more consensus module stuff (staking txs, querying consensus state)
  - staking TXs requiers https://github.com/oasisprotocol/oasis-core/issues/3862 probably makes sense deffering into a separate PR
- [x] some tests

Only low-level staking support (emitting messages) is available currently. Adding a higher level support for staking (e.g. adding staking to `consensus-accounts` module) will be done later, as it requires https://github.com/oasisprotocol/oasis-core/issues/3862.

Requires: 
- [x] https://github.com/oasisprotocol/oasis-core/pull/3823
- [x] https://github.com/oasisprotocol/oasis-core/pull/3848
- [x] https://github.com/oasisprotocol/oasis-core/pull/3879


tests/run-bank.sh
```diff
@@ -0,0 +1,90 @@
...
+TAR=tar
+
+# ANSI escape codes to brighten up the output.
+RED=$'\e[31;1m'
+GRN=$'\e[32;1m'
+CYAN=$'\e[36;1m'
+OFF=$'\e[0m'
+
+# The base directory for all the node and test env cruft.
+TEST_BASE_DIR=$(mktemp -d -t oasis-sdk-e2e-XXXXXXXXXX)
```
nhynes, line 25:
```suggestion
TEST_BASE_DIR="$(mktemp -d -t oasis-sdk-e2e-XXXXXXXXXX)"
```

tests/run-bank.sh
```diff
@@ -0,0 +1,90 @@
...
+# Kill all dangling processes on exit.
+cleanup() {
+	printf "${OFF}"
+	pkill -P $$ || true
+	wait || true
+}
+trap "cleanup" EXIT
+
+# Find build tools.
+if [[ "$(which go)" == "" ]]; then
```
nhynes, line 36:
```suggestion
if ! command -v go > /dev/null; then
```
is more standard

tests/run-bank.sh
```diff
@@ -0,0 +1,90 @@
...
+	printf "${RED}### Please install 'go'.${OFF}\n"
+	exit 1
+fi
+if [[ "$(which cargo)" == "" ]]; then
+	printf "${RED}### Please install 'cargo'.${OFF}\n"
```
nhynes, line 41:
```bash
has_cmd() { command -v $1 > /dev/null 2>&1; }
die_must_install() {
    printf "${RED}### Please install $1.${OFF}\n"
    exit 1
}
if ! has_cmd cargo; then
    die_must_install "'cargo'"
fi
if ! has_cmd go; then
    die_must_install "'go'"
fi
if ! has_cmd wget; then
    if !has_cmd curl; then
        die_must_install "'curl' or 'wget'"
    fi
    # ...
fi
```
maybe more readable idk

tests/run-bank.sh
```diff
@@ -0,0 +1,90 @@
...
+	exit 1
+fi
+
+# Find a downloader tool.
+if [[ "$(which wget)" == "" ]]; then
```
nhynes, line 46:
```suggestion
if ! has_cmd wget; then
```

tests/run-bank.sh
```diff
@@ -0,0 +1,90 @@
...
+	if [[ "$(which curl)" == "" ]]; then
```
nhynes, line 47:
```suggestion
	if ! has_cmd curl; then
```

tests/run-bank.sh
```diff
@@ -0,0 +1,90 @@
...
+		exit 1
+	else
+		DOWNLOAD="curl --progress-bar --location -o"
+	fi
+else
+	DOWNLOAD="wget --quiet --show-progress --progress=bar:force:noscroll -O"
+fi
+
+printf "${CYAN}### Building test simple-bank runtime...${OFF}\n"
+cd "${ROOT}"/runtimes/simple-bank
```
nhynes, line 58: you don't need to `cd` since the examples are in the cargo workspace
```
cargo build -p test-runtime-simple-bank
```
and then the `target` dir is at the root

tests/run-bank.sh
```diff
@@ -0,0 +1,90 @@
...
+cargo build
+cp "${ROOT}"/../target/debug/test-runtime-simple-bank "${TEST_BASE_DIR}"/
```
nhynes, line 60:
```suggestion
cp "${ROOT}/../target/debug/test-runtime-simple-bank" "${TEST_BASE_DIR}/"
```

tests/run-bank.sh
```diff
@@ -0,0 +1,90 @@
...
+
```
nhynes, line 61:
```suggestion
print_status() { printf "${CYAN}### $1...${OFF}\n"; }
print_status "Building e2e test harness"
```
and so forth. separates style from content

tests/run-bank.sh
```diff
@@ -0,0 +1,90 @@
...
+${TAR} -xf oasis-core.tar.gz --strip-components=1 oasis_core_${OASIS_RELEASE}_linux_amd64/oasis-node
+
+printf "${CYAN}### Unpacking oasis-core-runtime-loader...${OFF}\n"
+# ${TAR} -xf oasis-core.tar.gz --strip-components=1 oasis_core_${OASIS_RELEASE}_linux_amd64/oasis-core-runtime-loader
+
+printf "${CYAN}### Running end-to-end tests...${OFF}\n"
+./e2e --log.level=INFO \
+    --log.format=json \
+    --basedir.no_cleanup \
+	--e2e.node.binary="/tmp/peter/oasis-node" \
```
nhynes, line 82: nb

tests/runtimes/simple-bank/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+version = "0.1.0"
+authors = ["Oasis Protocol Foundation <info@oasisprotocol.org>"]
+edition = "2018"
+license = "Apache-2.0"
+
+[dependencies]
+oasis-runtime-sdk = { path = "../../../runtime-sdk" }
+
+serde = { version = "1.0.118", features = ["derive"] }
+serde_bytes = "0.11.5"
```
nhynes, line 12: unused dep

tests/runtimes/simple-bank/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+thiserror = "1.0"
+once_cell = "1.7.2"
```
nhynes, line 14: unused dep

tests/runtimes/simple-bank/src/lib.rs
```diff
@@ -0,0 +1,21 @@
+//! Simple bank runtime.
+use oasis_runtime_sdk::{self as sdk, core::common::version::Version, modules};
+
+/// Simple bank runtime.
+pub struct Runtime;
+
+impl sdk::Runtime for Runtime {
+    const VERSION: Version = Version::new(0, 1, 0);
```
nhynes, line 8:
```suggestion
    const VERSION: Version = sdk::version_from_cargo!();
```

tests/runtimes/simple-bank/src/lib.rs
```diff
@@ -0,0 +1,21 @@
+//! Simple bank runtime.
+use oasis_runtime_sdk::{self as sdk, core::common::version::Version, modules};
```
nhynes, line 2:
```suggestion
use oasis_runtime_sdk::{self as sdk, core::common::version::Version, modules};
```
can you re-export `Version` from the sdk crate root? It's awful to type

tests/runtimes/simple-bank/src/lib.rs
```diff
@@ -0,0 +1,21 @@
...
+    const VERSION: Version = Version::new(0, 1, 0);
+
+    type Modules = (
+        modules::accounts::Module,
+        modules::consensus::Module<modules::accounts::Module>,
+    );
+
+    fn genesis_state() -> <Self::Modules as sdk::module::MigrationHandler>::Genesis {
+        (
+            modules::accounts::Genesis::default(),
```
nhynes, line 17: I think that you can replace the tuple with `Default::default()`?

runtime-sdk/src/context.rs
```diff
@@ -129,31 +140,54 @@ impl<'a> DispatchContext<'a> {
...
+            values: &mut self.values,
+        };
+        f(ctx)
+    }
 }
 
-/// Per-transaction dispatch context.
-pub struct TxContext<'a, 'b> {
+/// Per-transaction/method dispatch sub-context.
+pub struct Context<'a, 'b> {
```
nhynes, line 190: I'd save this name for `trait Context`

for instance
```rust
pub trait Context {
    fn runtime_header(&self) -> &Header;
    fn runtime_round_results(&self) -> &RoundResults;
    fn runtime_state(&mut self) -> impl Store;
    fn emit_message(&mut self, msg: Message) -> Result<u32, Error>;
    fn value<V>(&mut self, key: &'static str) -> &mut V;
    fn take_value<V>(&mut self, key: &'static str) -> Box<V>;
}
```
at the very least, it'd prevent `DispatchContext` and `TxContext` from diverging (`emit_messages` vs `emit_message` is already telling)

tests/run-bank.sh
ptrus: Yeah this is just a copy of the existing `run-e2e.sh` script hacked to work with the custom oasis-core i need, that i plan to remove :-)
 
You'll probably want to file these as an issue to clean up the`run-e2e.sh`?

runtime-sdk/src/types/message.rs
```diff
@@ -0,0 +1,12 @@
...
+use serde::{Deserialize, Serialize};
+
+use oasis_core_runtime::{common::cbor, consensus};
+
+/// Result of a message being processed by the consensus layer.
+pub type MessageEvent = consensus::roothash::MessageEvent;
+
+/// Handler name and context to be called after message is executed.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct MessageEventHook(pub String, pub cbor::Value);
```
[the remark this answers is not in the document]
ptrus: Agreed, also like the suggested names, thx :+1:

runtime-sdk/src/types/message.rs
```diff
@@ -0,0 +1,12 @@
+use std::fmt::Debug;
+
+use serde::{Deserialize, Serialize};
+
```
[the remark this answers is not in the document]
ptrus: I'm pretty sure the convention throughout this codebase is:
```
<std>

<external>

<first-party external>

<internal>
```

runtime-sdk/src/modules/consensus/types.rs
```diff
@@ -0,0 +1,51 @@
+//! Consensus module types.
+use crate::types::{address::Address, token};
+
+use serde::{Deserialize, Serialize};
+
+/// Deposit into runtime call.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct Deposit {
+    #[serde(rename = "amount")]
```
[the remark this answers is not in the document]
ptrus: I think we are explicitly setting these throughout the whole codebase.

runtime-sdk/src/module.rs
```diff
@@ -95,6 +98,48 @@ pub trait MethodRegistrationHandler {
...
 }
 
+/// Metadata of the consensus message handling callback.
+pub struct MessageHandlerInfo {
+    pub name: &'static str,
+    pub handler: fn(&MessageHandlerInfo, &mut Context<'_, '_>, MessageEvent, cbor::Value),
+}
+
+/// Registry of message handlers registered by the module.
+pub struct MessageHandlerRegistry {
```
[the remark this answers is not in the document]
ptrus: I see a lot of `Default` impls throughout the code that do the same as deriving it. It looked to me we want to be explicitly defining defaults? Not sure why though.

runtime-sdk/src/context.rs
ptrus: Yeah i also considered doing this, had some trouble getting the `runtime_state` lifetimes matching and all between the two existing implementations. But i don't think i tried returning the store as a trait object. Might do another go at it.

ptrus: I agree that generally having this as a trait is nicer. We can then have all of the module methods defined for the `Context` trait and don't need the added `with_ctx` method on the `DynamicContext`, and can just use it directly from `{begin,end}_block` and message hooks.

nhynes: I can take a whack at the lifetimes if you get stuck again. No guarantees that I can do any more than you, but at least we'll both feel more confident that it's impossible :)

ptrus: `impl Store` doesn't work as `impl Trait` is not allowed in trait methods.
Returning a trait object also doesn't work, as the `storage::Store` cannot be made into an object due to generic methods.

Using an associated type does look promising, but cannot get lifetimes to work: started with something like this
```
pub trait Context 
where
    Self::S: storage::Store,
{
    type S;

   fn runtime_state(&mut self) -> Self::S;
}

impl<'a, 'b> Context for TxContext<'a, 'b> 
{
    type S = &'b mut storage::OverlayStore<storage::MKVSStore<&'b mut &'a mut dyn mkvs::MKVS>>;

    fn runtime_state(&mut self) -> Self::S {
        self._runtime_state()
    }
}
```

The problem is that this `'b` lifetime in line `type S = &'b mut storage....` would actually need to be the elided lifetime from the `fn runtime_state<'this>(&mut self) -> Self::S`. Not sure if there's a way to specify that restriction in the `type S = ` line.

The workaround i came up with would be turning things around a bit:
```
pub trait Context 
where
    Self::S: storage::Store,
{
    type S;

   fn runtime_state(&mut self) -> &mut Self::S;
}

impl<'a, 'b> Context for TxContext<'a, 'b> 
{
    type S = storage::OverlayStore<storage::MKVSStore<&'b mut &'a mut dyn mkvs::MKVS>>;

    fn runtime_state(&mut self) -> &mut Self::S {
        self._runtime_state()
    }
}
```

This works, but then we would need to change something in the `DispatchContext` as the method there currently doesn't return a mutable reference: 
```
pub fn runtime_state(&mut self) -> storage::MKVSStore<&mut dyn mkvs::MKVS>
```
But it's not clear to me how, i guess i need to look up a bit more the storage implementations, which i tried to mostly ignore :-)

This looks like the best direction to go, or is there anything obvious i missed?

runtime-sdk/src/module.rs
```diff
@@ -95,6 +98,48 @@ pub trait MethodRegistrationHandler {
...
+    /// Create a new method registry.
+    #[allow(clippy::new_without_default)]
+    pub fn new() -> Self {
+        Self {
+            handlers: BTreeMap::new(),
+        }
+    }
+
+    /// Register message handlers exposed by the module.
+    pub fn register_handler(&mut self, mi: MessageHandlerInfo) {
```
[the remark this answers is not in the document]
ptrus: No, handlers should be registered up front. You can however register multiple handlers (with different names) for the same consensus event. When one emits an event he also needs to provide the handler name and payload - one can emit same consensus event with different callback handlers. 

However i don't think there's a use case for dynamically registering or changing registered handlers.

runtime-sdk/src/context.rs
ptrus: See last commit, it introduces the `Context` trait, makes things much nicer, however there's a todo left for implementing the `runtime_state` for the `DispatchContext`

runtime-sdk/src/context.rs
```diff
@@ -31,6 +31,89 @@ pub enum Mode {
     SimulateTx,
 }
 
+/// Runtime sdk context.
+pub trait Context
+where
+    Self::S: storage::Store,
+{
+    type S;
```
kostko, line 39: Can't you just use:
```suggestion
{
    type S: storage::Store;
```

runtime-sdk/src/module.rs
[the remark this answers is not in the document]
kostko: Yeah feel free to just replace them with deriving the `Default` trait, I don't remember a good reason for why I put those there explicitly.

runtime-sdk/src/runtime.rs
```diff
@@ -19,7 +20,11 @@ pub trait Runtime {
     /// Runtime version.
     const VERSION: version::Version;
 
-    type Modules: AuthHandler + MigrationHandler + MethodRegistrationHandler + BlockHandler;
+    type Modules: AuthHandler
```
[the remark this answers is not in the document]
kostko: > Is there a reason these are here and not trait bounds on `Module`. @kostko ?

It could be done either way and we always only use modules here so maybe we can move them (in a subsequent issue/PR of course).

runtime-sdk/src/context.rs
kostko: > But it's not clear to me how, i guess i need to look up a bit more the storage implementations, which i tried to mostly ignore :-)

This is because the `MKVSStore` is keeping a `&mut mkvs::MKVS` internally. It is basically a thin wrapper around a `mkvs::MKVS` that just allows you to avoid the IO context everywhere.

kostko: You could create one `storage::MKVSStore` instance in the constructor (instead of storing `&mut mkvs::MKVS` in `runtime_storage`) and then return a mut reference to it. Maybe?

ptrus: i think that worked! thanks!

runtime-sdk/src/modules/consensus/mod.rs
```diff
@@ -0,0 +1,204 @@
...
+            Message::Staking {
+                v: 0,
+                msg: StakingMessage::Transfer(staking::Transfer {
+                    to: addr.into(),
+                    amount: amount.amount().clone(),
+                }),
+            },
+            hook,
+        )
+        .map_err(|_| Error::InvalidArgument)
```
kostko, line 91: I think there should be a more specific error (defined in the core module) for "out of message slots" (similar to out of gas). Later with https://github.com/oasisprotocol/oasis-core/issues/3760 we can have the transaction scheduler make sure that transactions are scheduled such that they don't go over this limit.

runtime-sdk/src/modules/consensus/mod.rs
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    fn do_query_account<C: Context>(ctx: &C, addr: Address) -> Result<ConsensusAccount, Error> {
+        let state = StakingImmutableState {
+            mkvs: ctx.consensus_state(),
+        };
```
kostko, line 157: Maybe have a constructor so you don't need to write this each time.

runtime-sdk/src/modules/consensus/mod.rs
```diff
@@ -0,0 +1,204 @@
...
+}
+
+impl API for Module {
+    fn transfer<C: Context>(
+        ctx: &mut C,
+        to: Address,
+        amount: &token::BaseUnits,
+        hook: MessageEventHookInvocation,
+    ) -> Result<(), Error> {
+        Self::do_transfer(ctx, to, amount, hook)
```
kostko, line 171: Is there a reason for this indirection?

ptrus: no reason, will remove

runtime-sdk/src/modules/consensus/mod.rs
ptrus: Agreed. So emit_message already returned a similar error. I moved that error (and renamed it) to the core module. Also updated this consensus helper module to return module::core errors, so that the error can be used directly

ptrus: hm actually i still need a way to propagate this error in other modules (e.g. consensus_accounts module cannot return it currently cause it implements it's own error type...). Hm i guess module defined error enums should actually "extend" the core module errors?

runtime-sdk/src/modules/consensus_accounts/mod.rs
```diff
@@ -0,0 +1,422 @@
...
+            ctx,
+            from,
+            &amount,
+            MessageEventHookInvocation {
+                hook_name: CONSENSUS_WITHDRAW_HANDLER.to_string(),
+                payload: cbor::to_value(types::ConsensusWithdrawContext {
+                    address: from,
+                    amount: amount.clone(),
+                }),
+            },
```
kostko, line 147: Could add a helper (e.g. a generic method like `MessageEventHookInvocation::new(hook_name, payload)` where `payload` must implement the `Serialize` trait) for this so that you don't need to call `to_string` and `to_value` explicitly?

runtime-sdk/src/dispatcher.rs
```diff
@@ -61,15 +63,29 @@ pub struct Dispatcher<R: Runtime> {
...
     _runtime: PhantomData<R>,
 }
 
+const MODULE: &str = "dispatcher";
+
+/// State schema constants.
+pub mod state {
+    /// Map of message idx to message handlers for messages emitted in previous round.
+    pub const HANDLERS: &[u8] = &[0x01];
+}
```
kostko, line 78: We should probably move this under the `core` module as that one currently has a similar thing for module migration metadata.

runtime-sdk/src/dispatcher.rs
```diff
@@ -183,6 +199,65 @@ impl<R: Runtime> Dispatcher<R> {
...
+
+        for event in message_events {
+            let handler = handlers
+                .remove(&event.index)
+                .expect("message event hook should exist");
+            self.dispatch_message(ctx, handler.hook_name, event, handler.payload)
+                .expect("message event should be handled");
+        }
+        if !handlers.is_empty() {
+            panic!("unhandled message event")
```
kostko, line 242: Instead of panicking this could also propagate the error as a `RuntimError`? And you could use the logger from `oasis-core-runtime` (BTW, should we add a logger helper to `ctx` to be able to easily get the logger from anywhere?) to report which handlers remained.

runtime-sdk/src/dispatcher.rs
```diff
@@ -183,6 +199,65 @@ impl<R: Runtime> Dispatcher<R> {
...
+        let method_info = match self
+            .consensus_message_handlers
+            .lookup_handler(&handler_name)
+        {
+            Some(method_info) => method_info,
+            None => {
+                // Message handler not found.
+                return Err(modules::core::Error::InvalidMethod);
+            }
+        };
```
kostko, line 219: Could use `.ok_or(modules::core::Error::InvalidMethod)?`.

runtime-sdk/src/module.rs
```diff
@@ -95,6 +98,48 @@ pub trait MethodRegistrationHandler {
     }
 }
 
+/// Metadata of the consensus message handling callback.
+pub struct MessageHandlerInfo {
+    pub name: &'static str,
```
kostko, line 103: Add documentation comments for public fields.

runtime-sdk/src/modules/accounts/mod.rs
```diff
@@ -341,19 +311,19 @@ impl API for Module {
...
         Ok(types::AccountBalances {
-            balances: BTreeMap::from_iter(account.iter()),
+            balances: account.iter().collect(),
         })
     }
 }
 
 impl Module {
     fn tx_transfer(ctx: &mut TxContext<'_, '_>, body: types::Transfer) -> Result<(), Error> {
-        // Reject transfers when they are disabled.
```
kostko, line 351: Where did this check go? Also we should add a test for disabled transfers.

runtime-sdk/src/modules/consensus_accounts/mod.rs
```diff
@@ -0,0 +1,422 @@
...
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct Parameters {
+    pub consensus_denomination: token::Denomination,
+}
+
+impl Default for Parameters {
+    fn default() -> Self {
+        Self {
+            // TODO: get this from consensus? (token symbol?).
```
kostko, line 61: I think it makes sense to ensure that the consensus token is named consistently in all ParaTimes. On the other hand we also don't want to constantly query the consensus layer for this denomination as it is a pretty constant thing (it cannot actually be changed in the consensus layer unless there is a hard fork).

It could be queried during initialization (in `MigrationHandler`) and set as a parameter here?

runtime-sdk/src/modules/consensus_accounts/mod.rs
```diff
@@ -0,0 +1,422 @@
...
+        if ctx.is_check_only() {
+            return Ok(());
+        }
+
+        // Ensure correct denomination.
+        let params = Self::params(ctx.runtime_state());
+        let denomination = params.consensus_denomination;
+        if amount.denomination() != &denomination {
+            return Err(Error::InvalidDenomination);
+        }
```
kostko, line 129: Maybe we could factor this out as something like `Self::ensure_consensus_denomination(amount.denomination())?`.

runtime-sdk/src/modules/consensus_accounts/mod.rs
```diff
@@ -0,0 +1,422 @@
...
+            &amount,
+            MessageEventHookInvocation {
+                hook_name: CONSENSUS_WITHDRAW_HANDLER.to_string(),
+                payload: cbor::to_value(types::ConsensusWithdrawContext {
+                    address: from,
+                    amount: amount.clone(),
+                }),
+            },
+        )
+        .map_err(|err| err.into())
```
kostko, line 149: Is this needed or would a simple `?` work?

runtime-sdk/src/types/address.rs
```diff
@@ -241,4 +249,13 @@ mod test {
...
         ));
     }
+
+    #[test]
+    fn test_address_into_consensus_address() {
+        let pk = PublicKey::Secp256k1("Arra3R5V////////////////////////////////////".into());
+        let addr = Address::from_pk(&pk);
+        let consensus_addr: ConsensusAddress = addr.into();
+
+        assert_eq!(addr.to_bech32(), consensus_addr.to_bech32(),)
```
kostko, line 259:
```suggestion
        assert_eq!(addr.to_bech32(), consensus_addr.to_bech32())
```

runtime-sdk/src/modules/consensus_accounts/mod.rs
```diff
@@ -0,0 +1,422 @@
...
+    InvalidArgument,
+    #[error("invalid denomination")]
+    #[sdk_error(code = 2)]
+    InvalidDenomination,
+    #[error("withdraw: insufficient runtime balance")]
+    #[sdk_error(code = 3)]
+    InsufficientWithdrawBalance,
+    #[error("core error")]
+    #[sdk_error(code = 4)]
+    CoreError(#[from] modules::core::Error),
```
kostko, line 43: Maybe just `Core`?

runtime-sdk/src/modules/consensus_accounts/mod.rs
ptrus: Good idea, thanks!

runtime-sdk/src/modules/consensus_accounts/mod.rs
ptrus: Can do: `?; Ok(())`

runtime-sdk/src/modules/accounts/mod.rs
ptrus: fixed and added a test

runtime-sdk/src/modules/consensus_accounts/mod.rs
ptrus: Yeah agreed, Done, also moved the denoomination part into the `consesus` module.

runtime-sdk/src/dispatcher.rs
ptrus: Done

runtime-sdk/src/context.rs
```diff
@@ -62,47 +152,102 @@ impl<'a> DispatchContext<'a> {
             },
             runtime_header: ctx.header,
             runtime_round_results: ctx.round_results,
-            runtime_storage: mkvs,
+            runtime_storage: storage::MKVSStore::new(ctx.io_ctx.clone(), mkvs),
+            consensus_state: &ctx.consensus_state,
             io_ctx: ctx.io_ctx.clone(),
+            logger: get_logger("runtime-sdk").new(o!("ctx" => "dispatch")),
```
kostko, line 158: It would be cool if it also set the `mode` key to know what mode this is executing in.

runtime-sdk/src/context.rs
```diff
@@ -62,47 +152,102 @@ impl<'a> DispatchContext<'a> {
...
+}
 
-    /// Results of executing the last successful runtime round.
-    pub fn runtime_round_results(&self) -> &roothash::RoundResults {
-        self.runtime_round_results
+impl<'a> Context for DispatchContext<'a> {
+    type S = storage::MKVSStore<&'a mut dyn mkvs::MKVS>;
+
+    fn get_logger(&self, module: &'static str) -> slog::Logger {
+        self.logger.new(o!("sdk-module" => module))
```
kostko, line 197:
```suggestion
        self.logger.new(o!("sdk_module" => module))
```

runtime-sdk/src/context.rs
```diff
@@ -125,130 +267,96 @@ impl<'a> DispatchContext<'a> {
...
 
-impl<'a, 'b> TxContext<'a, 'b> {
-    /// Whether the transaction is just being checked for validity.
-    pub fn is_check_only(&self) -> bool {
-        self.mode == Mode::CheckTx
+impl<'a, 'b> Context for TxContext<'a, 'b> {
+    type S = storage::OverlayStore<&'b mut storage::MKVSStore<&'a mut dyn mkvs::MKVS>>;
+
+    fn get_logger(&self, module: &'static str) -> slog::Logger {
+        self.logger.new(o!("sdk-module" => module))
```
kostko, line 304:
```suggestion
        self.logger.new(o!("sdk_module" => module))
```

runtime-sdk/src/modules/consensus/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+    InvalidArgument,
+    #[error("invalid denomination")]
+    #[sdk_error(code = 2)]
+    InvalidDenomination,
+    #[error("core error")]
+    #[sdk_error(code = 3)]
+    Core(#[from] modules::core::Error),
+    #[error("internal state error")]
+    #[sdk_error(code = 4)]
+    InternalStateError(#[from] StateError),
```
kostko, line 72: Add blank lines between the variants.

runtime-sdk/src/modules/consensus/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct Parameters {
+    pub consensus_denomination: token::Denomination,
+}
+
+impl Default for Parameters {
+    fn default() -> Self {
+        Self {
+            consensus_denomination: token::Denomination::from_str("TEST").expect("to work"),
```
kostko, line 38:
```suggestion
            consensus_denomination: token::Denomination::from_str("TEST").unwrap(),
```

runtime-sdk/src/context.rs
```diff
@@ -31,21 +27,126 @@ pub enum Mode {
     SimulateTx,
 }
 
+impl Mode {
+    /// Mode string representation.
+    pub fn to_str(&self) -> &'static str {
```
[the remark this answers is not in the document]
ptrus: So afaik to use it in the logger context as i do bellow (so that it's present in all log invocations) it needs to be [threadsafe (`'static`, `Send`, `Sync`)](https://docs.rs/slog/2.7.0/slog/struct.Logger.html#method.new). 

So I'm not sure if the display trait implementation will be useful. However I should probably move these strings into constants and use these to also implement the `Display` trait and maybe rename this method to `to_static_str` with a comment on why we need it.

Unless i missed something obvious?

runtime-sdk/src/modules/consensus/mod.rs
```diff
@@ -0,0 +1,278 @@
...
+    #[error("invalid argument")]
+    #[sdk_error(code = 1)]
+    InvalidArgument,
+
+    #[error("invalid denomination")]
+    #[sdk_error(code = 2)]
+    InvalidDenomination,
+
+    #[error("core error")]
+    #[sdk_error(code = 3)]
```
kostko, line 70:
```suggestion
    #[sdk_error(transparent)]
```
(After rebase on master.)

pro-wh:
ok, I think I'll have the `tx_value` stuff in my branch similar to how you've done the `tx_auth_info` and `tx_caller_address`, with `Option`s

runtime-sdk/src/context.rs
```diff
@@ -54,55 +165,114 @@ pub struct DispatchContext<'a> {
...
+        self.block_tags.push(event.to_tag());
+    }
+
+    fn emit_message(
+        &mut self,
+        msg: roothash::Message,
+        hook: MessageEventHookInvocation,
+    ) -> Result<(), Error> {
         // Check against maximum number of messages that can be emitted per round.
         if self.messages.len() > self.max_messages as usize {
```
kostko, line 259: Since we are adding a new message this should actually be:
```suggestion
        if self.messages.len() >= self.max_messages as usize {
```
Also would be cool to have a test for this.

runtime-sdk/src/context.rs
```diff
@@ -125,130 +292,96 @@ impl<'a> DispatchContext<'a> {
...
-    /// Returns an index of the emitted message that can be used to correlate the corresponding
-    /// result after the message has been processed (in the next round).
-    pub fn emit_message(&mut self, msg: roothash::Message) -> Result<u32, Error> {
+    fn emit_message(
+        &mut self,
+        msg: roothash::Message,
+        hook: MessageEventHookInvocation,
+    ) -> Result<(), Error> {
         // Check against maximum number of messages that can be emitted per round.
         if self.messages.len() >= self.max_messages as usize {
```
kostko, line 366: This one seems correct, but we should still add a test :-)

runtime-sdk/src/modules/consensus/mod.rs
```diff
@@ -0,0 +1,279 @@
+//! Consensus module.
```
kostko, line 1: Maybe say something like "Module for communicating with the consensus layer."? The module could also be called `consensus_msgs` or something?

runtime-sdk/src/modules/consensus/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+    ) -> Result<(), Error>;
+
+    /// Returns consensus token denomination.
+    fn consensus_denomination<C: Context>(ctx: &mut C) -> Result<token::Denomination, Error>;
+
+    /// Query consensus account info.
+    fn account<C: Context>(ctx: &C, addr: Address) -> Result<ConsensusAccount, Error>;
+
+    // TODO:
+    //  - add more staking queries as needed.
```
kostko, line 119: Should we open an issue for these? Or just remove the comment?

runtime-sdk/src/modules/consensus/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+}
+
+/// Consensus module.
+///
+/// Low level consensus module supporting runtime interactions with the consensus layer.
+pub struct Module {}
```
kostko, line 125:
```suggestion
pub struct Module;
```

runtime-sdk/src/modules/consensus/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+
+    fn init_or_migrate(
+        ctx: &mut DispatchContext<'_>,
+        meta: &mut modules::core::types::Metadata,
+        genesis: &Self::Genesis,
+    ) -> bool {
+        let version = meta.versions.get(Self::NAME).copied().unwrap_or_default();
+        if version == 0 {
+            // TODO: enable loading consensus denomination from consensus state after:
+            // https://github.com/oasisprotocol/oasis-core/issues/3868
```
kostko, line 264: Consider opening an issue in the SDK repository as well.

runtime-sdk/src/modules/consensus_accounts/mod.rs
```diff
@@ -0,0 +1,383 @@
...
+
+#[cfg(test)]
+mod test;
+pub mod types;
+
+/// Unique module name.
+const MODULE_NAME: &str = "consensus_accounts";
+
+#[derive(Error, Debug, oasis_runtime_sdk_macros::Error)]
+pub enum Error {
```
kostko, line 29: Add empty lines between the variants.

runtime-sdk/src/modules/consensus_accounts/mod.rs
```diff
@@ -0,0 +1,383 @@
...
+    ) -> Result<types::AccountBalance, Error> {
+        let denomination = Consensus::consensus_denomination(ctx)?;
+        let balances = Accounts::get_balances(ctx.runtime_state(), args.addr)
+            .map_err(|_| Error::InvalidArgument)?;
+        let balance = balances
+            .balances
+            .get(&denomination)
+            .ok_or(Error::InvalidArgument)?;
+        Ok(types::AccountBalance {
+            balance: balance.clone(),
```
kostko, line 197: Why is clone needed here?

runtime-sdk/src/modules/consensus_accounts/mod.rs
```diff
@@ -0,0 +1,383 @@
...
+        })
+    }
+
+    fn query_consensus_account(
+        ctx: &mut DispatchContext<'_>,
+        args: types::ConsensusAccountQuery,
+    ) -> Result<ConsensusAccount, Error> {
+        Consensus::account(ctx, args.addr).map_err(|__| Error::InvalidArgument)
```
kostko, line 205:
```suggestion
        Consensus::account(ctx, args.addr).map_err(|_| Error::InvalidArgument)
```

runtime-sdk/src/modules/consensus_accounts/test.rs
```diff
@@ -0,0 +1,412 @@
...
+            "emitted message should match"
+        );
+
+        assert_eq!(
+            CONSENSUS_WITHDRAW_HANDLER.to_string(),
+            hook.hook_name,
+            "emitted hook should match"
+        )
+
+        // TODO: support advancing the round. And ensure message is correctly processed.
```
kostko, line 133: Would be useful to have support for this in the testing framework. Shouldn't be hard to add, maybe open an issue?

runtime-sdk/src/modules/consensus_accounts/types.rs
```diff
@@ -0,0 +1,59 @@
...
+pub struct AccountBalance {
+    #[serde(rename = "balance")]
+    pub balance: token::Quantity,
+}
+
+/// Context for consensus transfer message handler.
+#[derive(Clone, Debug, Serialize, Deserialize, Default)]
+#[serde(deny_unknown_fields)]
+pub struct ConsensusTransferContext {
+    pub address: Address,
```
kostko, line 49: Add explicit renames, also below.

runtime-sdk/src/types/address.rs
```diff
@@ -241,4 +249,13 @@ mod test {
             Error::MalformedAddress
         ));
     }
+
+    #[test]
+    fn test_address_into_consensus_address() {
+        let pk = PublicKey::Secp256k1("Arra3R5V////////////////////////////////////".into());
```
kostko, line 255: This seems like a bad example to test as the consensus layer doesn't support secp256k1 keys so such an address would be unusable.

runtime-sdk/src/modules/consensus_accounts/mod.rs
```diff
@@ -0,0 +1,383 @@
...
+impl<Accounts: modules::accounts::API, Consensus: modules::consensus::API>
+    Module<Accounts, Consensus>
+{
+    /// Deposit in the runtime.
+    fn tx_deposit(ctx: &mut TxContext<'_, '_>, body: types::Deposit) -> Result<(), Error> {
+        let signer = &ctx
+            .tx_auth_info()
+            .expect("should be called with a transaction ctx")
+            .signer_info[0];
+        let address = Address::from_pk(&signer.public_key);
```
kostko, line 168: This should make sure that the address is compatible with the consensus layer (e.g., is derived from an Ed25519 key).

runtime-sdk/src/modules/consensus_accounts/mod.rs
```diff
@@ -0,0 +1,383 @@
...
+    }
+
+    /// Withdraw from the runtime.
+    fn tx_withdraw(ctx: &mut TxContext<'_, '_>, body: types::Withdraw) -> Result<(), Error> {
+        // Signer.
+        let signer = &ctx
+            .tx_auth_info()
+            .expect("should be called with a transaction context")
+            .signer_info[0];
+        let address = Address::from_pk(&signer.public_key);
```
kostko, line 180: This should make sure that the address is compatible with the consensus layer (e.g., is derived from an Ed25519 key). Otherwise if someone withdraws to an address derived from an secp256k1 key, the funds would be stuck at the consensus layer.

runtime-sdk/src/modules/consensus_accounts/mod.rs
ptrus: Cause the `balance` returned from balances map is a reference. Hm i guess i could replace the `get` with `remove` since we don't actually mutate the underlying storage here.

runtime-sdk/src/modules/consensus/mod.rs
ptrus: Hm yeah a more detailed doc comment is on the module struct itself. I'll move it up here and update it a bit.

runtime-sdk/src/modules/consensus_accounts/mod.rs
ptrus: Nice spot, good idea yeah :-) Added `ensure_compatible_tx_signer` to the main consensus module API.

runtime-sdk/src/modules/consensus/mod.rs
ptrus: About the module name, i think just `consensus` makes more sense than `consensus_msgs` as the module also contains methods for querying consensus state (currently only `account` though), and some other helper methods such as `consensus_denomination` and `ensure_compatible_tx_signer`.

runtime-sdk/src/modules/consensus_accounts/test.rs
ptrus: Done https://github.com/oasisprotocol/oasis-sdk/issues/110

runtime-sdk/src/modules/consensus/mod.rs
ptrus: Done https://github.com/oasisprotocol/oasis-sdk/issues/111

runtime-sdk/src/modules/consensus/mod.rs
ptrus: Just removed the comment since it's obvious where new queries should be added if ever needed .

runtime-sdk/src/context.rs
ptrus: Opened issue https://github.com/oasisprotocol/oasis-sdk/issues/112

runtime-sdk/src/modules/consensus/mod.rs
kostko: Makes sense.

runtime-sdk/src/context.rs
```diff
@@ -31,21 +28,135 @@ pub enum Mode {
...
+    fn from(m: &Mode) -> Self {
+        match m {
+            Mode::CheckTx => MODE_CHECK_TX,
+            Mode::ExecuteTx => MODE_EXECUTE_TX,
+            Mode::SimulateTx => MODE_SIMULATE_TX,
+        }
+    }
+}
+
+/// Runtime sdk context.
```
kostko, line 51:
```suggestion
/// Runtime SDK context.
```

tests/e2e/simple_consensus.go
```diff
@@ -0,0 +1,243 @@
...
+		return e.Transfer.Amount.Cmp(&withdraw.Amount) == 0
+	}); err != nil {
+		return fmt.Errorf("ensuring alice withdraw consensus event: %w", err)
+	}
+
+	withdraw = types.BaseUnits{
+		Amount:       *quantity.NewFromUint64(50),
+		Denomination: types.Denomination("TEST"),
+	}
+	log.Info("charlie withdrawing")
```
kostko, line 210: Should also test with dave (secp256k1).

kostko (approved):

tests/e2e/simple_consensus.go
ptrus: good idea, done :+1:

PR #94, opened 2021-04-13
(unfinished) runtime-sdk: gas

### LINKED ISSUE #89: Gas model and accounting
We should add a simple gas model and corresponding accounting, likely similar to what we do in the consensus layer with each operation costing a specific amount of "gas". The gas limit and fee then implicitly define a gas price and nodes can decide to set a minimum price they will accept.

There is the question on how to estimate gas cost? Could be done as a special query like `core.EstimateGas`.

pro-wh:
fixes #89

the story so far:

- setting up a new `tx_values` field in TxContext similar to `values`, but it's transaction-local
- there's no batch-wide gas limit
- the core module manages a `core.GasUsed` tx_value with a u64 of how much gas the tx has used
- it provides a `use_gas` API function to increase that and check it against the tx auth info's fee's gas limit
- unit tests for that
- adding the skeleton for a `core.EstimateGas` query method, the idea from #89. dunno how to set up a new context inside a query yet tho


runtime-sdk/src/context.rs
```diff
@@ -177,6 +178,9 @@ pub struct TxContext<'a, 'b> {
 
     /// Per-context values.
     values: &'b mut BTreeMap<&'static str, Box<dyn Any>>,
+
+    /// Per-transaction values.
+    tx_values: BTreeMap<&'static str, Box<dyn Any>>,
```
nhynes, line 183: should this also be borrowed as `&'b mut`?

runtime-sdk/src/context.rs
```diff
@@ -271,6 +275,31 @@ impl<'a, 'b> TxContext<'a, 'b> {
             .map(|x| x.downcast().expect("type should stay the same"))
             .unwrap_or_default()
     }
+
+    /// Fetches or sets a value associated with the transaction.
+    pub fn tx_value<V>(&mut self, key: &'static str) -> &mut V
```
nhynes, line 280:
```suggestion
    pub fn tx_value<V>(&mut self, key: &str) -> &mut V
```
I think that the static requirement is a bit heavy, but w/e

runtime-sdk/src/context.rs
```diff
@@ -271,6 +275,31 @@ impl<'a, 'b> TxContext<'a, 'b> {
...
+        where
+            V: Any + Default,
+    {
+        self.tx_values
+            .entry(key)
+            .or_insert_with(|| Box::new(V::default()))
```
nhynes, line 286:
```suggestion
            .or_insert_with(Default::default)
```
should work

runtime-sdk/src/context.rs
```diff
@@ -271,6 +275,31 @@ impl<'a, 'b> TxContext<'a, 'b> {
...
+            .downcast_mut()
+            .expect("type should stay the same")
```
nhynes, line 288: I generally like to use `unwrap()` with a comment to prevent useless strings from filling up my `.text` section

runtime-sdk/src/modules/core/test.rs
```diff
@@ -0,0 +1,36 @@
+use crate::testing::mock;
+
+use super::{
+    API as _,
+    Module as Core,
+};
+
+#[test]
+fn test_use_gas() {
```
nhynes, line 9:
```suggestion
fn test_use_gas() {
    const MAX_GAS = 1000;
```
this constant is suggested for use below to make the code clearer due to fewer magic numbers

runtime-sdk/src/modules/core/test.rs
```diff
@@ -0,0 +1,36 @@
...
+    let mut mock = mock::Mock::default();
+    let mut ctx = mock.create_ctx();
+    let mut tx = mock::transaction();
+    tx.auth_info.fee.gas = 1000;
```
nhynes, line 13:
```suggestion
    tx.auth_info.fee.gas = MAX_GAS;
```

runtime-sdk/src/modules/core/test.rs
```diff
@@ -0,0 +1,36 @@
...
+
+    ctx.with_tx(tx.clone(), |mut tx_ctx, _call| {
+        Core::use_gas(&mut tx_ctx, 100)
```
nhynes, line 16:
```suggestion
        Core::use_gas(&mut tx_ctx, MAX_GAS)
```

runtime-sdk/src/modules/core/test.rs
```diff
@@ -0,0 +1,36 @@
...
+            .expect("using gas under limit should succeed");
+    });
+
+    ctx.with_tx(tx.clone(), |mut tx_ctx, _call| {
+        Core::use_gas(&mut tx_ctx, 950)
```
nhynes, line 21:
```suggestion
        Core::use_gas(&mut tx_ctx, MAX_GAS)
```

runtime-sdk/src/modules/core/test.rs
```diff
@@ -0,0 +1,36 @@
...
+            .expect("gas across separate transactions shouldn't accumulate");
+    });
+
+    ctx.with_tx(tx.clone(), |mut tx_ctx, _call| {
+        Core::use_gas(&mut tx_ctx, 600).unwrap();
```
nhynes, line 26:
```suggestion
        Core::use_gas(&mut tx_ctx, MAX_GAS).unwrap();
```

runtime-sdk/src/modules/core/test.rs
```diff
@@ -0,0 +1,36 @@
...
+        Core::use_gas(&mut tx_ctx, 600)
```
nhynes, line 27:
```suggestion
        Core::use_gas(&mut tx_ctx, 1)
```

runtime-sdk/src/context.rs
kostko: Oh, in this case feel free to also fix the existing stuff.

runtime-sdk/src/context.rs
kostko: No strong preference, just make it consistent also in existing code.

runtime-sdk/src/context.rs
pro-wh: this one is owned, with the lifecycle matching the context itself

runtime-sdk/src/context.rs
pro-wh: ok

runtime-sdk/src/context.rs
pro-wh: do we get good stack traces?

pro-wh: this seems to be prevalent in places this PR won't :crossed_fingers: touch. let's split this out into a separate issue

pro-wh: https://github.com/oasisprotocol/oasis-sdk/issues/99

runtime-sdk/src/context.rs
pro-wh: Oh you know what, it's inserting into `...Map<..., Box<dyn Any>>`, so all it knows is the `Any` in this context, so it doesn't know to use `V`

```
error[E0277]: the trait bound `(dyn std::any::Any + 'static): std::default::Default` is not satisfied
   --> runtime-sdk/src/context.rs:109:9
    |
109 | /         self.values
110 | |             .entry(key)
111 | |             .or_insert_with(Default::default)
    | |_____________________________________________^ the trait `std::default::Default` is not implemented for `(dyn std::any::Any + 'static)`
    | 
   ::: /home/wh0/.rustup/toolchains/nightly-2020-08-29-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/default.rs:84:1
    |
84  |   pub trait Default: Sized {
    |   ------------------------ required by this bound in `std::default::Default::default`
    |
    = note: required because of the requirements on the impl of `std::default::Default` for `std::boxed::Box<(dyn std::any::Any + 'static)>`

error[E0277]: the trait bound `dyn std::any::Any: std::default::Default` is not satisfied
   --> runtime-sdk/src/context.rs:111:29
    |
111 |             .or_insert_with(Default::default)
    |                             ^^^^^^^^^^^^^^^^ the trait `std::default::Default` is not implemented for `dyn std::any::Any`
    |
    = note: required because of the requirements on the impl of `std::default::Default` for `std::boxed::Box<dyn std::any::Any>`
    = note: required by `std::default::Default::default`
```

runtime-sdk/src/types/transaction.rs
```diff
@@ -148,6 +148,9 @@ pub enum CallResult {
 
         #[serde(rename = "code")]
```
pro-wh, line 149: why are these renamed?

pro-wh:
update: there's now a draft of estimate gas. this version puts only the method registry in the context.

runtime-sdk/src/types/transaction.rs
kostko: To make it explicit how the fields are called when serialized.

client-sdk/go/types/transaction.go
```diff
@@ -184,8 +184,9 @@ func (cr *CallResult) IsSuccess() bool {
 
 // FailedCallResult is a failed call result.
 type FailedCallResult struct {
-	Module string `json:"module"`
-	Code   uint32 `json:"code"`
+	Module  string `json:"module"`
+	Code    uint32 `json:"code"`
+	Message string `json:"message"`
```
kostko, line 189: The reason why there was no message field before was to reduce the output size. I guess we could at least make it optional (`omitempty` here).

runtime-sdk/src/types/transaction.rs
```diff
@@ -148,6 +148,9 @@ pub enum CallResult {
...
         code: u32,
+
+        #[serde(rename = "message")]
+        message: String,
```
kostko, line 153: Make it optional using `#[serde(skip_serializing_if = "String::is_empty")]` and `#[serde(default)]`.

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -30,10 +42,103 @@ pub enum Error {
...
+            mode: Mode::SimulateTx,
+            runtime_header: ctx.runtime_header,
+            runtime_round_results: ctx.runtime_round_results,
+            runtime_storage: ctx.runtime_storage,
+            io_ctx: ctx.io_ctx.clone(),
+            methods: ctx.methods,
+            max_messages: ctx.max_messages,
+            messages: Vec::new(),
+            values: BTreeMap::new(),
+        };
```
kostko, line 108: This should probably be factored out into a `with_simulation` method on the context.

kostko:
Nice!

> there's no batch-wide gas limit

I think we should support this. Would simply moving the gas accumulator from a per-tx value to a per-context value do the trick? I see that currently you are checking the tx gas limit directly in `use_gas` but maybe you could make the gas accumulator a struct with attributes like `batch_gas_limit` and `tx_gas_limit` (besides the `gas_used`) and then set the `tx_gas_limit` in the `AuthHandler` and `batch_gas_limit` in `BlockHandler`?

> adding message to CallResult::Failed. why did we discard this before? this will be more important now that modules will have to wrap errors from the core module for gas errors

Yeah the reason was to save on space.

I don't think that adding messages will solve the problem of wrapped errors as this still has issues if you want to match errors in a client application. I believe that the SDK should use [`source()`](https://doc.rust-lang.org/std/error/trait.Error.html#method.source) on the errors to resolve the most inner SDK error and use the module/code from that error.

kostko:
> I don't think that adding messages will solve the problem of wrapped errors as this still has issues if you want to match errors in a client application. I believe that the SDK should use [`source()`](https://doc.rust-lang.org/std/error/trait.Error.html#method.source) on the errors to resolve the most inner SDK error and use the module/code from that error.

See #100 for my attempt at implementing this.

runtime-sdk/src/types/transaction.rs
pro-wh: hm ok

client-sdk/go/types/transaction.go
pro-wh: ok

runtime-sdk/src/types/transaction.rs
pro-wh: ok

pro-wh:
> > I don't think that adding messages will solve the problem of wrapped errors as this still has issues if you want to match errors in a client application. I believe that the SDK should use [`source()`](https://doc.rust-lang.org/std/error/trait.Error.html#method.source) on the errors to resolve the most inner SDK error and use the module/code from that error.
> 
> See #100 for my attempt at implementing this.

cool, that should help too. 

> I don't think that adding messages will solve the problem of wrapped errors as this still has issues if you want to match errors in a client application. I believe that the SDK should use [`source()`](https://doc.rust-lang.org/std/error/trait.Error.html#method.source) on the errors to resolve the most inner SDK error and use the module/code from that error.

Message and source seem to serve different purposes. The change in this branch to message is to make propagate it outside of the runtime, for example to help _debug_ when working with runtime clients. The source attribute seems to be meant for handling specific conditions from within the runtime, to help _develop_ certain functionality. I've looked at #100, and the intention there seems to be to strip most of the information before it leaves the runtime.

runtime-sdk/src/types/transaction.rs
pro-wh: I should say: these often won't be empty in our usage, with (all?) errors in this project having non-empty displays

runtime-sdk/src/modules/core/mod.rs
pro-wh: ok

runtime-sdk/src/context.rs
pro-wh: could this apply to existing `values` related APIs too? we could open a separate issue if the scope is thus larger

pro-wh:
> I think we should support this. Would simply moving the gas accumulator from a per-tx value to a per-context value do the trick?

_copying_ it maybe. we did want per-tx gas limit as well, judging from the auth info struct

> I see that currently you are checking the tx gas limit directly in `use_gas` but maybe you could make the gas accumulator a struct with attributes like `batch_gas_limit` and `tx_gas_limit` (besides the `gas_used`)

I'll see if we can do something that doesn't duplicate the data. 

> and then set the `tx_gas_limit` in the `AuthHandler` and `batch_gas_limit` in `BlockHandler`?

what should we set the batch gas limit from? maybe some runtime parameter?

runtime-sdk/src/storage/overlay.rs
```diff
@@ -47,14 +47,17 @@ impl<S: Store> Store for OverlayStore<S> {
...
             .insert(key.as_ref().to_owned(), value.to_owned());
         self.dirty.insert(key.as_ref().to_owned());
+        last
     }
 
-    fn remove<K: AsRef<[u8]>>(&mut self, key: K) {
+    fn remove<K: AsRef<[u8]>>(&mut self, key: K) -> Option<Vec<u8>> {
         // For dirty values, remove from the overlay.
+        // TODO: Has this always been the case, where we diverge from what the Store trait defines
+        // as "remove" in the OverlayStore case?
```
pro-wh, line 60: I was going to see if it would be reasonable to make Store not throw away the last value in `insert` and `remove`, and I'm distracted by this `remove` implementation in OverlayStore. OverlayStore seems to change the meaning of `remove`. Here `remove` doesn't make future `get` return `None`. Here `remove` sort of resets it to the original value, which might still have been present.

kostko: No, this does not reset it to the original value as it only removes it from the overlay, but not from the dirty set. So further `get` queries will still return `None`.

kostko:
> _copying_ it maybe. we did want per-tx gas limit as well, judging from the auth info struct

No need to copy, as I mentioned we can have separate `batch_gas_limit` and `tx_gas_limit` in the accumulator and then the per-transaction limit can be updated in the auth handler. This works because we don't need to rollback gas accounting.

> what should we set the batch gas limit from? maybe some runtime parameter?

Yes it should be a runtime parameter that can be configured at genesis. Since you already made `core` a proper module, you could define some parameters and genesis for it.

runtime-sdk/src/context.rs
kostko: Yes this would apply to the existing APIs as well. No strong preference, but in practice these keys will have static lifetime as they will be defined as constants.

runtime-sdk/src/modules/accounts/mod.rs
```diff
@@ -75,12 +81,15 @@ pub enum Event {
 pub struct Parameters {
     #[serde(rename = "transfers_disabled")]
     pub transfers_disabled: bool,
+    #[serde(rename = "gas_cost_tx_transfer")]
+    pub gas_cost_tx_transfer: u64,
```
kostko, line 85: Maybe a separate "GasCosts" struct and then have "tx_transfer" as field?

pro-wh: :shrug: okay

runtime-sdk/src/storage/overlay.rs
pro-wh: oh right. None in overlay and key present in dirty means to remove on commit

pro-wh:
incorporated changes from main. trying out rebasing this time, but it's a little weird with older non-transparent-error now appearing to come after the introduction of transparent errors. anyway.

also added batch gas. still keeping tx_values though. I think the explicitness compared to managing value lifecycles implicitly during authenticate_tx, not to mention doing non-authentication work \:

runtime-sdk/src/modules/accounts/mod.rs
```diff
@@ -43,6 +45,14 @@ pub enum Error {
     #[error("forbidden by policy")]
     #[sdk_error(code = 3)]
     Forbidden,
+
+    #[error("core: {0}")]
+    #[sdk_error(transparent)]
+    Core(
+        #[from]
+        #[source]
```
kostko, line 53: Since `from` implies `source` you only need `from`.

runtime-sdk/src/modules/accounts/mod.rs
```diff
@@ -69,18 +79,29 @@ pub enum Event {
...
+    pub tx_transfer: u64,
+}
+
 /// Parameters for the accounts module.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub struct Parameters {
     #[serde(rename = "transfers_disabled")]
     pub transfers_disabled: bool,
+    #[serde(rename = "gas_cost")]
```
kostko, line 96:
```suggestion
    #[serde(rename = "gas_costs")]
```

kostko (commented):
> also added batch gas. still keeping tx_values though. I think the explicitness compared to managing value lifecycles implicitly during authenticate_tx, not to mention doing non-authentication work :

Looks good :+1:

runtime-sdk/src/modules/accounts/mod.rs
pro-wh: thx I'll try that

pro-wh: looks good

runtime-sdk/src/modules/accounts/mod.rs
pro-wh: ah yes, here and in the key value runtime

runtime-sdk/src/context.rs
pro-wh: thx. moving to #104

client-sdk/ts-web/rt/src/accounts.ts
```diff
@@ -12,6 +12,7 @@ export const MODULE_NAME = 'accounts';
 export const ERR_INVALID_ARGUMENT_CODE = 1;
 export const ERR_INSUFFICIENT_BALANCE_CODE = 2;
 export const ERR_FORBIDDEN_CODE = 3;
+export const ERR_CORE_CODE = 4;
```
pro-wh, line 15: oops, ts part needs more updates

pro-wh:
something's wrong with the genesis an parameter initialization. it keeps seeing batch_gas=0. how do I set up the core module parameters?

pro-wh:
it's because I was doing a query before the first transaction. I just found out that we don't call init_or_migrate until the first batch comes in, so we won't have a batch gas limit transferred from the genesis parameters yet ):

kostko:
> it's because I was doing a query before the first transaction. I just found out that we don't call init_or_migrate until the first batch comes in, so we won't have a batch gas limit transferred from the genesis parameters yet ):

That's true. We could call this for queries as well and we could also make executor nodes always propose a block on genesis (so that any migration can happen as soon as possible).

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -30,10 +40,206 @@ pub enum Error {
...
+
+/// Genesis state for the accounts module.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct Genesis {
+    #[serde(rename = "parameters")]
+    pub parameters: Parameters,
+}
+
+impl Default for Genesis {
```
kostko, line 96: Feel free to derive `Default`, #92 changes a bunch of these to use default derivation instead of explicitly defining these.

runtime-sdk/src/context.rs
```diff
@@ -147,9 +156,31 @@ impl<'a> DispatchContext<'a> {
...
+    /// values.
+    pub fn with_simulation<F, R>(&mut self, f: F) -> R
+    where
+        F: FnOnce(&mut DispatchContext<'_>) -> R,
+    {
+        let mut sim_ctx = DispatchContext {
+            mode: Mode::SimulateTx,
+            runtime_header: self.runtime_header,
+            runtime_round_results: self.runtime_round_results,
+            runtime_storage: self.runtime_storage,
```
kostko, line 175: Note that this way anything in the simulation context could still overwrite storage if it commits something. I think it is enough to mention this in the method documentation.

runtime-sdk/src/context.rs
```diff
@@ -271,6 +305,31 @@ impl<'a, 'b> TxContext<'a, 'b> {
...
+            .entry(key)
+            .or_insert_with(|| Box::new(V::default()))
+            .downcast_mut()
+            .expect("type should stay the same")
+    }
+
+    /// Takes a value associated with the transaction.
+    ///
+    /// The previous value is removed so subsequent fetches will return the default value.
+    pub fn take_tx_value<V>(&mut self, key: &'static str) -> Box<V>
```
kostko, line 324: No tests for this one?

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -30,10 +40,206 @@ pub enum Error {
...
+    #[sdk_error(code = 10)]
+    BatchOutOfGas,
+}
+
+/// Parameters for the accounts module.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct Parameters {
+    #[serde(rename = "batch_gas")]
+    pub batch_gas: u64,
```
kostko, line 70: Should this be `max_batch_gas`?

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -30,10 +40,206 @@ pub enum Error {
...
+}
+
+impl module::Parameters for Parameters {
+    type Error = std::convert::Infallible;
+}
+
+pub trait API {
+    fn use_gas(ctx: &mut TxContext<'_, '_>, gas: u64) -> Result<(), Error>;
+    fn batch_use_gas(ctx: &mut DispatchContext<'_>, gas: u64) -> Result<(), Error>;
+}
```
kostko, line 86: Add documentation comments.

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -30,10 +40,206 @@ pub enum Error {
...
+    #[error("batch gas overflow")]
+    #[sdk_error(code = 9)]
+    BatchGasOverflow,
+
+    #[error("batch out of gas")]
+    #[sdk_error(code = 10)]
+    BatchOutOfGas,
+}
+
+/// Parameters for the accounts module.
```
kostko, line 65:
```suggestion
/// Parameters for the core module.
```

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -30,10 +40,206 @@ pub enum Error {
...
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct Parameters {
+    #[serde(rename = "batch_gas")]
+    pub batch_gas: u64,
+}
+
+impl Default for Parameters {
```
kostko, line 73: Feel free to derive `Default` if there's nothing non-default.

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -30,10 +40,206 @@ pub enum Error {
...
+            return Err(Error::BatchOutOfGas);
+        }
+        *batch_gas_used = batch_new_gas_used;
+
+        Ok(())
+    }
+}
+
+impl Module {
+    fn query_estimate_gas(
```
kostko, line 175: Add documentation comment.

runtime-sdk/src/modules/core/mod.rs
pro-wh: ok

runtime-sdk/src/modules/core/mod.rs
pro-wh: I'm fine with max_batch_gas. that would have been my second choice after batch_gas (the latter for consistency with `gas` in fee)

runtime-sdk/src/context.rs
pro-wh: would be better to have an overlay

kostko: Yeah it would be if you can make it so easily.

pro-wh: oh right I can't. I just realized there's no overlay defined for wrapping the panicky kind of MKVS. comment it is!

pro-wh: https://github.com/oasisprotocol/oasis-core/issues/3904

runtime-sdk/src/context.rs
pro-wh: adding

runtime-sdk/src/modules/core/mod.rs
pro-wh: converted from #92

pro-wh:
merged with the context changes. I haven't gone and made a struct SimulationContext that implements Context. it turned out to be simpler to make a with_tx_simulation function. we'll revisit this later if we ever need a non-transaction-specific simulation context.

I tried writing use_gas in a new way with the optional transaction info, but we can't access the auth info and the values at the same time. thanks, rust

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -42,6 +52,64 @@ pub enum Error {
...
+
+/// Parameters for the core module.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct Parameters {
+    #[serde(rename = "batch_gas")]
+    pub max_batch_gas: u64,
+}
+
+impl Default for Parameters {
```
kostko, line 85: This was resolved but it still has the custom Default implementation? Also below.

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -51,3 +119,135 @@ pub mod state {
...
+    fn migrate(_ctx: &mut DispatchContext<'_>, _from: u32) -> bool {
+        // No migrations currently supported.
+        false
+    }
+}
+
+impl API for Module {
+    fn use_gas<C: Context>(ctx: &mut C, gas: u64) -> Result<(), Error> {
+        let new_gas_used = match (
+            // todo: oh shoot we can't borrow for both of these at the same time
```
kostko, line 144: Yeah you'll need to refactor this. Probably just do something like `ctx.tx_auth_info().map(|ai| ai.fee.gas)` to avoid holding a reference to `ctx` just for the gas limit specified in the transaction.

runtime-sdk/src/modules/core/mod.rs
pro-wh: woa let me check

pro-wh: oh oops I must have gotten confused with similar changes in the account module. of course other PRs would not have touched the new code in core

runtime-sdk/src/modules/core/mod.rs
pro-wh: yup

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -42,6 +52,50 @@ pub enum Error {
...
+    #[error("batch out of gas")]
+    #[sdk_error(code = 14)]
+    BatchOutOfGas,
+}
+
+/// Parameters for the core module.
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct Parameters {
+    #[serde(rename = "batch_gas")]
```
kostko, line 81:
```suggestion
    #[serde(rename = "max_batch_gas")]
```

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -42,6 +52,50 @@ pub enum Error {
...
+    pub max_batch_gas: u64,
+}
+
+impl module::Parameters for Parameters {
+    type Error = std::convert::Infallible;
+}
+
+pub trait API {
+    fn use_gas<C: Context>(ctx: &mut C, gas: u64) -> Result<(), Error>;
+}
```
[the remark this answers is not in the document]
pro-wh: ah yeah thanks for resurfacing this. I'll close out the old one

runtime-sdk/src/modules/core/mod.rs
pro-wh: moving to new thread

runtime-sdk/src/modules/core/mod.rs
pro-wh: bumping

pro-wh:
- **we now run migration for queries too**
- finished fixing up smaller conflicts with main
- added doc comments
- incorporated other smaller review comments

kostko (approved):

PR #97, opened 2021-04-19
e2e: Add tests for accounts

abukosek:
Part of #21.  Adds accounts to the simple-keyvalue runtime and two new end-to-end tests for accounts, also correctly handles account nonces in `kvInsert` and `kvRemove`.

### DIFF
--- a/tests/e2e/runtime.go
+++ b/tests/e2e/runtime.go
@@ -129,15 +129,15 @@ func (sc *runtimeScenario) Fixture() (*oasis.NetworkFixture, error) {
 				Executor: registry.ExecutorParameters{
 					GroupSize:       2,
 					GroupBackupSize: 1,
-					RoundTimeout:    20,
-					MaxMessages:     128,
+					RoundTimeout:    30,
+					MaxMessages:     256,
 				},
 				TxnScheduler: registry.TxnSchedulerParameters{
 					Algorithm:         registry.TxnSchedulerSimple,
-					MaxBatchSize:      1,
-					MaxBatchSizeBytes: 1024,
+					MaxBatchSize:      10,
+					MaxBatchSizeBytes: 16 * 1024 * 1024, // 16 MB.
 					BatchFlushTimeout: 1 * time.Second,
-					ProposerTimeout:   20,
+					ProposerTimeout:   30,
 				},
 				Storage: registry.StorageParameters{
 					GroupSize:               2,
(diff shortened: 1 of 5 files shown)


abukosek:
The `ts-web-rt` CI test fails with:
```
client: transaction check failed: runtime: check tx failed: runtime error: module: core code: 4 message:
  [object Object]
  From Your Spec Code:
      at Context.eval (http://localhost:42280/__cypress/tests?p=cypress/integration/playground.spec.js:106:17)
```
The error message points to line 106 in an 11 line file and the error message is `[object Object]`, so that's not very helpful.
Error code 4 is "invalid nonce".

Could someone with more knowledge of what that test should be doing point me at what to fix?

kostko:
Take a look at this:
* https://github.com/oasisprotocol/oasis-sdk/blob/main/client-sdk/ts-web/rt/playground/src/index.js

My guess would be that these transactions are missing proper nonces:
* https://github.com/oasisprotocol/oasis-sdk/blob/main/client-sdk/ts-web/rt/playground/src/index.js#L149-L152

See here for how to query for nonces:
* https://github.com/oasisprotocol/oasis-bridge/blob/master/client-sdk/ts-web/playground/src/index.js#L100-L106

abukosek:
Thanks, I'll have a look! :)

tests/e2e/runtime.go
```diff
@@ -129,15 +129,15 @@ func (sc *runtimeScenario) Fixture() (*oasis.NetworkFixture, error) {
...
 				},
 				TxnScheduler: registry.TxnSchedulerParameters{
 					Algorithm:         registry.TxnSchedulerSimple,
-					MaxBatchSize:      1,
-					MaxBatchSizeBytes: 1024,
-					BatchFlushTimeout: 1 * time.Second,
-					ProposerTimeout:   20,
+					MaxBatchSize:      10,
+					MaxBatchSizeBytes: 16 * 1024 * 1024, // 16 MB.
+					BatchFlushTimeout: 20 * time.Second,
```
kostko, line 139: Why did you increase the flush timeout, this can delay tests? I would leave it as 1 second.

abukosek: It seemed a bit flaky with shorter timeouts, I'll try to change this back to 1s.

abukosek: OK, it seems to work fine with 1s too :man_shrugging:

pro-wh:
looks good from the js side.

> The error message points to line 106 in an 11 line file and the error message is `[object Object]`, so that's not very helpful.
> Error code 4 is "invalid nonce".

worse, message was blank. `[object Object]` is the next line. it might be the `metadata` dictionary from grpc, which we don't use

kostko (approved):

PR #100, opened 2021-04-21
runtime-sdk: Properly handle error chains

kostko:
SDK errors can now implement the source method which is taken into account when
converting an error into a call result. This makes it possible to nest errors
and make sure that module and code from the most inner error are used.

Seems like it could be useful in #92 and #94 where we started to nest errors.

### DIFF
--- a/runtime-sdk-macros/src/error_derive.rs
+++ b/runtime-sdk-macros/src/error_derive.rs
@@ -1,7 +1,7 @@
-use darling::{util::Flag, FromDeriveInput, FromVariant};
+use darling::{util::Flag, FromDeriveInput, FromField, FromVariant};
 use proc_macro2::TokenStream;
-use quote::{format_ident, quote};
-use syn::{DeriveInput, Ident};
+use quote::{format_ident, quote, quote_spanned};
+use syn::{DeriveInput, Ident, Index, Member, Path};
 
 use crate::generators::{self as gen, CodedVariant};
 
@@ -28,9 +28,14 @@ struct Error {
 struct ErrorVariant {
     ident: Ident,
 
+    fields: darling::ast::Fields<ErrorField>,
+
     /// The explicit ID of the error code. Overrides any autonumber set on the error enum.
     #[darling(default, rename = "code")]
     code: Option<u32>,
+
+    #[darling(default, rename = "transparent")]
+    transparent: Flag,
 }
 
 impl CodedVariant for ErrorVariant {
@@ -45,6 +50,14 @@ impl CodedVariant for ErrorVariant {
     }
 }
 
+#[derive(FromField)]
+#[darling(forward_attrs(source, from))]
+struct ErrorField {
+    ident: Option<Ident>,
+
+    attrs: Vec<syn::Attribute>,
+}
+
 pub fn derive_error(input: DeriveInput) -> TokenStream {
     let error = match Error::from_derive_input(&input) {
         Ok(error) => error,
@@ -53,17 +66,18 @@ pub fn derive_error(input: DeriveInput) -> TokenStream {
 
     let error_ty_ident = &error.ident;
 
-    let code_converter = gen::enum_code_converter(
+    let module_name = error
+        .module_name
+        .unwrap_or_else(|| syn::parse_quote!(MODULE_NAME));
+
+    let (module_name_body, code_body) = convert_variants(
         &format_ident!("self"),
+        module_name,
         &error.data.as_ref().take_enum().unwrap(),
         error.autonumber.is_some(),
     );
 
     let sdk_crate = gen::sdk_crate_path();
-
-    let module_name = error
-        .module_name
-        .unwrap_or_else(|| syn::parse_quote!(MODULE_NAME));
 
     gen::wrap_in_const(quote! {
         use #sdk_crate::{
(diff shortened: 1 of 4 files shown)


runtime-sdk-macros/src/error_derive.rs
```diff
@@ -68,37 +83,96 @@ pub fn derive_error(input: DeriveInput) -> TokenStream {
...
-                RuntimeError::new(#error_ty_ident::module_name(), err.code(), &err.to_string())
+                RuntimeError::new(err.module_name(), err.code(), &err.to_string())
             }
         }
     })
 }
 
+fn convert_error_sources(enum_binding: &Ident, variants: &[&ErrorVariant]) -> TokenStream {
+    if variants.is_empty() {
+        return quote!(std::option::Option::None);
```
nhynes, line 113: what kind of monster redefines `None`? This line is 100% a good idea, but I shudder to think...

runtime-sdk-macros/src/error_derive.rs
```diff
@@ -68,37 +83,96 @@ pub fn derive_error(input: DeriveInput) -> TokenStream {
...
+    }
+
+    let match_arms = variants.iter().map(|variant| {
+        let variant_ident = &variant.ident;
+        let mut maybe_sources = variant
+            .fields
+            .iter()
+            .enumerate()
+            .filter(|(_, f)| !f.attrs.is_empty())
+            .map(|(field_index, f)| (field_index, f.ident.clone()));
```
nhynes, line 123:
```suggestion
            .filter_map(|(i, f)| f.attrs.is_empty().then(|| (i, &f.ident)))
```
should also work, I think

runtime-sdk-macros/src/error_derive.rs
```diff
@@ -170,4 +257,56 @@ mod tests {
...
+        );
+
+        let input: syn::DeriveInput = syn::parse_quote!(
+            #[derive(Error)]
+            #[sdk_error(module_name = "THE_MODULE_NAME")]
+            pub enum Error {
+                #[sdk_error(code = 1)]
+                Foo(
+                    #[from]
+                    #[sdk_source]
```
nhynes, line 302: It might be cool to use the same helper attrs as thiserror, but it'd require a lot of work not to be in an uncanny valley of feature parity.

runtime-sdk/src/error.rs
```diff
@@ -29,26 +29,171 @@ use crate::types::transaction::CallResult;
 /// ```
 pub trait Error: std::error::Error {
     /// Name of the module that emitted the error.
-    fn module_name() -> &'static str;
+    fn module_name(&self) -> &'static str;
```
nhynes, line 32: Yeah, I agree that the other syntax is a bit awkward. It's probably safe to assume that nobody will ever work with an error independently of an instance (c.f. marker types).

runtime-sdk/src/error.rs
```diff
@@ -29,26 +29,171 @@ use crate::types::transaction::CallResult;
...
+#[doc(hidden)]
+pub mod private {
+    use super::Error;
+
+    pub trait AsDynError {
+        fn as_dyn_error(&self) -> &(dyn Error + 'static);
+    }
+
+    impl<T: Error + 'static> AsDynError for T {
+        #[inline]
```
nhynes, line 74: I would hope that everyone uses LTO. The kind of person who doesn't use LTO on release builds is the same one who declares `struct None`.

runtime-sdk/src/error.rs
```diff
@@ -29,26 +29,171 @@ use crate::types::transaction::CallResult;
...
+    }
+
+    #[derive(thiserror::Error, Debug, oasis_runtime_sdk_macros::Error)]
+    #[sdk_error(module_name = "MODULE_NAME_3")]
+    enum ParentParentError {
+        #[error("nested nested error")]
+        #[sdk_error(code = 1)]
+        Nested(
+            #[source]
+            #[sdk_source]
```
nhynes, line 131: So this actually means something different from _source_. It's actually marking the error as _transparent_. The error code isn't actually needed when setting the source either but is unfortunately required by a sane implementation of `code()`.

How about something like:
```
enum Error {
    #[sdk_error(code = 1, forward_to = ".0")
    Nested(#[source] ParentError)
}
```
where `foward_to` is a field of the variant

kostko: Yeah this makes sense, it's just that the syntax seems uglier than annotating the field.

runtime-sdk-macros/src/error_derive.rs
kostko: He he

runtime-sdk/src/error.rs
kostko: Actually, it's not really forwarding everything since you can still get the whole chain by following `source()`. It's just that the default implementation of `to_call_result` will try to find the inner-most error.

kostko: But maybe we can make it forward everything :-)

What if we call it `#[sdk_error(transparent)]` (without the code) and then it will forward (actually forward module, code and everything) the field that has the `#[source]` or `#[from]` attributes?

runtime-sdk/src/error.rs
```diff
@@ -29,26 +29,171 @@ use crate::types::transaction::CallResult;
...
 
     /// Error code uniquely identifying the error.
     fn code(&self) -> u32;
 
     /// Converts the error into a call result.
     fn to_call_result(&self) -> CallResult {
-        CallResult::Failed {
-            module: Self::module_name().to_owned(),
-            code: self.code(),
+        // If there is an underlying error, use that one.
```
pro-wh, line 39: why is it more correct to discard the outer error?

kostko: I'll change this so it will be up to the error definition to decide which one to use. See my reply about `transparent` above. Thoughts?

runtime-sdk/src/error.rs
nhynes: I like it. I didn't want to suggest your writing more annotations code, though. I wonder whether thiserror shouldn't have collapsed it onto the field attr, though.

runtime-sdk/src/error.rs
pro-wh: ok moving discussion there

runtime-sdk/src/error.rs
pro-wh: transparent is a good convenience which I think will be good in some cases like gas accounting. but I'm worried that it won't be general enough. would there be cases where we do want context from the outer error?

if there were, for example a method that did two transfers atomically, we might desire to have errors like `two_transfers_module::Error::SecondTransferFail(source=accounts::Error::InsufficientBalance)`

kostko: Added the `#[sdk_error(transparent)]` option, please take another look.

> if there were, for example a method that did two transfers atomically, we might desire to have errors like

Yeah but I don't see how we could fit this into a single module/code?

nhynes (approved):

runtime-sdk-macros/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![feature(proc_macro_diagnostic)]
+#![feature(bool_to_option)]
```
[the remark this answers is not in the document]
kostko: Yeah could be, we can bump it.

pro-wh:
thanks, there's a lot I don't get about why model errors this way, but transparent will be convenient

kostko:
> thanks, there's a lot I don't get about why model errors this way, but transparent will be convenient

Any recommendations on how to make this better? So you are saying that the module/code representation of errors is not the best?

pro-wh:
> Any recommendations on how to make this better? So you are saying that the module/code representation of errors is not the best?

Not about the module/code representation, but the type system that segregates error values by module, rather than by qualities like the shape of data they store or by what operations they support. If there were a "basic" error type like this

```rust
struct ModuleCode {
  module: &'static str,
  code: u32,
  friendly_name: &'static str,
}
struct BasicError {
  module_code: &'static ModuleCode,
  // perhaps some instance info: source, owned string message, etc
}
// impls for BasicError
```

We wouldn't need to wrap errors for transparent forwarding. There may be some "but you can't ___" with this idea too.

One idea for detailed a error type system is to represent statically what can go wrong, but I don't think we're currently getting good use out of that by having an enum for an entire module's worth of errors.

PR #101, opened 2021-04-22
client-sdk/go/crypto: Implement secp256k1 signer

### LINKED ISSUE #95: Implement missing secp256k1 signer
Currently the secp256k1 signer implementation in the [Go client SDK](https://github.com/oasisprotocol/oasis-sdk/tree/main/client-sdk/go/crypto/signature/secp256k1) is missing and should be filled in. We could use the `github.com/btcsuite/btcd/btcec` library (unless there are any better alternatives).

abukosek:
Closes #95.

### DIFF
--- a/client-sdk/go/testing/testing.go
+++ b/client-sdk/go/testing/testing.go
@@ -1,10 +1,13 @@
 package testing
 
 import (
+	"crypto/sha512"
+
 	memorySigner "github.com/oasisprotocol/oasis-core/go/common/crypto/signature/signers/memory"
 
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/crypto/signature"
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/crypto/signature/ed25519"
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/crypto/signature/secp256k1"
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/types"
 )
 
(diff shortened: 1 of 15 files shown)


kostko (commented):
Would be cool to have an E2E test where we use a secp256k1 account to sign a transaction. This means you would want to add a test key or three based on secp256k1.

abukosek:
OK, I'll add that :)

pro-wh:
do we have any external things to test compatibility against? I have a draft of similar signing/verifying in the typescript lib, so if you have something, I'd like to test the typescript part against that as well.

abukosek:
The library I'm using (`btcec`) already has extensive tests, so I didn't feel the need to repeat them here, since this PR basically just wraps their library.  I do test if our code properly generates a signature that's in their examples, but that's just a quick sanity check.
If you find any offical test vectors, we can add those in a separate PR, just to be extra sure.

client-sdk/go/testing/testing.go
```diff
@@ -22,11 +25,22 @@ func newTestKey(seed string) TestKey {
 	}
 }
 
+func newSecp256k1TestKey(seed string) TestKey {
```
kostko, line 28: Rename the other one to `newEd25519TestKey`?

runtime-sdk/Cargo.toml
```diff
@@ -13,6 +13,7 @@ oasis-runtime-sdk-macros = { path = "../runtime-sdk-macros", optional = true }
 serde = { version = "1.0.118", features = ["derive"] }
 serde_bytes = "0.11.5"
 ed25519-dalek = { version = "1.0.1", features = ["serde"] }
+sha2 = "0.9.3"
 k256 = { version = "0.6.0", features = ["ecdsa", "sha256"] }
```
kostko, line 17: Remove the `sha256` feature now?

runtime-sdk/src/testing/keys.rs
```diff
@@ -36,6 +36,41 @@ macro_rules! test_key {
     }
 }
 
+/// Define a secp256k1 test key.
+macro_rules! test_key_secp256k1 {
```
kostko, line 40: Rename the other one to `test_key_ed25519`.

kostko (approved):

abukosek:
Thanks!

Issue #113, opened 2021-04-29
Add consensus_accounts ts/go clients


ptrus:
Add ts/go clients for the consensus_accounts module added in https://github.com/oasisprotocol/oasis-sdk/pull/92

PR #115, opened 2021-04-30
consensus sdk clients

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-sdk/issues/113
TODO:
- [x] ts-web consensus-sdk client tests
- [x] requires https://github.com/oasisprotocol/oasis-core/pull/3914

### DIFF
--- a/client-sdk/ts-web/rt/cypress/integration/playground.spec.js
+++ b/client-sdk/ts-web/rt/cypress/integration/playground.spec.js
@@ -1,11 +1,27 @@
 /// <reference types="cypress" />
 
 describe('playground', () => {
-    it('finishes', () => {
+    it('finishes simple-keyvalue', () => {
         cy.visit('http://localhost:42280/');
         // This is similar to `.its('playground')`, except that
         // (i) it doesn't retry if `w.playground` rejects, and
         // (ii) it passes when `w.playground` fulfills with `undefined`.
-        cy.window().then({timeout: 120_000}, (w) => w.playground);
+        cy.window()
+            .then({timeout: 120_000}, (w) => w.playground)
+            .then((w) => {
+                expect(w.playground).to.be.ok;
+            });
+    });
+
+    it('finishes simple-consensus', () => {
+        cy.visit('http://localhost:42280/consensus.html');
+        // This is similar to `.its('playground')`, except that
+        // (i) it doesn't retry if `w.playground` rejects, and
+        // (ii) it passes when `w.playground` fulfills with `undefined`.
+        cy.window()
+            .then({timeout: 120_000}, (w) => w.playground)
+            .then((w) => {
+                expect(w.playground).to.be.ok;
+            });
     });
 });
--- a/client-sdk/ts-web/rt/playground/src/consensus.js
+++ b/client-sdk/ts-web/rt/playground/src/consensus.js
@@ -1,0 +1,118 @@
+// @ts-check
+
+import * as oasis from '@oasisprotocol/client';
+
+import * as oasisRT from './../..';
+
+const CONSENSUS_RT_ID = oasis.misc.fromHex(
+    '8000000000000000000000000000000000000000000000000000000000000001',
+);
+const FEE_FREE = /** @type {oasisRT.types.BaseUnits} */ ([
+    oasis.quantity.fromBigInt(0n),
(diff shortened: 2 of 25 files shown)


client-sdk/ts-web/rt/playground/webpack.config.js
```diff
@@ -1,10 +1,24 @@
...
     },
-};
+    {
+        mode: 'development',
+        entry: './src/consensus.js',
+        output: {
+            library: {
+                name: 'playground-consensus',
+                type: 'window',
+                export: 'playground-consensus',
```
pro-wh, line 19: :one: does this work? it's still exported as `playground` and accessed from cypress as `playground`

tests/e2e/go.mod
```diff
@@ -16,7 +16,7 @@ replace (
 )
 
 require (
-	github.com/oasisprotocol/oasis-core/go v0.2101.1
-	github.com/oasisprotocol/oasis-sdk/client-sdk/go v0.0.0-20210328195842-4de788c1c6f7
-	google.golang.org/grpc v1.36.1
+	github.com/oasisprotocol/oasis-core/go v0.2101.1-0.20210504072550-da25d939a3d4
+	github.com/oasisprotocol/oasis-sdk/client-sdk/go v0.0.0-00010101000000-000000000000
```
pro-wh, line 20: just curious, what does this `00010101000000-000000000000` part mean? it used to have a date/time and some hash there. how do you get it to be like this?

client-sdk/ts-web/rt/src/consensus_accounts.ts
```diff
@@ -0,0 +1,66 @@
...
+export interface ConsensusBalanceQuery {
+    address: Uint8Array;
+}
+
+/**
+ * Consensus account balance.
+ */
+export interface AccountBalance {
+    balance: Uint8Array;
+}
```
pro-wh, line 48: :two: I had been putting these data types in `types.ts`. Is that okay with you?

client-sdk/ts-web/rt/cypress/integration/playground.spec.js
```diff
@@ -1,11 +1,19 @@
 /// <reference types="cypress" />
 
 describe('playground', () => {
-    it('finishes', () => {
+    it('finishes keyvalue', () => {
```
pro-wh, line 4: maybe simple-keyvalue? for consistency with simple-consensus below

pro-wh (commented):
looks good, but I'd like to get your response on comments :one: and :two:

client-sdk/ts-web/rt/playground/src/consensus.js
```diff
@@ -0,0 +1,94 @@
...
+        console.log('alice deposit into runtime');
+        var enc = new TextEncoder();
+        const DEPOSIT_AMNT = /** @type {oasisRT.types.BaseUnits} */ ([oasis.quantity.fromBigInt(50n), enc.encode("TEST")]);
+        const twDeposit = consensusWrapper.callDeposit()
+            .setBody({
+                amount: DEPOSIT_AMNT,
+            })
+            .setSignerInfo([siAlice1])
+            .setFeeAmount(FEE_FREE)
+            .setFeeGas(0n);
```
pro-wh, line 68: oh and we'll have to adjust these or the test runtime's gas costs with #117

tests/e2e/go.mod
ptrus: actually not sure, i'll revert it

client-sdk/ts-web/rt/src/consensus_accounts.ts
ptrus: Oh yeah sure i can move these

client-sdk/ts-web/rt/playground/webpack.config.js
ptrus: Good spot, i never tested this with cypress locally but always manually via browser. Updated so that `playground` is exported for both and now it works correctly with cypress.

The main problem was that before the cypress spec passed if the `window.playground` object was undefined. I updated the spec so that it now also ensures that the `window.playground` is actually defined.

pro-wh (approved):
lg thanks

Issue #91, opened 2021-04-09
Multisig accounts


kostko:
The SDK should support multisig accounts.

PR #118, opened 2021-05-01
multisig brainstorming

### LINKED ISSUE #120: Add parameter for maximum number of transaction signers
We should add a parameter to the `core` module that defines the maximum number of transaction signers that can be specified in a transaction.

pro-wh:
#91 from https://github.com/oasisprotocol/oasis-core/pull/3094

due to scope changes, this will fix #120

### DIFF
--- a/client-sdk/go/modules/consensusaccounts/consensus_accounts.go
+++ b/client-sdk/go/modules/consensusaccounts/consensus_accounts.go
@@ -42,7 +42,7 @@ func (a *v1) Deposit(ctx context.Context, signer signature.Signer, nonce uint64,
 	}
 
 	tx := types.NewTransaction(nil, methodDeposit, deposit)
-	tx.AppendSignerInfo(types.AddressSpec{Solo: &types.PublicKey{PublicKey: signer.Public()}}, nonce)
+	tx.AppendSignerInfo(types.AddressSpec{Signature: &types.PublicKey{PublicKey: signer.Public()}}, nonce)
 	stx := tx.PrepareForSigning()
 	if err = stx.AppendSign(info.ChainContext, signer); err != nil {
 		return err
@@ -63,7 +63,7 @@ func (a *v1) Withdraw(ctx context.Context, signer signature.Signer, nonce uint64
 	}
 
 	tx := types.NewTransaction(nil, methodWithdraw, withdraw)
-	tx.AppendSignerInfo(types.AddressSpec{Solo: &types.PublicKey{PublicKey: signer.Public()}}, nonce)
+	tx.AppendSignerInfo(types.AddressSpec{Signature: &types.PublicKey{PublicKey: signer.Public()}}, nonce)
 	stx := tx.PrepareForSigning()
 	if err = stx.AppendSign(info.ChainContext, signer); err != nil {
 		return err
--- a/client-sdk/go/modules/consensusaccounts/consensus_accounts.go
+++ b/client-sdk/go/modules/consensusaccounts/consensus_accounts.go
@@ -42,7 +42,7 @@ func (a *v1) Deposit(ctx context.Context, signer signature.Signer, nonce uint64,
 	}
 
 	tx := types.NewTransaction(nil, methodDeposit, deposit)
-	tx.AppendSignerInfo(signer.Public(), nonce)
+	tx.AppendSignerInfo(types.AddressSpec{Solo: &types.PublicKey{PublicKey: signer.Public()}}, nonce)
 	stx := tx.PrepareForSigning()
 	if err = stx.AppendSign(info.ChainContext, signer); err != nil {
 		return err
@@ -63,7 +63,7 @@ func (a *v1) Withdraw(ctx context.Context, signer signature.Signer, nonce uint64
 	}
 
 	tx := types.NewTransaction(nil, methodWithdraw, withdraw)
-	tx.AppendSignerInfo(signer.Public(), nonce)
+	tx.AppendSignerInfo(types.AddressSpec{Solo: &types.PublicKey{PublicKey: signer.Public()}}, nonce)
 	stx := tx.PrepareForSigning()
 	if err = stx.AppendSign(info.ChainContext, signer); err != nil {
 		return err
--- a/client-sdk/go/modules/consensusaccounts/consensus_accounts.go
+++ b/client-sdk/go/modules/consensusaccounts/consensus_accounts.go
@@ -42,7 +42,7 @@ func (a *v1) Deposit(ctx context.Context, signer signature.Signer, nonce uint64,
 	}
 
 	tx := types.NewTransaction(nil, methodDeposit, deposit)
-	tx.AppendSignerInfo(types.AddressSpec{Signature: &types.PublicKey{PublicKey: signer.Public()}}, nonce)
+	tx.AppendAuthSignature(signer.Public(), nonce)
 	stx := tx.PrepareForSigning()
 	if err = stx.AppendSign(info.ChainContext, signer); err != nil {
 		return err
@@ -63,7 +63,7 @@ func (a *v1) Withdraw(ctx context.Context, signer signature.Signer, nonce uint64
 	}
 
 	tx := types.NewTransaction(nil, methodWithdraw, withdraw)
-	tx.AppendSignerInfo(types.AddressSpec{Signature: &types.PublicKey{PublicKey: signer.Public()}}, nonce)
+	tx.AppendAuthSignature(signer.Public(), nonce)
 	stx := tx.PrepareForSigning()
 	if err = stx.AppendSign(info.ChainContext, signer); err != nil {
 		return err
--- a/client-sdk/go/types/address.go
+++ b/client-sdk/go/types/address.go
@@ -19,7 +19,8 @@ var (
 	AddressV0Ed25519Context = staking.AddressV0Context
 	// AddressV0Secp256k1Context is the unique context for v0 Ed25519-based addresses.
 	AddressV0Secp256k1Context = address.NewContext("oasis-runtime-sdk/address: secp256k1", 0)
-	AddressV0MultisigContext  = address.NewContext("oasis-runtime-sdk/address: multisig", 0)
+	// AddressV0MultisigContext is the unique context for v0 multisig addresses.
+	AddressV0MultisigContext = address.NewContext("oasis-runtime-sdk/address: multisig", 0)
 	// AddressBech32HRP is the unique human readable part of Bech32 encoded
 	// staking account addresses.
 	AddressBech32HRP = staking.AddressBech32HRP
@@ -97,6 +98,7 @@ func NewAddressFromBech32(data string) (a Address) {
 	return
 }
 
+// NewAddressFromMultisig creates a new address from the given multisig configuration.
 func NewAddressFromMultisig(config *MultisigConfig) Address {
 	return (Address)(address.NewAddress(AddressV0MultisigContext, cbor.Marshal(config)))
 }
--- a/client-sdk/go/types/address.go
+++ b/client-sdk/go/types/address.go
@@ -3,6 +3,7 @@ package types
 import (
 	"encoding"
 
+	"github.com/oasisprotocol/oasis-core/go/common/cbor"
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/address"
 	"github.com/oasisprotocol/oasis-core/go/common/encoding/bech32"
 	staking "github.com/oasisprotocol/oasis-core/go/staking/api"
@@ -18,6 +19,7 @@ var (
 	AddressV0Ed25519Context = staking.AddressV0Context
 	// AddressV0Secp256k1Context is the unique context for v0 Ed25519-based addresses.
 	AddressV0Secp256k1Context = address.NewContext("oasis-runtime-sdk/address: secp256k1", 0)
+	AddressV0MultisigContext  = address.NewContext("oasis-runtime-sdk/address: multisig", 0)
 	// AddressBech32HRP is the unique human readable part of Bech32 encoded
 	// staking account addresses.
 	AddressBech32HRP = staking.AddressBech32HRP
@@ -93,4 +95,8 @@ func NewAddressFromBech32(data string) (a Address) {
 		panic(err)
 	}
 	return
+}
+
+func NewAddressFromMultisig(config *MultisigConfig) Address {
+	return (Address)(address.NewAddress(AddressV0MultisigContext, cbor.Marshal(config)))
 }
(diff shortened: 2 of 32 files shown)


runtime-sdk/src/modules/accounts/types.rs
```diff
@@ -4,12 +4,32 @@ use std::collections::BTreeMap;
...
+    SignerZero,
+    MultisigAddress(MultisigAddressSpec),
+}
 
 /// Transfer call.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub struct Transfer {
+    // todo: maybe skip serializing if SignerZero
+    pub from: FromSpec,
```
kostko, line 32: Hm would be better if multisig would work for all transactions. You could make it a separate public key type that includes this spec?

pro-wh: I last brought up this idea of treating multisig as a signature algorithm, conceptually next to ed25519 and others, when we were discussing multisig for oasis-core. but at that time we concluded that time that it was too weird. are there any new considerations now that would change our opinion on it?

kostko: Don't remember what the reasoning was then for Oasis Core, but SDK already supports multiple different signature schemes  (Ed25519 and Secp256k1) and treating "Multisig v1" as another one makes sense to me.

pro-wh: for now I'm going to move the types and stuff from modules/accounts out to the addresses module or something

pro-wh:
update: multisig is moved out to an alternative at the transaction structure level, where SignerInfo.public_key is now an enum allowing you to choose between PublicKey and a multisig config. Likewise, UnverifiedTransaction.1 T is now an enum allowing you to choose between Signature and a multisig signature set. were :pinching_hand: this close to putting "multisig v0" as a signing algorithm next to ed25519 and secp256k1, but not quite. in this state, we can at least use PublicKey and Signature inside the multisig structures without getting all cyclic.

this all won't build yet, as I haven't updated the surrounding code

pro-wh:
update: spreading the changes through the rest of the rust part of the sdk, added a verification routine:

check single-key vs multisig consistency -> zip -> extract present signatures for multisig -> bail if weight is below threshold -> flatten public key and signature lists -> call batch verify

pro-wh:
update: adding sanity checker for multisig configurations. need to add calls to it.

runtime-sdk/src/crypto/multisig/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+        if self.threshold == 0 {
+            return Err(Error::InvalidConfig);
+        }
+        let mut total: u64 = 0;
+        for (i, signer) in self.signers.iter().enumerate() {
+            if self
+                .signers
+                .iter()
+                .take(i)
+                .any(|other_signer| signer.public_key == other_signer.public_key)
```
pro-wh, line 41: I don't know how big these get. we might later prefer to use a hash. would have to figure out how that works on the secp256k1 side.

kostko: There should be parameters to limit these, related to #120 (this one is for transaction signers, but the issue is similar). Also gas costs should probably depend on the number of signers as it is more costly to verify multiple signatures.

pro-wh:
yup, gotta add that, gotta add renames, gotta add doc comments

pro-wh:
opening this for review at this stage. here's the current idea for a data model.

some feedback is in already:
- need doc comments
- need some conventional attributes on rust structures
- need multisig config size limit

we need some discussion on how multisig conceptually fits in to the existing multiple-accounts-i.e.-nonces being able to sign.

in this draft it's not anticipated that a principal that has a key should distinguish between signing any specific SignerInfo slot or any specific entry inside a multisig proof.

btw, the terminology: 'proof,' as introduced in this PR refers to a generalization of (i) a signature satisfying the authentication requirement of a single-public-key account or a (ii) set of signatures satisfying the authentication requirement of a multisig configuration. also 'config' and 'solo.'

and a little weirdness with the encoding of unverified transactions: it's focused on lists that match up one-to-one with the various signer info of the tx, but this results in needing duplicate signatures if the same key would be used in solo and/or multiple multisig proofs. maybe we could flatten and dedup these signatures, if we prefer it that way. it would be more work on the verification side to match them up, maybe

runtime-sdk/src/types/transaction.rs
```diff
@@ -116,12 +128,42 @@ pub struct Fee {
     pub gas: u64,
 }
 
+/// Common information that specifies an address as well as how to authenticate.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub enum AddressSpec {
+    #[serde(rename = "solo")]
+    Solo(PublicKey),
```
ptrus, line 135: Maybe `Single`? or `Singlesig`? No strong opinions though.

kostko (commented):
> we need some discussion on how multisig conceptually fits in to the existing multiple-accounts-i.e.-nonces being able to sign.

I guess the use case would be that one account pays for fees and another account authorizes some action?

> btw, the terminology: 'proof,' as introduced in this PR refers to a generalization of (i) a signature satisfying the authentication requirement of a single-public-key account or a (ii) set of signatures satisfying the authentication requirement of a multisig configuration. also 'config' and 'solo.' let me know your thoughts on this too.

Yeah auth proof and multisig config seem fine. Added a comment regarding "solo".

> and a little weirdness with the encoding of unverified transactions: it's focused on lists that match up one-to-one with the various signer info of the tx, but this results in needing duplicate signatures if the same key would be used in solo and/or multiple multisig proofs. maybe we could flatten and dedup these signatures, if we prefer it that way. it would be more work on the verification side to match them up, maybe. let me know your thoughts on this too

I would rather keep it simple and not dedup.

runtime-sdk/src/types/transaction.rs
pro-wh: forgot to mention this: in oasis-core, we proposed to not to have multiple structures allowed. you would instead make a multisig config with one key. we could do that here if we wanted. or maybe having a system with room for alternate ways to authenticate a transaction is a flexibility we like. or some combination of both where we choose keep an AddressSpec enum with only a multisig variant for now.

runtime-sdk/src/types/transaction.rs
```diff
@@ -130,7 +172,10 @@ pub struct SignerInfo {
 impl SignerInfo {
     /// Create a new signer info from public key and nonce.
     pub fn new(public_key: PublicKey, nonce: u64) -> Self {
-        Self { public_key, nonce }
+        Self {
```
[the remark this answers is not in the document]
pro-wh: I just searched, and it's in a bunch of accounts tests

runtime-sdk/src/types/transaction.rs
pro-wh: thanks for the ideas. how about `Signature`/`Multisig`? there may be things other than multisig that "single" will be distinguished from in the future

pro-wh:
adding some limits for number of authinfo slots in tx and number of signer slots in multisig config

runtime-sdk/src/crypto/multisig/mod.rs
pro-wh: added these limits. haven't added gas cost though

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -212,7 +223,25 @@ impl module::Module for Module {
...
 
-impl module::AuthHandler for Module {}
+impl module::AuthHandler for Module {
+    fn approve_utx(
+        ctx: &mut DispatchContext<'_>,
+        utx: &UnverifiedTransaction,
+    ) -> Result<(), Error> {
+        let params = Self::params(ctx.runtime_state());
+        if utx.1.len() > params.max_tx_signers as usize {
+            return Err(Error::TooManyAuth);
```
kostko, line 233: Should add test cases for these two failure cases.

kostko (commented):
The new spec naming looks good to me. Also the signature limits are nice, would be nice to also charge gas per signature slot.

runtime-sdk/src/types/transaction.rs
kostko: Sounds good to me!

pro-wh: changed!

runtime-sdk/src/modules/core/mod.rs
pro-wh: added!

client-sdk/go/modules/consensusaccounts/consensus_accounts.go
```diff
@@ -42,7 +42,7 @@ func (a *v1) Deposit(ctx context.Context, signer signature.Signer, nonce uint64,
 	}
 
 	tx := types.NewTransaction(nil, methodDeposit, deposit)
-	tx.AppendSignerInfo(signer.Public(), nonce)
+	tx.AppendSignerInfo(types.AddressSpec{Solo: &types.PublicKey{PublicKey: signer.Public()}}, nonce)
```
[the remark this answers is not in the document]
pro-wh: added separate methods!

pro-wh:
> The new spec naming looks good to me. Also the signature limits are nice, would be nice to also charge gas per signature slot.

tracking in #141

client-sdk/go/types/address.go
```diff
@@ -18,6 +19,7 @@ var (
 	AddressV0Ed25519Context = staking.AddressV0Context
 	// AddressV0Secp256k1Context is the unique context for v0 Ed25519-based addresses.
 	AddressV0Secp256k1Context = address.NewContext("oasis-runtime-sdk/address: secp256k1", 0)
+	AddressV0MultisigContext  = address.NewContext("oasis-runtime-sdk/address: multisig", 0)
```
kostko, line 22: Add documentation comment.

client-sdk/go/types/crypto.go
```diff
@@ -71,3 +72,69 @@ func (pk *PublicKey) UnmarshalJSON(data []byte) error {
...
+// MultisigConfig is a multisig configuration.
+// A set of signers with total "weight" greater than or equal to a "threshold" can authenticate
+// for the configuration.
+type MultisigConfig struct {
+	Signers   []MultisigSigner `json:"signers"`
+	Threshold uint64           `json:"threshold"`
+}
+
+// Verify performs some sanity checks.
+func (mc *MultisigConfig) Verify() error {
```
kostko, line 91: Should we call this `Validate` or `ValidateBasic` instead?

client-sdk/go/types/transaction.go
```diff
@@ -56,35 +79,61 @@ type TransactionSigner struct {
...
 // The signer must be specified in the AuthInfo.
 func (ts *TransactionSigner) AppendSign(ctx signature.Context, signer signature.Signer) error {
 	pk := signer.Public()
-	index := -1
+	any := false
 	for i, si := range ts.tx.AuthInfo.SignerInfo {
-		if !si.PublicKey.Equal(pk) {
-			continue
-		}
+		switch {
```
kostko, line 103: Oh I see, so now you sign all the slots that correspond to the signer.

runtime-sdk/src/crypto/multisig/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+    #[serde(rename = "signers")]
+    pub signers: Vec<Signer>,
+    /// The threshold.
+    #[serde(rename = "threshold")]
+    pub threshold: u64,
+}
+
+impl Config {
+    /// Performs some sanity checks.
+    pub fn verify(&self) -> Result<(), Error> {
```
kostko, line 54: Maybe `validate` or `validate_basic`?

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -8,7 +8,10 @@ use crate::{
     context::{Context, DispatchContext},
     error,
     module::{self, Module as _, QueryMethodInfo},
-    types::transaction,
+    types::{
+        transaction,
+        transaction::{AuthProof, UnverifiedTransaction},
```
kostko, line 13:
```suggestion
        transaction::{self, AuthProof, UnverifiedTransaction},
```

kostko (approved):

client-sdk/go/types/crypto.go
Yawning: This should be called anything but `Verify` probably.

client-sdk/go/types/crypto.go
```diff
@@ -71,3 +72,69 @@ func (pk *PublicKey) UnmarshalJSON(data []byte) error {
...
+	}
+	if total < mc.Threshold {
+		return fmt.Errorf("impossible threshold")
+	}
+	return nil
+}
+
+// Batch checks that enough signers have signed and returns vectors of public keys and signatures
+// for batch verification of those signatures. This internally calls `Verify`.
+func (mc *MultisigConfig) Batch(signatureSet [][]byte) ([]PublicKey, [][]byte, error) {
```
Yawning, line 119: Why not do the actual verification?  I'm of mixed opinion if a batch should fail if it as any invalid signatures regardless of threshold or not.  The way this is setup right now is an all-or-nothing design as far as I can tell.

(I assume allowing a MultisigConfig with mixed public-key types is deliberate, though this will complicate batch verification logic.)

client-sdk/go/types/address.go
pro-wh: ok

client-sdk/go/types/crypto.go
pro-wh: ok ValidateBasic

client-sdk/go/types/transaction.go
pro-wh: yeah :shrug: the typescript client is different in style

runtime-sdk/src/crypto/multisig/mod.rs
pro-wh: validate_basic

client-sdk/go/types/crypto.go
pro-wh: we'll have to have an additional step to separate the public key types. but are we moving away from batch signature verification in general?

this isn't meant to be all-or-nothing. it'll leave out the `nil`s from the signature set (Option<Signature> in the rust side). we'd advise that the transaction sender nil out invalid signatures on their own and avoid submitting a transaction with some valid and some invalid signatures

pro-wh: I'll open an issue to discuss this

pro-wh: #143

PR #119, opened 2021-05-01
ts-web/core: ADR 0008 implementation for ts-web

Esya:
Rework of #65 to use SLIP-0010 instead of BIP32-ED25519

# Summary

Implements [ADR 0008: Standard Account Key Generation](https://github.com/oasisprotocol/oasis-core/pull/3656/) in ts-web/core, standardizing the key generation process using [BIP-0039] to generate the seed, and [SLIP-0010] for hierarchical deterministic key generation. 

The value proposition mentioned in [ADR 0008] is as follows : 

> Adopting these standards will allow the Oasis ecosystem to:
> 
> - Make key derivation the same across different applications (i.e. wallets).
> - Allow users to hold keys in hardware wallets.
> - Allow users to hold keys in cold storage more reliably (i.e. using the
  familiar 24 word mnemonics).
> - Define how users can generate multiple keys from a single seed (i.e.
  the 24 or 12 word mnemonic).

# Usage

## Mnemonic and derivation
We add a new class `HDKey` inside `ts-web/core` that allows hierarchical deterministic keys to be generated from [BIP-0039] mnemonics and derived using [SLIP-0010], such as : 

```typescript
const HDKey = oasis.hdkey.HDKey;

// From existing mnemonic
const root = await HDKey.fromMnemonic('practice acoustic taxi');
const key = root.derivePath("m/44'/474'/0''");

// Generating a secure mnemonic
const mnemonic = await HDKey.generateMnemonic();
const root = await HDKey.fromMnemonic(mnemonic);
const key = root.derivePath("m/44'/474'/0'");
```

## Signing with HDKey

```typescript
const root = HDKey.fromMnemonic('practice acoustic taxi');
const key = root.derivePath("m/44'/474'/0'");

const tw = oasis.staking.transferWrapper();

// ... Build transaction ...

const signer = new NaclSigner(key.keyPair, 'this key is not important');
await tw.sign(new oasis.signature.BlindContextSigner(signer), chainContext);
```

# Other changes to ts-web
## Jest support

This pull request adds `jest` as a framework for unit and integration tests, since the only existing tool was cypress that is more suited for e2e tests. While we should keep cypress for when testing browser integration, I propose this becomes standard for unit & integration tests and added a sane and simple `jest.config.js` in `ts-web` that can then be extended or reused by the underlying packages. 

This PR also adds a github action job to run the jest tests for `core`

## Unified webpack configurations

The webpack configuration files from the various playgrounds (`core`, `rt` and `signer-ledger`) were changed to use a main one in `ts-web` .

Webpack 5 does not automatically include polyfills anymore and they must now be explicitly listed - so we polyfill `stream` (needed for crypto/cipher-base part of bip39) for the browser.

[ADR 0008]: https://github.com/oasisprotocol/oasis-core/pull/3656/
[BIP-0032]: https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki
[BIP-0039]: https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki
[SLIP-0010]: https://github.com/satoshilabs/slips/blob/master/slip-0010.md

### DIFF
--- a/.github/workflows/ci-test.yaml
+++ b/.github/workflows/ci-test.yaml
@@ -317,3 +317,26 @@ jobs:
         env:
           TEST_NODE_BINARY: "${{ github.workspace }}/tests/untracked/oasis-node"
           TEST_RUNTIME_LOADER: "${{ github.workspace}}/tests/untracked/oasis-core-runtime-loader"
+
+  jest-ts-web-core:
+    name: jest-ts-web-core
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Set up Node.js LTS
+        uses: actions/setup-node@v2.1.4
+        with:
+          node-version: "14.x"
+
+      - name: Set up npm
+        run: npm install npm@7 -g
+
+      - name: Install Node deps
+        working-directory: client-sdk/ts-web
+        run: npm ci
+
+      - name: Run tests
+        working-directory: client-sdk/ts-web/core
+        run: npm test
--- a/client-sdk/ts-web/core/package.json
+++ b/client-sdk/ts-web/core/package.json
@@ -9,7 +9,8 @@
     "scripts": {
         "prepare": "tsc",
         "check-playground": "cd playground && tsc -p jsconfig.json",
-        "playground": "cd playground && webpack s -c webpack.config.js"
+        "playground": "cd playground && webpack s -c webpack.config.js",
+        "test": "jest"
     },
     "dependencies": {
         "bech32": "^1.1.4",
--- a/client-sdk/ts-web/core/package.json
+++ b/client-sdk/ts-web/core/package.json
@@ -15,6 +15,7 @@
     "dependencies": {
         "bech32": "^1.1.4",
         "cborg": "^1.1.2",
+        "bip39": "^3.0.4",
         "grpc-web": "^1.2.1",
         "js-sha512": "^0.8.0",
         "tweetnacl": "^1.0.3"
--- a/client-sdk/ts-web/core/src/hdkey.ts
+++ b/client-sdk/ts-web/core/src/hdkey.ts
@@ -1,0 +1,100 @@
+import {sha512} from 'js-sha512';
+import {SignKeyPair, sign} from 'tweetnacl';
+import {generateMnemonic, mnemonicToSeed, validateMnemonic} from 'bip39';
+import {concat} from './misc';
+
+const ED25519_CURVE = 'ed25519 seed';
+const HARDENED_OFFSET = 0x80000000;
+const pathRegex = new RegExp("^m(\\/[0-9]+')+$");
+
+/**
+ * HDKey handles hierarchical key generation according to ADR 0008
+ * https://github.com/oasisprotocol/oasis-core/blob/master/docs/adr/0008-standard-account-key-generation.md
+ */
+export class HDKey {
+    public readonly keypair: SignKeyPair;
+
+    /**
+     * Generates the keypair matching the supplied parameters
+     * @param mnemonic BIP-0039 Mnemonic
+     * @param index Account index
+     * @param passphrase Optional BIP-0039 passphrase
+     * @returns SignKeyPair for these parameters
+     */
+    public static async getAccountSigner(
+        mnemonic: string,
+        index: number = 0,
+        passphrase?: string,
+    ): Promise<SignKeyPair> {
+        if (index < 0 || index > 0x7fffffff) {
+            throw new Error('Account number must be >= 0 and <= 2147483647');
+        }
+
+        const seed = await mnemonicToSeed(mnemonic, passphrase);
+        const key = HDKey.makeHDKey(ED25519_CURVE, seed);
+        return key.derivePath(`m/44'/474'/${index}'`).keypair;
+    }
+
+    /**
+     * Generates a mnemonic
+     * @param strength Length in bits of the generated mnemonic
+     * @returns Generated BIP-0039 Mnemonic
+     */
+    public static generateMnemonic(strength = 256): string {
+        return generateMnemonic(strength);
+    }
+
+    private constructor(
+        private readonly privateKey: Uint8Array,
+        private readonly chainCode: Uint8Array,
+    ) {
+        this.keypair = sign.keyPair.fromSeed(privateKey);
+    }
+
+    /**
+     * Returns the HDKey for the given derivation path
+     * using SLIP-0010
+     * @param path Derivation path, starting with m/
+     * @returns Instance of HDKey
+     */
+    private derivePath(path: string): HDKey {
+        if (!pathRegex.test(path)) {
+            throw new Error(
+                "Invalid derivation path. Valid paths must use a format similar to : m/44'/474'/0' and all indexes must be hardened",
+            );
+        }
+
+        const segments = path
+            .split('/')
+            .slice(1)
+            .map((val: string): string => val.replace("'", ''))
+            .map((el) => parseInt(el, 10));
+
+        return segments.reduce((parent, segment) => parent.derive(segment + HARDENED_OFFSET), this);
+    }
+
+    /**
+     * Derive the child key at the specified index
+     * @param index
+     * @returns Instance of HDKey
+     */
+    private derive(index: number): HDKey {
+        const buffer = new ArrayBuffer(4);
+        const view = new DataView(buffer);
+        view.setUint32(0, index);
+
+        const data = concat(new Uint8Array([0]), this.privateKey, new Uint8Array(buffer));
+        return HDKey.makeHDKey(this.chainCode, data);
+    }
+
+    private static makeHDKey(hmacKey: string | Uint8Array, data: Uint8Array): HDKey {
+        //@ts-ignore The types of js-sha512 are outdated
+        const hash = sha512.hmac.arrayBuffer(hmacKey, data);
+
+        const I = new Uint8Array(hash);
+        const IL = I.slice(0, 32);
+        const IR = I.slice(32);
+
+        return new HDKey(IL, IR);
+    }
+}
--- a/client-sdk/ts-web/core/test/adr-0008-vectors.json
+++ b/client-sdk/ts-web/core/test/adr-0008-vectors.json
@@ -1,0 +1,742 @@
+[
+  {
+    "kind": "standard account key generation",
+    "bip39_mnemonic": "abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about",
+    "bip39_passphrase": "",
+    "bip39_seed": "5eb00bbddcf069084889a8ab9155568165f5c453ccb85e70811aaed6f6da5fc19a5ac40b389cd370d086206dec8aa6c43daea6690f20ad3d8d48b2d2ce9e38e4",
+    "oasis_accounts": [
+      {
+        "bip32_path": "m/44'/474'/0'",
+        "private_key": "fb181e94e95cc6bedd2da03e6c4aca9951053f3e9865945dbc8975a6afd217c3ad55bbb7c192b8ecfeb6ad18bbd7681c0923f472d5b0c212fbde33008005ad61",
+        "public_key": "ad55bbb7c192b8ecfeb6ad18bbd7681c0923f472d5b0c212fbde33008005ad61",
+        "address": "oasis1qqx0wgxjwlw3jwatuwqj6582hdm9rjs4pcnvzz66"
+      },
+      {
+        "bip32_path": "m/44'/474'/1'",
+        "private_key": "1792482bcb001f45bc8ab15436e62d60fe3eb8c86e8944bfc12da4dc67a5c89b73fd7c51a0f059ea34d8dca305e0fdb21134ca32216ca1681ae1d12b3d350e16",
+        "public_key": "73fd7c51a0f059ea34d8dca305e0fdb21134ca32216ca1681ae1d12b3d350e16",
+        "address": "oasis1qr4xfjmmfx7zuyvskjw9jl3nxcp6a48e8v5e27ty"
+      },
+      {
+        "bip32_path": "m/44'/474'/2'",
+        "private_key": "765be01f40c1b78dd807e03a5099220c851cfe55870ab082be2345d63ffb9aa40f85ea84b81abded443be6ab3e16434cdddebca6e12ea27560a6ed65ff1998e0",
+        "public_key": "0f85ea84b81abded443be6ab3e16434cdddebca6e12ea27560a6ed65ff1998e0",
+        "address": "oasis1qqtdpw7jez243dnvmzfrhvgkm8zpndssvuwm346d"
+      },
+      {
+        "bip32_path": "m/44'/474'/3'",
+        "private_key": "759b3c2af3d7129072666677b37e9e7b6d22c8bbf634816627e1704f596f60c411ebdac05bfa37b746692733f15a02be9842b29088272354012417a215666b0e",
+        "public_key": "11ebdac05bfa37b746692733f15a02be9842b29088272354012417a215666b0e",
+        "address": "oasis1qqs7wl20gfppe2krdy3tm4298yt9gftxpc9j27z2"
+      },
+      {
+        "bip32_path": "m/44'/474'/4'",
+        "private_key": "db77a8a8508fd77083ba63f31b0a348441d4823e6ba73b65f354a93cf789358d8753c5da6085e6dbf5969773d27b08ee05eddcb3e11d570aaadf0f42036e69b1",
+        "public_key": "8753c5da6085e6dbf5969773d27b08ee05eddcb3e11d570aaadf0f42036e69b1",
+        "address": "oasis1qq8neyfkydj874tvs6ksljlmtxgw3plkkgf69j4w"
+      },
+      {
+        "bip32_path": "m/44'/474'/5'",
+        "private_key": "318e1fba7d83ca3ea57b0f45377e77391479ec38bfb2236a2842fe1b7a624e8800e1a8016629f2882bca2174f29033ec2a57747cd9d3c27f49cc6e11e38ee7bc",
+        "public_key": "00e1a8016629f2882bca2174f29033ec2a57747cd9d3c27f49cc6e11e38ee7bc",
+        "address": "oasis1qrdjslqdum7wwehz3uaw6t6xkpth0a9n8clsu6xq"
+      },
+      {
+        "bip32_path": "m/44'/474'/6'",
+        "private_key": "63a7f716e1994f7a8ab80f8acfae4c28c21af6b2f3084756b09651f4f4ee38606b85d0a8a9747faac85233ad5e4501b2a6862a4c02a46a0b7ea699cf2bd38f98",
+        "public_key": "6b85d0a8a9747faac85233ad5e4501b2a6862a4c02a46a0b7ea699cf2bd38f98",
+        "address": "oasis1qzlt62g85303qcrlm7s2wx2z8mxkr5v0yg5me0z3"
+      },
+      {
+        "bip32_path": "m/44'/474'/7'",
+        "private_key": "34af69924c04d75c79bd120e03d667ff6287ab602f9285bb323667ddf9f25c974f49a7672eeadbf78f910928e3d592d17f1e14964693cfa2afd94b79f0d49f48",
+        "public_key": "4f49a7672eeadbf78f910928e3d592d17f1e14964693cfa2afd94b79f0d49f48",
+        "address": "oasis1qzw2gd3qq8nse6648df32zxvsryvljeyyyl3cxma"
+      },
+      {
+        "bip32_path": "m/44'/474'/8'",
+        "private_key": "aa5242e7efe8dee05c21192766a11c46531f500ff7c0cc29ed59523c5e618792c0a24bf07953520f21c1c25882d9dbf00d24d0499be443fdcf07f2da9601d3e5",
+        "public_key": "c0a24bf07953520f21c1c25882d9dbf00d24d0499be443fdcf07f2da9601d3e5",
+        "address": "oasis1qqzjkx9u549r87ctv7x7t0un29vww6k6hckeuvtm"
+      },
+      {
+        "bip32_path": "m/44'/474'/9'",
+        "private_key": "b661567dcb9b5290889e110b0e9814e72d347c3a3bad2bafe2969637541451e5da8c9830655103c726ff80a4ac2f05a7e0b948a1986734a4f63b3e658da76c66",
+        "public_key": "da8c9830655103c726ff80a4ac2f05a7e0b948a1986734a4f63b3e658da76c66",
+        "address": "oasis1qpawhwugutd48zu4rzjdcgarcucxydedgq0uljkj"
+      },
+      {
+        "bip32_path": "m/44'/474'/2147483647'",
+        "private_key": "cc05cca118f3f26f05a0ff8e2bf5e232eede9978b7736ba10c3265870229efb19e7c2b2d03265ce4ea175e3664a678182548a7fc6db04801513cff7c98c8f151",
+        "public_key": "9e7c2b2d03265ce4ea175e3664a678182548a7fc6db04801513cff7c98c8f151",
+        "address": "oasis1qq7895v02vh40yc2dqfxhldww7wxsky0wgfdenrv"
+      }
+    ]
+  },
+  {
+    "kind": "standard account key generation",
+    "bip39_mnemonic": "equip will roof matter pink blind book anxiety banner elbow sun young",
+    "bip39_passphrase": "",
+    "bip39_seed": "ed2f664e65b5ef0dd907ae15a2788cfc98e41970bc9fcb46f5900f6919862075e721f37212304a56505dab99b001cc8907ef093b7c5016a46b50c01cc3ec1cac",
+    "oasis_accounts": [
+      {
+        "bip32_path": "m/44'/474'/0'",
+        "private_key": "4e9ca1a4c2ed90c90da93ea181557ef9f465f444c0b7de35daeb218f9390d98545601f761af17dba50243529e629732f1c58d08ffddaa8491238540475729d85",
+        "public_key": "45601f761af17dba50243529e629732f1c58d08ffddaa8491238540475729d85",
+        "address": "oasis1qqjkrr643qv7yzem6g4m8rrtceh42n46usfscpcf"
+      },
+      {
+        "bip32_path": "m/44'/474'/1'",
+        "private_key": "2d0d2e75a13fd9dc423a2db8dfc1db6ebacd53f22c8a7eeb269086ec3b443eb627ed04a3c0dcec6591c001e4ea307d65cbd712cb90d85ab7703c35eee07a77dd",
+        "public_key": "27ed04a3c0dcec6591c001e4ea307d65cbd712cb90d85ab7703c35eee07a77dd",
+        "address": "oasis1qp42qp8d5k8pgekvzz0ld47k8ewvppjtmqg7t5kz"
+      },
+      {
+        "bip32_path": "m/44'/474'/2'",
+        "private_key": "351749392b02c6b7a5053bc678e71009b4fb07c37a67b44558064dc63b2efd9219456a3f0cf3f4cc5e6ce52def57d92bb3c5a651fa9626b246cfec07abc28724",
+        "public_key": "19456a3f0cf3f4cc5e6ce52def57d92bb3c5a651fa9626b246cfec07abc28724",
+        "address": "oasis1qqnwwhj4qvtap422ck7qjxf7wm89tgjhwczpu0f3"
+      },
+      {
+        "bip32_path": "m/44'/474'/3'",
+        "private_key": "ebc13ccb62142ed5b600f398270801f8f80131b225feb278d42982ce314f896292549046214fdb4729bf7a6ee4a3bbd0f463c476acc933b2c7cce084509abee4",
+        "public_key": "92549046214fdb4729bf7a6ee4a3bbd0f463c476acc933b2c7cce084509abee4",
+        "address": "oasis1qp36crawwyk0gnfyf0epcsngnpuwrz0mtu8qzu2f"
+      },
+      {
+        "bip32_path": "m/44'/474'/4'",
+        "private_key": "664b95ad8582831fb787afefd0febdddcf03343cc1ca5aa86057477e0f22c93b331288192d442d3a32e239515b4c019071c57ee89f91942923dd4c1535db096c",
+        "public_key": "331288192d442d3a32e239515b4c019071c57ee89f91942923dd4c1535db096c",
+        "address": "oasis1qz8d2zptvf44y049g9dtyqya4g0jcqxmjsf9pqa3"
+      },
+      {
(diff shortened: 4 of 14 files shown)


client-sdk/ts-web/babel.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+    presets: [['@babel/preset-env', {targets: {node: 'current'}}], '@babel/preset-typescript'],
+};
```
nhynes, line 3: Please configure your editor to add newlines at the end of the file.

.github/workflows/ci-test.yaml
```diff
@@ -297,3 +297,26 @@ jobs:
...
+    name: jest-ts-web-core
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Set up Node.js LTS
+        uses: actions/setup-node@v2.1.4
+        with:
+          node-version: "14.x"
```
nhynes, line 311: strictly speaking, the quotes are unnecessary

client-sdk/ts-web/core/src/hdkey.ts
```diff
@@ -0,0 +1,93 @@
...
+        return generateMnemonic(strength);
+    }
+
+    /**
+     * Returns the HDKey for the given derivation path
+     * @param path Derivation path, starting with m/
+     * @returns Instance of HDKey
+     */
+    public derivePath(path: string): HDKey {
+        if (!path.match(pathRegex)) {
```
nhynes, line 58:
```suggestion
        if (!pathRegex.test(path)) {
```

client-sdk/ts-web/core/src/hdkey.ts
```diff
@@ -0,0 +1,93 @@
...
+            throw new Error(
+                "Invalid derivation path. Valid paths must use a format similar to : m/44'/474'/0' and all indexes must be hardened",
+            );
+        }
+
+        const segments = path
+            .split('/')
+            .slice(1)
+            .map(replaceDerive)
```
nhynes, line 67: I was going to suggest making this an explicit arrow fn just in case `replaceDerive` ever gets a second argument, but I saw that it's a simple function that should just be inlined.

client-sdk/ts-web/core/src/hdkey.ts
```diff
@@ -0,0 +1,93 @@
...
+    /**
+     * Derive the child key at the specified index
+     * @param index
+     * @returns Instance of HDKey
+     */
+    protected derive(index: number): HDKey {
+        const indexBuffer = Buffer.allocUnsafe(4);
+        indexBuffer.writeUInt32BE(index, 0);
+
+        const data = Buffer.concat([Buffer.alloc(1, 0), this.privateKey, indexBuffer]);
```
nhynes, line 82: If you care enough about performance to `allocUnsafe`, why not pre-alloc and write the index (and prefix) at the correct offset?

client-sdk/ts-web/core/src/hdkey.ts
```diff
@@ -0,0 +1,93 @@
...
+
+        //@ts-ignore The types of js-sha512 are outdated
+        const hash = sha512.hmac.arrayBuffer(this.chainCode, data);
+
+        const I = new Uint8Array(hash);
+        const IL = I.slice(0, 32);
+        const IR = I.slice(32);
```
nhynes, line 89: I see this incantation twice. How about something like
```ts
function makeHDKey(hmacKey: string, seed: Uint8Array): HDKey {
    //@ts-ignore The types of js-sha512 are outdated
    const I = Buffer.from(sha512.hmac.arrayBuffer(hmacKey, seed));
    const IL = I.slice(0, 32);
    const IR = I.slice(32);
    return new HDKey(IL, IR);
}
```
?

client-sdk/ts-web/core/test/hdkey.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    bip_passphrase: string,
+    bip39_seed: string,
+    oasis_accounts: {
+        private_key: string,
+        public_key: string,
+        address: string
+    }[]
+}
+
+const adr0008Vectors: Adr0008Vector[] = require('./adr-0008-vectors.json');
```
nhynes, line 25: this require is going to mess up ES modules when they become mainstream. TS supports JSON `import`. Would you prefer to use that?

client-sdk/ts-web/core/test/hdkey.test.ts
```diff
@@ -0,0 +1,71 @@
...
+const uint2hex = (array: Uint8Array) => Buffer.from(array).toString('hex');
+const uint2base64 = (array: Uint8Array) => Buffer.from(array).toString('base64');
+
+describe('HDKey', () => {
+    describe('Adr-0008 Vectors', () => {
+        // @Todo waiting on oasis-core test vectors
+        adr0008Vectors.forEach((vector, index) => {
+            it(`Case #${index}`, async () => {
+                // This can be a bit slow on CI servers.
+                jest.setTimeout(10000)
```
nhynes, line 37: where did your semis go?

client-sdk/ts-web/jest.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+    testMatch: ['**/test/**/*.ts'],
+    testEnvironment: 'node',
+    transform: {
+        '^.+\\.[jt]s?$': 'babel-jest',
```
nhynes, line 5: Do you have any reason to prefer this over [ts-jest](https://www.npmjs.com/package/ts-jest)?

client-sdk/ts-web/core/test/hdkey.test.ts
Esya: It's in a test repository so it should not be published and should always run in jest's context, and using `import x from 'test.json'` makes it require an additional variable to type it, or would require manually adding declaration `.d.ts` files for those vectors which we might not want for tests only ?

client-sdk/ts-web/jest.config.js
Esya: It's faster, and avoids type-checking twice as well (see the PR's body)

client-sdk/ts-web/core/test/hdkey.test.ts
Esya: Oops, I'm strongly anti-semi so I usually let the linter add them in for me - I added the test folder to the lint/fmt commands so that they're also linted.

.github/workflows/ci-test.yaml
Esya: I followed the exact template of the other e2e test invocations in this file - should they all be fixed in this PR ?

client-sdk/ts-web/babel.config.js
Esya: Fixed those files, if it is standard for this repository it should probably be in a standardized [editorconfig](https://editorconfig.org/) config file along with other standards at the root of the repository

client-sdk/ts-web/core/src/hdkey.ts
Esya: Do you mean this ?

```typescript
    const data = Buffer.allocUnsafe(37);
    data.writeUInt8(0, 0)
    data.fill(this.privateKey, 1, 33)
    data.writeUInt32BE(index, 33);
```

nhynes: yeah that looks about right

client-sdk/ts-web/jest.config.js
nhynes: Thanks for the writeup. Are you sure you don't want typechecked tests? I find it beneficial to have early feedback that I'm using the APIs incorrectly [when writing tests].

client-sdk/ts-web/core/test/hdkey.test.ts
nhynes: It's not about publishing. It's that if the package is published as ESM (so `type: module` in `package.json`), Node will interpret the entire package as `.js == ESM`, and hence disable `require`. If it's significant effort to get JSON imports working, I'd not worry about it, but if it's easy then it'd be a nice way to add near-future compatibility.

.github/workflows/ci-test.yaml
nhynes: nah. thanks for offering, though!

client-sdk/ts-web/core/src/hdkey.ts
Esya: After careful consideration and benchmarking I think it's better to keep it as it is as it's more performant while also being easier to read

```javascript
const Benchmark = require('benchmark');
const suite = new Benchmark.Suite;

const index = 0;
const pk = new Uint8Array(32)

suite
.add('Buffer#prealloc', function() {
    const data = Buffer.allocUnsafe(37);
    data.writeUInt8(0, 0);
    data.fill(pk, 1, 33)
    data.writeUInt32BE(index, 33);
})
.add('Buffer#concat', function() {
    const indexBuffer = Buffer.allocUnsafe(4);
    indexBuffer.writeUInt32BE(index, 0);

    const data = Buffer.concat([Buffer.alloc(1, 0), pk, indexBuffer]);
})
/*
Buffer#prealloc x 3,620,673 ops/sec ±0.34% (292 runs sampled)
Buffer#concat x 4,385,810 ops/sec ±0.35% (291 runs sampled)
Fastest is Buffer#concat
*/
````

File: client-sdk/ts-web/core/test/hdkey.test.ts (hunk 7 above)
Comment (Esya) [reply client-sdk/ts-web/core/test/hdkey.test.ts]: Ah, I see what you mean. I'll `import` them and assign to another typed variable, it's not the prettiest but the other options are not really relevant at this time.

File: client-sdk/ts-web/jest.config.js (hunk 9 above)
Comment (Esya) [reply client-sdk/ts-web/jest.config.js:5]: You're right - In my other projects we usually have a separate step with a different tsconfig to typecheck the tests as well but I think this would be a bit overkill right now - I'll amend the PR to use ts-jest even if it's a bit slower, if one day it becomes a problem it can be revisited and optimised.

Comment (Esya) [reply client-sdk/ts-web/jest.config.js]: I've changed the PR to use ts-jest, and updated the tsconfig to turn on esModuleInterop / resolveJsonModule - which are sane defaults to have on anyways for better interoperability.

Comment (nhynes) [reply client-sdk/ts-web/jest.config.js]: thanks! that's everything from me. (ping @pro-wh for an approving ✅)

**Review by nhynes — APPROVED** [review]

File: .github/workflows/ci-test.yaml (hunk 2 above)
Comment (pro-wh) [reply .github/workflows/ci-test.yaml:311]: it's "I copied from an earlier test" all the way down. I wonder if it originates from a specimen that didn't have the ".x" and would be parsed as a number

File: client-sdk/ts-web/webpack.config.js [hunk 10]
```diff
@@ -0,0 +1,19 @@
...
+            name: 'playground',
+            type: 'window',
+            export: 'playground',
+        },
+    },
+    resolve: {fallback: {stream: require.resolve('stream-browserify')}},
+    plugins: [
+        new webpack.ProvidePlugin({
+            process: 'process/browser.js',
+            Buffer: ['buffer', 'Buffer'],
```
pro-wh, line 16: I hadn't seen it done this way. maybe we could use this to let ledger-js compile with webpack 5. all the docs on webpack say is "ya'd better make sure your stuff works without process and Buffer before you choose to use webpack 5" :sweat: 

Could we have these polyfills isolated to where it's needed, rather than in this shared config file? Stream in `core` and process and Buffer in `signer-ledger`. Possibly Buffer in `core`, although I've left a comment on the possibility of that not being needed there.

client-sdk/ts-web/core/src/hdkey.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /**
+     * Derive the child key at the specified index
+     * @param index
+     * @returns Instance of HDKey
+     */
+    protected derive(index: number): HDKey {
+        const indexBuffer = Buffer.allocUnsafe(4);
+        indexBuffer.writeUInt32BE(index, 0);
```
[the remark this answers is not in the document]
pro-wh: https://github.com/oasisprotocol/oasis-sdk/blob/main/client-sdk/ts-web/rt/src/event.ts#L6 similar construction here, for reference

client-sdk/ts-web/core/test/adr-0008-vectors.json
```diff
@@ -0,0 +1,466 @@
+[
```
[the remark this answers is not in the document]
pro-wh: ah I just saw that they're proposed to be added in the oasis-core adr-0008 implementation. @tjanez could you check off on these matching https://github.com/oasisprotocol/oasis-core/pull/3918?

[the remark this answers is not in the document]
tjanez: @pro-wh, the test vectors have been updated in https://github.com/oasisprotocol/oasis-core/pull/3918.

@Esya, could you generate new ones via `make -C go staking/gen_account_vectors_extended`?

client-sdk/ts-web/core/src/hdkey.ts
```diff
@@ -0,0 +1,89 @@
+import {sha512} from 'js-sha512';
```
tjanez, line 1: Perhaps it would make sense to slip SLIP-0010 implementation into a separate package?
Or is there an existing JS implementation that would be suitable?

client-sdk/ts-web/core/src/hdkey.ts
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        const segments = path
+            .split('/')
+            .slice(1)
+            .map((val: string): string => val.replace("'", ''))
+            .map((el) => parseInt(el, 10));
+
+        return segments.reduce((parent, segment) => parent.derive(segment + HARDENED_OFFSET), this);
+    }
```
tjanez, line 64: I think it would make sense to have a high-level function analogous to (taken from https://github.com/oasisprotocol/oasis-core/pull/3918/):
```go
// Generate signer for the given mnemonic, passphrase and account according to
// ADR 0008.
func GetAccountSigner(
	mnemonic string,
	passphrase string,
	number uint32,
) (signature.Signer, BIP32Path, error) {
...
}
```
In this case, it wouldn't return a signer but the `SignKeyPair` object.

That way, ts-web users don't need to worry about the low-level BIP-0032 paths but can just access Oasis accounts by their numbers (for a given wallet).

client-sdk/ts-web/webpack.config.js
Esya: Stream is actually required in all of them - but since the configs have changed a bit between them and I'm not aware of the specifics I added them separately

client-sdk/ts-web/core/src/hdkey.ts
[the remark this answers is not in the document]
Esya: Done!

client-sdk/ts-web/core/test/adr-0008-vectors.json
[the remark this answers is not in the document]
Esya: Updated the vectors

client-sdk/ts-web/webpack.config.js
Esya: Actually, we still need buffer because of bip39, and in all three packages

```
  > Buffer is not defined

When Cypress detects uncaught errors originating from your application it will automatically fail the current test.

This behavior is configurable, and you can choose to turn this off by listening to the `uncaught:exception` event.

https://on.cypress.io/uncaught-exception-from-application
      at eval (webpack://playground/../../node_modules/core-util-is/lib/util.js?:103:20)
      at Object.../../node_modules/core-util-is/lib/util.js (http://localhost:42280/main.js:676:1)
      at __webpack_require__ (http://localhost:42280/main.js:1648:32)
      at fn (http://localhost:42280/main.js:1848:21)
      at eval (webpack://playground/../../node_modules/readable-stream/lib/_stream_readable.js?:67:26)
      at Object.../../node_modules/readable-stream/lib/_stream_readable.js (http://localhost:42280/main.js:1144:1)
      at __webpack_require__ (http://localhost:42280/main.js:1648:32)
      at fn (http://localhost:42280/main.js:1848:21)
      at eval (webpack://playground/../../node_modules/readable-stream/readable-browser.js?:1:28)
      at Object.../../node_modules/readable-stream/readable-browser.js (http://localhost:42280/main.js:1228:1)
```

pro-wh:
thanks for the changes. what was it that uses stream? https://github.com/bitcoinjs/bip39/blob/master/ts_src/index.ts I didn't see any references in this file. something even farther indirect?

client-sdk/ts-web/core/src/hdkey.ts
```diff
@@ -0,0 +1,107 @@
+import {sha512} from 'js-sha512';
+import {SignKeyPair, sign} from 'tweetnacl';
+import {generateMnemonic, mnemonicToSeed} from 'bip39';
+import {concat} from './misc';
+
+const ED25519_CURVE = 'ed25519 seed';
+const HARDENED_OFFSET = 0x80000000;
+const pathRegex = new RegExp("^m(\\/[0-9]+')+$");
+
+export class HDKey {
```
tjanez, line 10: Would it make sense to mention that this implements ADR 0008 somewhere?

client-sdk/ts-web/core/src/hdkey.ts
```diff
@@ -0,0 +1,107 @@
...
+    public readonly keypair: SignKeyPair;
+
+    constructor(public readonly privateKey: Uint8Array, public readonly chainCode: Uint8Array) {
+        this.keypair = sign.keyPair.fromSeed(privateKey);
+    }
+
+    /**
+     * Generates the keypair matching the supplied parameters
+     * @param mnemonic Bip39 Mnemonic
```
tjanez, line 19: Nit: I would refer to BIP 39 in a docstring as "BIP-0039".

client-sdk/ts-web/core/src/hdkey.ts
```diff
@@ -0,0 +1,107 @@
...
+     * Generates a mnemonic
+     * @param strength Length in bits of the generated mnemonic
+     * @returns Generated BIP39 Mnemonic
+     */
+    public static generateMnemonic(strength = 256): string {
+        return generateMnemonic(strength);
+    }
+
+    /**
+     * Returns the HDKey for the given derivation path
```
tjanez, line 63: I would mention SLIP-0010 in this docstring and the docstrings of other functions that implement SLIP-0010.

client-sdk/ts-web/core/test/hdkey.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    }[];
+}
+
+const adr0008Vectors: Adr0008Vector[] = adr0008VectorsRaw;
+const slip10Vectors: Slip0010Vector[] = slip0010VectorsRaw;
+
+const uint2hex = (array: Uint8Array) => Buffer.from(array).toString('hex');
+
+describe('HDKey', () => {
+    describe('Adr-0008 Vectors', () => {
```
tjanez, line 34: Nit:
```suggestion
    describe('ADR 0008 Vectors', () => {
```

client-sdk/ts-web/core/src/hdkey.ts
tjanez: > I'm pretty happy with the exposed API for HDKey - the underlying implementation of SLIP-0010 can be changed in the future if we want to extract my implementation while being a non-breaking change and without impacting the exposed API.
> 

Ok, let's leave it as is for now.

> The others I found were already part of specific crypto's ecosystems, in JS without typings, had no tests, or a combination of those factors. (And I've seen none using dataviews & typed-arrays exclusively instead of buffer)

Cool, thanks for the explanation.

tjanez:
Another nit for commit messages:
- "ts-web: adds jest support" -> "ts-web: add jest support"
- "ts-web/core: adr 0008 implementation for ts-web" -> "ts-web/core: ADR 0008 implementation for ts-web"

pro-wh:
> > thanks for the changes. what was it that uses stream? https://github.com/bitcoinjs/bip39/blob/master/ts_src/index.ts I didn't see any references in this file. something even farther indirect?
> 
> ```
> └─┬ @oasisprotocol/client@0.1.0-alpha1
>   └─┬ bip39@3.0.4
>     └─┬ create-hash@1.2.0
>       └─┬ md5.js@1.3.5
>         └─┬ hash-base@3.1.0
>           └── readable-stream@3.6.0
> ```

MD5, interesting. Of course bip39 doesn't use that https://github.com/bitcoinjs/bip39/blob/200c6a93ba0bf946d982e133a00a2c081240af9a/ts_src/index.ts#L61. Maybe we can advise webpack not to worry about that part of the dependency tree. I'll file a separate issue for that. Thanks for the info.

client-sdk/ts-web/core/package.json
```diff
@@ -9,10 +9,12 @@
     "scripts": {
         "prepare": "tsc",
         "check-playground": "cd playground && tsc -p jsconfig.json",
-        "playground": "cd playground && webpack s -c webpack.config.js"
+        "playground": "cd playground && webpack s -c webpack.config.js",
+        "test": "jest"
     },
     "dependencies": {
         "bech32": "^1.1.4",
+        "bip39": "^3.0.4",
```
pro-wh, line 17: could you move this one line to the other commit please :pray:

client-sdk/ts-web/core/src/hdkey.ts
```diff
@@ -0,0 +1,112 @@
...
+     */
+    public static async fromMnemonic(mnemonic: string, passphrase?: string): Promise<HDKey> {
+        const seed = await mnemonicToSeed(mnemonic, passphrase);
+        return HDKey.fromSeed(seed);
+    }
+
+    /**
+     * Generates a mnemonic
+     * @param strength Length in bits of the generated mnemonic
+     * @returns Generated BIP39 Mnemonic
```
tjanez, line 60:
```suggestion
     * @returns Generated BIP-0039 Mnemonic
```

tjanez (commented):
Just a small nit

client-sdk/ts-web/core/test/slip-0010-vectors.json
```diff
@@ -0,0 +1,82 @@
+[
+    {
+        "seed": "000102030405060708090a0b0c0d0e0f",
```
Yawning, line 3: This test vector set being in the library is somewhat questionable, though it isn't your fault.

The spec reads as thus:
> 1. Generate a seed byte sequence S of 512 bits according to BIP-0039.

I have no idea why the SLIP-0010 test vectors include results that are generated from invalid input.  The equivalent Go code in oasis-core strictly enforces that `S` is exactly 512 bits.  I would suggest that at a minimum, the user-callable API also does the same.

client-sdk/ts-web/core/src/hdkey.ts
```diff
@@ -0,0 +1,112 @@
...
+        const key = await HDKey.fromMnemonic(mnemonic, passphrase);
+        return key.derivePath(`m/44'/474'/${index}'`).keypair;
+    }
+
+    /**
+     * Returns an instance of HDKey that can then be used for derivation
+     * @param seed Master seed for key derivation
+     * @returns Instance of HDKey
+     */
+    public static fromSeed(seed: Uint8Array): HDKey {
```
tjanez, line 42: Can we make this function match the Go implementation and reject seeds smaller than 64 bytes?

Reference: https://github.com/oasisprotocol/oasis-core/blob/27c2ceaccfa696351c7d7ca89645e3887fac8d4f/go/common/crypto/slip10/slip10.go#L28-L43

Note that this is actually what [SLIP-0010](https://github.com/satoshilabs/slips/blob/master/slip-0010.md#master-key-generation) specifies, but their [_Test vector 1 for ed25519_](https://github.com/satoshilabs/slips/blob/master/slip-0010.md#test-vector-1-for-ed25519) violates that :slightly_smiling_face: .

Esya: By specification, SLIP-0010 is supposed to be compatible with BIP-0032 :

> Generate a seed byte sequence S of a chosen length (between 128 and 512 bits; 256 bits is advised) from a (P)RNG.

These vectors, being 128 bits, are to test that the underlying implementation works with older 128 bits BIP-0032 keys. 

If the key derivation happens through BIP-0039, the seed will be 512 bits anyways, but I don't think that we should exclude users that want to reuse existing seeds if they absolutely want to

client-sdk/ts-web/core/test/slip-0010-vectors.json
Esya: It's for the BIP-0032 compatibility which allows 128 bits seeds - see [my response here](https://github.com/oasisprotocol/oasis-sdk/pull/119#discussion_r631505798)

client-sdk/ts-web/core/src/hdkey.ts
Yawning: > These vectors, being 128 bits, are to test that the underlying implementation works with older 128 bits BIP-0032 keys.

There is nothing in the SLIP-0010 spec that allows for this, apart from `It uses a process similar and compatible to BIP-0032. For other curves it uses a different salt than BIP-0032.` and `Curve = "Bitcoin seed" for the secp256k1 curve (this is compatible to BIP-0032).`.

On the other hand:

SLIP-0010: `Generate a seed byte sequence S of 512 bits according to BIP-0039.`

BIP-0032: `Generate a seed byte sequence S of a chosen length (between 128 and 512 bits; 256 bits is advised) from a (P)RNG.`

> After thinking about it some more, I strongly think that deviating from all the other implementations and libraries out there, and cutting half of the test vectors would be problematic. It is ultimately up to the user to use the sdk as they see fit, and here it would be similar to refusing 3 words or 6 words mnemonics : while it's not optimal/recommended, it's still valid and the sdk should not be so opinionated that it refuses a valid key from a user

I disagree.

 * It is possible to still test the other test vector.  If typescript doesn't have a way to test package private functions, it might be more annoying to do so, but I do not write typescript ever.
 * ADR-0008 mandates BIP-0039, which produces 512 bit `S`.
 * SLIP-0010 explicitly defines 512 bit `S` (BIP-0032 explicitly does not, but we are not using BIP-0032, we are using SLIP-0010).
 * SLIP-0010's definition of the algorithm runs contrary to their compatibility claims, and the algorithm should be implemented as specified.

In general, I tend to favor the most concrete/narrow reading/interpretation of a specification possible because not doing so leads to problems down the lines.  Because it would helps clarify things, amending ADR-0008 to point out that SLIP-0010 is poorly specified and self contradictory, along with the interpretation we use is fine.

tjanez (approved):

client-sdk/ts-web/core/test/hdkey.test.ts
```diff
@@ -0,0 +1,46 @@
...
+            it(`Case #${index}`, async () => {
+                // This can be a bit slow on CI servers.
+                jest.setTimeout(10000);
+                const passphrase =
+                    vector.bip39_passphrase && vector.bip39_passphrase !== ''
+                        ? vector.bip39_passphrase
+                        : null;
+
+                for (let account of vector.oasis_accounts) {
+                    const index = Number(account.bip32_path.split('/').pop().replace("'", ''));
```
pro-wh, line 33: maintainability idea: assert that the beginning of the path starts with the expected values. in case future specs come and someone adds new entries in the json file without knowing that we ignore the beginning

client-sdk/ts-web/jest.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+    testMatch: ['**/test/**/*.ts'],
```
pro-wh, line 2: could you confirm the following and add it as a comment

> We load this file from the packages' jest.config.js files (e.g. see core/jest.config.js), so it'll be resolved relative to the package's directory.

pro-wh (approved):

tjanez:
@Esya, can you rebase and take a look at @pro-wh's [last 2 comments](https://github.com/oasisprotocol/oasis-sdk/pull/119#pullrequestreview-659964172)?

client-sdk/ts-web/core/test/hdkey.test.ts
```diff
@@ -0,0 +1,46 @@
+import {HDKey} from '../src/hdkey';
```
tjanez, line 1: Can you add tests that will check if `getAccountSigner()` function call is passed:
- invalid mnemonic
- invalid account key number (i.e. bigger than `MaxAccountKeyNumber` in Go implementation).

Here are the corresponding tests in Go implementation:
https://github.com/oasisprotocol/oasis-core/pull/3918/files#diff-f92eefb38a2a59271dc7d0efad661a2cdb2034b2e1ba9df00672dc6a092b8b32R62-R81

BTW, in Go implementation the account key number has `uint32` type so there is no need to check if the number is < 0. How does the `number` type in TS behave? It will probably need a check that the passed number is not < 0.

pro-wh: numbers are double precision floats. this one flows into a setUint32, which silently converts them to uint32 in some specific way

Esya: @tjanez I've added a check for the account number

For the validity of the mnemonic this should be changed in the go implementation, the user implementation should raise a warning, but should not prevent the user from using any mnemonic, see [Bitcoinjs's reminder for developers](https://github.com/bitcoinjs/bip39#reminder-for-developers) : 

> Please remember to allow recovery from mnemonic phrases that have invalid checksums (or that you don't have the wordlist) When a checksum is invalid, warn the user that the phrase is not something generated by your app, and ask if they would like to use it anyway. This way, your app only needs to hold the wordlists for your supported languages, but you can recover phrases made by other apps in other languages.

Also from [BIP-0039](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#From_mnemonic_to_seed) : 

> Although using a mnemonic not generated by the algorithm described in "Generating the mnemonic" section is possible, this is not advised and software must compute a checksum for the mnemonic sentence using a wordlist and issue a warning if it is invalid.

client-sdk/ts-web/core/src/hdkey.ts
```diff
@@ -0,0 +1,100 @@
...
+     * @param passphrase Optional BIP-0039 passphrase
+     * @returns SignKeyPair for these parameters
+     */
+    public static async getAccountSigner(
+        mnemonic: string,
+        index: number = 0,
+        passphrase?: string,
+    ): Promise<SignKeyPair> {
+        if (index < 0 || index > 0x7fffffff) {
+            throw new Error('Account number must be >= 0 and <= 8388607');
```
pro-wh, line 30: comparison checks against 2147483647, should the error message match?

client-sdk/ts-web/core/test/hdkey.test.ts
pro-wh: thanks for the explanation. I'm not opposed to adding the tests. a smoketest for different language mnemonic to make sure it works, and another non-official-test-vector test to make sure invalid account key number fails as expected

client-sdk/ts-web/core/src/hdkey.ts
Esya: Good catch, missed a pair of `ff` !

client-sdk/ts-web/core/src/hdkey.ts
MBANKm: cosmos1hkmy9n3mg5j8t4fsvvm7zzw4x0lar5n3eu2tt9

PR #124, opened 2021-05-06
runtime-sdk: Add rewards module

### LINKED ISSUE #93: Rewards module
We should implement a simple module for disbursing rewards to node operators. Rewards should come from a specific module account and be disbursed accoording to a genesis configuration parameter.

kostko:
Fixes #93 

TODO
* [ ] More tests


runtime-sdk/src/modules/rewards/types.rs
```diff
@@ -0,0 +1,308 @@
...
+    pub fn for_disbursement(
+        &self,
+        threshold_numerator: u64,
+        threshold_denominator: u64,
+    ) -> impl Iterator<Item = Address> + '_ {
+        let max_v = self
+            .pending
+            .iter()
+            .fold(0, |acc, (_, action)| std::cmp::max(acc, action.value()));
+        let threshold = (threshold_numerator * max_v) / threshold_denominator;
```
pro-wh, line 125: this should be fine because we only increment the values by one for anything

client-sdk/ts-web/rt/src/types.ts
```diff
@@ -45,6 +45,25 @@ export interface AccountsTransferEvent {
...
+ * Parameters for the rewards module.
+ */
+export interface RewardsParameters {
+    schedule: RewardsRewardSchedule;
+
+    participation_threshold_numerator: number;
+    participation_threshold_denominator: number;
+}
+
+export interface RewardsRewardSchedule {
```
pro-wh, line 58: might as well fill in the doc comments for the schedule and step too

pro-wh (approved):
rewards

runtime-sdk/src/modules/rewards/types.rs
kostko: Added some checks nevertheless.

PR #125, opened 2021-05-07
runtime-sdk: Do not vote for obviously malformed transactions

### LINKED ISSUE #56: Do not vote for obviously malformed transactions
Currently a bad proposer can include malformed transactions in a block and while they will be rejected during processing they will still be included.

If a batch includes obviously malformed transactions (not following the transaction format or having invalid signatures), the batch should be rejected as invalid by the runtime, returning an error from `RuntimeExecuteTxBatchRequest` (which will cause the workers to not vote for the proposal and indicate a failure instead).

kostko:
Fixes #56

### DIFF
--- a/runtime-sdk/src/dispatcher.rs
+++ b/runtime-sdk/src/dispatcher.rs
@@ -31,18 +31,20 @@ use crate::{
     storage, types,
 };
 
+/// Unique module name.
+const MODULE_NAME: &str = "dispatcher";
+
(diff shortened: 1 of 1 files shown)


pro-wh (approved):
alright

but I worry about our style of discarding so much diagnostic info

kostko:
What do you mean, what is discarded in this case?

pro-wh:
oh nvm it's not. then what uses `Aborted`?

PR #135, opened 2021-05-13
tests/e2e: Add transaction generator

### LINKED ISSUE #50: Add transaction and query fuzzing tests
We should have some transaction/query E2E fuzzing tests similar to what we have with the `txsource` workloads in Oasis Core. It should be made in such a way that can be easily reused in other (external) modules and runtimes.

Requires #21.

abukosek:
Closes #50.

This PR adds a reusable random transaction generator to the e2e tests.

Currently, an e2e test uses the transaction generator to generate random transactions and queries for the `simple-keyvalue` runtime (including transfers between accounts) for 1 minute and then stops.
A separate long-term test that generates random transactions and queries for 15 minutes is scheduled to run once per day at midnight UTC.

### DIFF
--- a/.github/workflows/ci-longtest.yaml
+++ b/.github/workflows/ci-longtest.yaml
@@ -1,0 +1,48 @@
+# NOTE: This name appears in GitHub's Checks API and in workflow's status badge.
+name: ci-longtest
+
+# Trigger the workflow when:
+on:
+  # Every day at 00:00 UTC.
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+
+  e2e-rt-long:
+    # NOTE: This name appears in GitHub's Checks API.
+    name: e2e-rt-long
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Set up Rust
+        uses: actions-rs/toolchain@v1
+
+      - name: Set up Go 1.16
+        uses: actions/setup-go@v2.1.3
+        with:
+          go-version: "1.16.x"
+
+      - name: Install Oasis dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install bubblewrap build-essential libseccomp-dev libssl-dev protobuf-compiler
+
+      - name: Download artifacts
+        working-directory: tests/
+        run: ./download-artifacts.sh
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Run long end-to-end tests
+        run: |
+          ./tests/run-e2e.sh \
+            --e2e/test-runtime-simple-keyvalue.txgen.coins_per_acct=100000 \
+            --e2e/test-runtime-simple-keyvalue.txgen.num_accounts=100 \
+            --e2e/test-runtime-simple-keyvalue.txgen.duration=15m
+        env:
+          TEST_NODE_BINARY: "${{ github.workspace }}/tests/untracked/oasis-node"
+          TEST_RUNTIME_LOADER: "${{ github.workspace}}/tests/untracked/oasis-core-runtime-loader"
+
--- a/tests/consts.sh
+++ b/tests/consts.sh
@@ -1,5 +1,7 @@
 OASIS_CORE_VERSION=21.1-dev
-# To download a pre-released version of oasis-core built by oasis-core/release-dev action
-# set artifact ID to use here.
-# To download a released version of oasis-core unset this variable.
+# To download a pre-released version of oasis-core built by
+# oasis-core/release-dev action, set the artifact ID to use here.
+# You must set GITHUB_TOKEN as well, if it is unset, the release version
+# specified above will be downloaded.
 GITHUB_ARTIFACT=58512799 # 5214f87
+
--- a/tests/download-artifacts.sh
+++ b/tests/download-artifacts.sh
@@ -1,16 +1,21 @@
-#!/bin/sh -eux
+#!/usr/bin/env bash
+set -o nounset -o pipefail -o errexit
+
+# Get the configuration from the current working directory.
 . ./consts.sh
 
 mkdir -p untracked
 if [ ! -e "untracked/oasis-node" ]; then
     (
         cd untracked
-        if [ ! -z "${GITHUB_ARTIFACT:-}" ]; then
+        if [ ! -z "${GITHUB_ARTIFACT:-}" ] && [ ! -z "${GITHUB_TOKEN:-}" ]; then
             # Authentication is required to download the artifacts, although those are public.
+            echo "### Downloading github artifact ${GITHUB_ARTIFACT}..."
             curl -fL -o oasis-core.zip -H "Authorization: Bearer ${GITHUB_TOKEN}" "https://api.github.com/repos/oasisprotocol/oasis-core/actions/artifacts/${GITHUB_ARTIFACT}/zip"
             unzip oasis-core.zip
         else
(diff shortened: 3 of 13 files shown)


tests/e2e/simplekvtest.go
```diff
@@ -490,3 +501,83 @@ func KVRewardsTest(log *logging.Logger, conn *grpc.ClientConn, rtc client.Runtim
...
+	var balance uint64
+	if q, ok := ab.Balances[types.NativeDenomination]; ok {
+		// We can do this only because the account's balance fits into an uint64.
+		balance = q.ToBigInt().Uint64()
+	} else {
+		return fmt.Errorf("Alice's account is missing native denomination balance")
+	}
+
+	numAccounts := 10
+	coinsPerAccount := uint64(200)
```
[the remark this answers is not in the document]
abukosek: It takes about 3 seconds to create and fund an account, so I'd rather do this in a separate long-term test.

tests/e2e/txgen/txgen.go
```diff
@@ -0,0 +1,169 @@
...
+		subErrCount uint64
+		okCount     uint64
+	)
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil, genErrCount, subErrCount, okCount
+		case <-ticker.C:
+			// Choose random account and txn generator.
```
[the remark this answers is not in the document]
abukosek: The `MaxBatchSize` in our oasis-sdk e2e fixture is currently 10, I'll increase it :)

tests/e2e/txgen/txgen.go
```diff
@@ -0,0 +1,169 @@
...
+			acct := accounts[rand.Intn(len(accounts))]
+			gen := txGens[rand.Intn(len(txGens))]
+
+			go func(acct signature.Signer, gen GenerateTx) {
+				// Generate random transaction or perform random query.
+				if tx, err := gen(ctx, rtc, acct, accounts); err != nil {
+					atomic.AddUint64(&genErrCount, 1)
```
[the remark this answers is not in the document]
abukosek: I had that at first, but there was too much spam -- since the transactions are generated in parallel, there are a lot of invalid nonce errors and since the generated keys are random, a lot of them don't exist in storage, etc.
It would be better to keep a histogram of error messages instead and return that.

tests/e2e/simplekvtest.go
```diff
@@ -490,3 +501,83 @@ func KVRewardsTest(log *logging.Logger, conn *grpc.ClientConn, rtc client.Runtim
...
+	}
+	log.Info("accounts generated", "num_secp256k1", numS, "num_ed25519", numE)
+
+	// Generate random transactions for 1 minute.
+	log.Info("generating transactions")
+	txgenCtx, cancel := context.WithTimeout(ctx, 60*time.Second)
+	defer cancel()
+
+	// Generate a new random tx every 250ms until txgenCtx timeouts.
+	err, genErrs, subErrs, ok := txgen.Generate(txgenCtx, rtc, accts, append(txgen.DefaultTxGenerators, DefaultKVTxGenerators...), 250*time.Millisecond)
```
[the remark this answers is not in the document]
abukosek: Oops, I'll fix it.

abukosek:
Thanks for the review -- I've addressed the comments, so this is now ready for another pass :)

tests/e2e/simplekvtest.go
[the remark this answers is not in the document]
kostko: Oh I see, yeah in that case you could make it configurable via the scenario arguments and then run a longer instance daily.

tests/consts.sh
```diff
@@ -1,5 +1,7 @@
 OASIS_CORE_VERSION=21.1-dev
-# To download a pre-released version of oasis-core built by oasis-core/release-dev action
-# set artifact ID to use here.
-# To download a released version of oasis-core unset this variable.
+# To download a pre-released version of oasis-core built by
+# oasis-core/release-dev action, set the artifact ID to use here.
+# You must set GITHUB_TOKEN as well, if it is unset, the release version
+# specified above will be downloaded.
```
pro-wh, line 5: is it really correct for the user to choose which version in this way? we used to use these prerelease versions because we needed some features from them, and it would have been incorrect to use the released version

tests/download-artifacts.sh
```diff
@@ -1,16 +1,21 @@
-#!/bin/sh -eux
+#!/usr/bin/env bash
+set -o nounset -o pipefail -o errexit
+
+# Get the configuration from the current working directory.
```
pro-wh, line 4: woa nice catch. do we intentionally use different versions in various tests?

tests/e2e/simplekvtest.go
```diff
@@ -490,3 +504,79 @@ func KVRewardsTest(log *logging.Logger, conn *grpc.ClientConn, rtc client.Runtim
...
+	}
+
+	if ok == 0 {
+		return fmt.Errorf("no generated transactions were submitted successfully")
+	}
+
+	// Note that submission errors are fine here, since we're going to get
+	// invalid nonce errors a lot, because the txs are generated in parallel.
+	// Transaction generation errors are also fine, since queries can fail
+	// due to yet nonexisting keys in the keyvalue storage, etc.
```
pro-wh, line 579: so this is more to test a network's stability?

tests/consts.sh
abukosek: By default, the release version given above doesn't exist anyways, so the download will fail.  I imagine that at some point the whole thing will stabilize enough that we will be able to use a released version instead -- it's very annoying to have to provide a github token to run the tests locally (this was my primary motivation for this change).

tests/download-artifacts.sh
abukosek: Yeah -- I initially fixed this to use the `tests/consts.sh` file everywhere and the ts-web tests failed, so I changed it to what it is now.  I think we should use the same config for everything, not sure why whoever wrote this made it working-directory-dependent.

tests/e2e/simplekvtest.go
abukosek: From what I understand, that was the goal, yes :)
Later, we could make a separate test using the txgen that does more intelligent insert/get/remove operations that should always succeed, so that it can fail whenever an error happens.

tests/e2e/txgen/txgen.go
```diff
@@ -0,0 +1,204 @@
...
+		return err
+	}
+
+	// Get current nonce for the signer's account.
+	ac := accounts.NewV1(rtc)
+	nonce, err := ac.Nonce(ctx, client.RoundLatest, types.NewAddress(signer.Public()))
+	if err != nil {
+		return err
+	}
+	tx.AppendSignerInfo(signer.Public(), nonce)
```
pro-wh, line 98: we're changing the signature of AppendSignerInfo to be more general. could you replace the new calls with AppendAuthSignature

.github/workflows/ci-longtest.yaml
```diff
@@ -0,0 +1,44 @@
...
+          sudo apt-get install bubblewrap build-essential libseccomp-dev libssl-dev protobuf-compiler
+
+      - name: Download artifacts
+        working-directory: tests/
+        run: ./download-artifacts.sh
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Run long end-to-end tests
+        run: ./tests/run-e2e.sh --e2e/test-runtime-simple-keyvalue.txgen.coins_per_acct=100000 --e2e/test-runtime-simple-keyvalue.txgen.num_accounts=100 --e2e/test-runtime-simple-keyvalue.txgen.duration=15m
```
kostko, line 40: Break this into multiple lines to make the arguments more clear.

kostko (approved):

tests/e2e/simplekvtest.go
abukosek: Filed as #152 :)

PR #139, opened 2021-05-15
runtime-sdk-macros: add handler registration macro

### LINKED ISSUE #24: Add procedural macro for module derivation
Currently there is quite some boilerplate involved with defining a module, it's internal API, callable method dispatch and query method dispatch. It would be cool if there was a procedural macro that made this simpler.

nhynes:
the syntax is

```rust
#[sdk::handlers(rpc_namespace = "module")]
impl Module {
  #[sdk::handler(name = "MyCallable")] // module.MyCallable
  fn callable(ctx: &TxContext, arg1: A1, arg2: A2) -> Result<T, E> {} // multiple args are deserialized as a tuple/array
  fn my_query(ctx: &DispatchContext, arg1: A1) -> Result<T, E> {} // inferred name is `module.MyQuery`
}
```

I still need to add tests, but it's working on the parcel paratime.

mostly resolves #24


kostko:
Rebase.

kostko (commented):
I suggest you split stuff related to the Rust client into a separate PR and we merge the macros first.

nhynes:
> I suggest you split stuff related to the Rust client into a separate PR and we merge the macros first.

Sounds good. This is mostly just the "Nick gets Parcel working branch" atm. Once I get the darn thing deployed, I'll make this PR more manageable.

The current highlights are:
* [the Parcel API using handler macros](https://github.com/oasislabs/parcel-readonly/blob/real-runtime/paratime/src/api.rs)
* [a thin wrapper for the generated client](https://github.com/oasislabs/parcel-readonly/blob/real-runtime/gateway/src/client.rs)
* [an implementation of the handler traits](https://github.com/oasislabs/parcel-readonly/blob/real-runtime/runtime/src/module.rs#L748)

I still need to add a rust signer to make the gateway actually useful. I'll lump that in here, but split that out eventually, too.

PR #144, opened 2021-05-19
runtime-sdk: actually call the approve_unverified_tx step

pro-wh:
those default implementations are silent but deadly :facepalm:


pro-wh:
it will actually decrease coverage to put this code to use and add tests for it? :open_mouth:

kostko (approved):

PR #145, opened 2021-05-20
runtime-sdk: use gas for authentication

### LINKED ISSUE #141: use gas per signature slot
> would be nice to also charge gas per signature slot

https://github.com/oasisprotocol/oasis-sdk/pull/118#pullrequestreview-661879404

pro-wh:
fixes #141

### DIFF
--- a/runtime-sdk/src/context.rs
+++ b/runtime-sdk/src/context.rs
@@ -15,7 +15,6 @@ use oasis_core_runtime::{
 
 use crate::{
     event::Event,
-    module::MethodRegistry,
     modules::core::Error,
     storage,
     types::{address::Address, message::MessageEventHookInvocation, transaction},
@@ -172,9 +171,6 @@ pub struct DispatchContext<'a> {
     pub(crate) io_ctx: Arc<IoContext>,
     pub(crate) logger: slog::Logger,
 
-    /// The runtime's methods, in case you need to look them up for some reason.
-    pub(crate) methods: &'a MethodRegistry,
-
     pub(crate) block_tags: Tags,
 
     /// Maximum number of messages that can be emitted.
@@ -188,11 +184,7 @@ pub struct DispatchContext<'a> {
 
 impl<'a> DispatchContext<'a> {
     /// Create a new dispatch context from the low-level runtime context.
-    pub(crate) fn from_runtime(
-        ctx: &'a RuntimeContext<'_>,
-        mkvs: &'a mut dyn mkvs::MKVS,
-        methods: &'a MethodRegistry,
-    ) -> Self {
+    pub(crate) fn from_runtime(ctx: &'a RuntimeContext<'_>, mkvs: &'a mut dyn mkvs::MKVS) -> Self {
         let mode = if ctx.check_only {
             Mode::CheckTx
         } else {
@@ -208,7 +200,6 @@ impl<'a> DispatchContext<'a> {
             io_ctx: ctx.io_ctx.clone(),
             logger: get_logger("runtime-sdk")
                 .new(o!("ctx" => "dispatch", "mode" => Into::<&'static str>::into(&mode))),
-            methods,
             block_tags: Tags::new(),
             max_messages: ctx.max_messages,
             messages: Vec::new(),
--- a/runtime-sdk/src/dispatcher.rs
+++ b/runtime-sdk/src/dispatcher.rs
@@ -105,11 +105,11 @@ impl<R: Runtime> Dispatcher<R> {
 
     fn dispatch_call(
         &self,
-        mut ctx: &mut TxContext<'_, '_>,
+        ctx: &mut TxContext<'_, '_>,
         call: types::transaction::Call,
     ) -> types::transaction::CallResult {
         // Use up gas for authenticating the transaction.
-        if let Err(e) = modules::core::Module::use_gas_for_auth(&mut ctx) {
+        if let Err(e) = modules::core::Module::use_gas_for_auth(ctx) {
             return e.to_call_result();
         }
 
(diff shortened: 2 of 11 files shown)


pro-wh:
why does the go e2e test still work with hardcoded 200 gas for kv insert?

pro-wh:
oh shoot, EstimateGas won't work, with the authentication step bypassed

pro-wh:
oh right, authenticate_tx isn't even called in a transaction context. I'll move it somewhere else

pro-wh:
I've moved it out into an ad-hoc direct call from dispatcher into the core module. it's now in a special `use_gas_for_auth` function.

next, I'm moving the gas estimation to be in the dispatcher, so that we can have a unified piece of code to (i) use gas for auth, (ii) look up the method, and (iii) call the handler. the result is two philosophical changes:

1. gas is no longer encapsulated within the core module. it is spread throughout the whole sdk
2. the core module is special, and the runtime calls into it directly. it may no longer be possible to build runtimes without this module

we now look up the method inside the transaction context. all I know is that the compiler is satisfied with that. any reason for it not to be in there?

also a little weird thing: this current draft would not be able to charge batch gas for attempting to authenticate a transaction and failing. getting it to do that would require some sort of multi-step thing to assess how much gas it will cost, increasing the batch gas, doing the authentication, then obtaining a transaction context and increasing the transaction gas. hm

pro-wh:
gas estimation is now moved

runtime-sdk/src/dispatcher.rs
```diff
@@ -346,10 +373,25 @@ impl<R: Runtime> transaction::dispatcher::Dispatcher for Dispatcher<R> {
         // TODO: Get rid of StorageContext (pass mkvs in ctx).
         StorageContext::with_current(|mkvs, _| {
             // Prepare dispatch context.
-            let mut ctx = DispatchContext::from_runtime(&ctx, mkvs, &self.methods);
+            let mut ctx = DispatchContext::from_runtime(&ctx, mkvs);
             // Perform state migrations if required.
             self.maybe_init_state(&mut ctx);
 
+            match method {
+                "core.EstimateGas" => {
```
pro-wh, line 381: note: our linter doesn't like this single-case match. but it feels so much like a match

also side note: I previously tried adding a method to the method registry for this, but the registry is built for static functions, where as this implementation accesses the dispatcher instance. and rust probably wouldn't be a good sport about having a self-referential closure.

runtime-sdk/src/context.rs
```diff
@@ -172,9 +171,6 @@ pub struct DispatchContext<'a> {
     pub(crate) io_ctx: Arc<IoContext>,
     pub(crate) logger: slog::Logger,
 
-    /// The runtime's methods, in case you need to look them up for some reason.
-    pub(crate) methods: &'a MethodRegistry,
```
kostko, line 176: This may actually be useful for thing like calling transactions on behalf of a smart contract etc.

runtime-sdk/src/dispatcher.rs
```diff
@@ -102,27 +103,53 @@ impl<R: Runtime> Dispatcher<R> {
...
-    ) -> Result<DispatchResult, Error> {
-        // Run pre-processing hooks.
-        if let Err(err) = R::Modules::authenticate_tx(ctx, &tx) {
-            return Ok(err.to_call_result().into());
+        mut ctx: &mut TxContext<'_, '_>,
+        call: types::transaction::Call,
+    ) -> types::transaction::CallResult {
+        // Use up gas for authenticating the transaction.
+        if let Err(e) = modules::core::Module::use_gas_for_auth(&mut ctx) {
+            return e.to_call_result();
```
kostko, line 113: Hm, any particular reason why you didn't just add this same call to the previous implementation of an "estimate gas" query instead of turning all the things around?

pro-wh: lower risk that they'd accidentally diverge if we make a mistake maintaining these two somewhat-far-away copies of the same steps

runtime-sdk/src/context.rs
pro-wh: been thinking about this, and there are two ideas that make this previous attempt look very hacky. but this previous attempt was the result of some seemingly incidental possibly temporary opportunity.

currently there's this awkward imbalance where the method registry use function pointers and is type-agnostic of the runtime, while other module-provided functionality (e.g. authenticate tx) uses static dispatch and thus depends on runtime type. overall, the dispatcher needs both, so it depends on the runtime type. we went with this give-them-the-method-registry-only ~hack which (i) was all we needed before tx auth gas and (ii) worked because we used function pointers for those methods.

the goal for these two ideas is to make dispatcher access more general, so we can additionally call the other things, such as a hypothetical new pre-transaction callback. we'd do this by passing the entire dispatcher into the method handler.

1. give handlers a dispatcher type parameter. this would let us pass a Dispatcher<R> into it without having to make the module as a whole know about R, which I think wouldn't work because R::Modules needs to contain the module.
2. make the whole dispatcher use function pointers and not depend on R. or other dynamic dispatch. then we'll be able to pass the dispatcher reference into the handler

kostko: Yeah, agreed on liking the static dispatch. Should be doable, just wondering how a bunch of `match` statements would compare to a map lookup? Probably doesn't matter unless you have lots of methods.

pro-wh: dispatch from within methods is reinstated in #147. and with that, I've moved gas to be encapsulated in the core module again. yay

pro-wh:
this is now inextricable from #150. there are no unsettle comments here, so closing. if I've missed any, please transfer them

PR #146, opened 2021-05-24
runtime-sdk: support for transaction priority and weights

ptrus:
Requires: https://github.com/oasisprotocol/oasis-core/pull/3965  https://github.com/oasisprotocol/oasis-core/pull/3988


kostko (commented):
One of the weights should also be gas usage. And the maximum should be based on the configured max batch gas.

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -268,6 +331,11 @@ impl module::MethodRegistrationHandler for Module {
             name: "core.EstimateGas",
             handler: Self::_query_estimate_gas_handler,
         });
+
+        methods.register_query(module::QueryMethodInfo {
+            name: "core.BatchWeightLimits",
```
kostko, line 336: May be useful to define this method name in `oasis-core-runtime` as it will be part of the Host API?

ptrus: yeah agreed

runtime-sdk/src/dispatcher.rs
```diff
@@ -50,13 +51,28 @@ pub enum Error {
 pub struct DispatchResult {
     pub result: types::transaction::CallResult,
     pub tags: Tags,
+    // Transaction priority.
```
kostko, line 54: Could change these to doc comments (e.g. `///`) and add them to other fields as well.

runtime-sdk/src/dispatcher.rs
```diff
@@ -18,14 +18,15 @@ use oasis_core_runtime::{
...
-    types::CheckTxResult,
+    types::{CheckTxMetadata, CheckTxResult, CheckTxWeight},
 };
 
 use crate::{
     context::{BatchContext, Context, RuntimeBatchContext},
     error::{Error as _, RuntimeError},
     module::{self, AuthHandler, BlockHandler, MethodHandler},
     modules,
+    modules::core::API,
```
kostko, line 29:
```suggestion
    modules::core::API as _,
```
(To avoid having a general "API" type imported.)

runtime-sdk/src/modules/accounts/mod.rs
```diff
@@ -521,6 +521,17 @@ impl module::AuthHandler for Module {
                 .add(&tx.auth_info.fee.amount);
 
             // TODO: Emit event that fee has been paid.
+
+            // Bump transaction priority.
```
kostko, line 525: Should this prioritize based on gas price instead of fee directly?

runtime-sdk/src/modules/consensus/mod.rs
```diff
@@ -152,6 +154,11 @@ impl API for Module {
     ) -> Result<(), Error> {
         Self::ensure_consensus_denomination(ctx, amount.denomination())?;
 
+        if ctx.is_check_only() {
+            Core::add_weight(ctx, CheckTxWeight::ConsensusMessages, 1)?;
```
kostko, line 158: Should this weight be added in any case not just during checks? Then we could reject transactions in case it exceeds the per-batch maximum values (because a bad proposer could still include such transactions)?

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -299,6 +367,9 @@ impl module::MethodHandler for Module {
                 let args = cbor::from_value(args).map_err(|_| Error::InvalidArgument)?;
                 Ok(cbor::to_value(&Self::query_estimate_gas(ctx, args)?))
             })()),
+            BATCH_WEIGHT_LIMIT_QUERY_METHOD => {
+                module::DispatchResult::Handled(Ok(cbor::to_value(&Self::get_weight_limits(ctx))))
+            }
```
kostko, line 372: Should we handle this in the dispatcher and make it so that any module can provide per-batch weight limits? E.g., we could extend `BlockHandler` with `get_block_weight_limits` and invoke this in the `Dispatcher` instead of the core module?

runtime-sdk/src/modules/consensus/mod.rs
ptrus: Hm yeah maybe? Although we already do have custom checks for messages and batch gas. Maybe we could remove those checks and generalize it into weight limits checks?

kostko: Yeah I think this makes sense.

runtime-sdk/src/modules/core/mod.rs
ptrus: Yeah makes sense, done

runtime-sdk/src/dispatcher.rs
```diff
@@ -107,11 +117,19 @@ impl<R: Runtime> Dispatcher<R> {
...
-        match R::Modules::dispatch_call(ctx, &call.method, call.body) {
-            module::DispatchResult::Handled(result) => result,
-            module::DispatchResult::Unhandled(_) => {
-                modules::core::Error::InvalidMethod.to_call_result()
+        match call.method.as_str() {
+            // Internal methods.
+            BATCH_WEIGHT_LIMIT_QUERY_METHOD => {
+                let block_weight_limits = R::Modules::get_block_weight_limits(ctx);
+                CallResult::Ok(cbor::to_value(block_weight_limits))
             }
```
kostko, line 114: Hm this doesn't seem right, it should be down below in the `query` handler?

runtime-sdk/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = "2018"
 license = "Apache-2.0"
 
 [dependencies]
-oasis-core-runtime = { git = "https://github.com/oasisprotocol/oasis-core", rev = "c287752" }
+oasis-core-runtime = { git = "https://github.com/oasisprotocol/oasis-core", rev = "c82aa4b0384c79cd0c0d834530f42a420b47a26e" }
```
kostko, line 9: Bump to tag `v21.2` now.

client-sdk/go/go.mod
```diff
@@ -12,8 +12,7 @@ replace (
 require (
 	github.com/btcsuite/btcd v0.21.0-beta
 	github.com/fxamacker/cbor/v2 v2.2.1-0.20210517032302-bdd38cd1c8c0 // indirect
-	github.com/oasisprotocol/oasis-core/go v0.2101.1-0.20210517160830-c287752b61b7
-	github.com/prometheus/common v0.24.0 // indirect
+	github.com/oasisprotocol/oasis-core/go v0.2101.1-0.20210604185237-c82aa4b0384c
```
kostko, line 15: Bump to `v0.2102.0`.

runtime-sdk/src/dispatcher.rs
ptrus: fixed

runtime-sdk/src/modules/accounts/mod.rs
ptrus: Yeah i guess that way it doesn't prioritize higher gas amount transactions? (although maybe it should?). Anyway, changed to use gas price instead.

runtime-sdk/src/modules/consensus/mod.rs
ptrus: I changed so that weights are added in all cases here, but i didn't add the per-batch weight checks. I have an initial implementation of the batch-limit checks, but I like it less than the current specialized gas accounting and message length checks. Maybe i'd suggest deferring on that until we add any more "weights" and just opening an issue for now.

I think that while "messages" and "gas" are the only "weights" that are used it doesn't make to much sense generalizing this. Manily since:
- gas accounting does a bit more than just checking batch weight limits (also tracks that the transaction amount spent is within the fee.gas_amount limit), so we have a special case here
- messages are quite first-class-y in the current code (e.g. both `Context` and `Dispatcher` are aware of messages), so i think it makes sense having the emitted message limits enforced the way it's done currently

tests/consts.sh
```diff
@@ -6,7 +6,7 @@
 
 # Released version from GitHub Releases.
 # e.g. '21.1.2'
-OASIS_CORE_VERSION=''
+OASIS_CORE_VERSION='21.2'
```
pro-wh, line 9: can we also bump BUILD_NUMBER to 5121

runtime-sdk/src/dispatcher.rs
```diff
@@ -249,6 +275,25 @@ impl<R: Runtime> Dispatcher<R> {
         ));
         store.insert(&modules::core::state::MESSAGE_HANDLERS, &message_handlers);
     }
+
+    /// Process the given runtime query.
+    pub fn query_method<C: BatchContext>(
```
kostko, line 280:
```suggestion
    pub fn dispatch_query<C: BatchContext>(
```

runtime-sdk/src/modules/consensus/mod.rs
kostko: Ah right, message limits are also enforced, ok agreed no need for a generalized mechanism.

kostko: Ah right, message limits are also enforced, ok agreed no need for a generalized mechanism.

tests/consts.sh
ptrus: bumped

runtime-sdk/src/modules/consensus_accounts/mod.rs
```diff
@@ -119,12 +119,13 @@ const CONSENSUS_WITHDRAW_HANDLER: &str = "consensus.WithdrawIntoRuntime";
...
     for Module<Accounts, Consensus>
 {
-    fn deposit<C: Context>(
+    fn deposit<C: TxContext>(
         ctx: &mut C,
         from: Address,
         amount: token::BaseUnits,
     ) -> Result<(), Error> {
         if ctx.is_check_only() {
+            Core::add_weight(ctx, TransactionWeight::ConsensusMessages, 1)?;
```
kostko, line 128: Also move these outside of `is_check_only` branch to be consistent.

runtime-sdk/src/types/transaction.rs
```diff
@@ -134,7 +137,14 @@ pub struct Fee {
     #[serde(rename = "gas")]
     pub gas: u64,
 }
-
+impl Fee {
```
kostko, line 140: Add empty line above.

runtime-sdk/src/types/transaction.rs
```diff
@@ -134,7 +137,14 @@ pub struct Fee {
...
+    pub fn gas_price(&self) -> Quantity {
```
kostko, line 141: Add documentation comment.

kostko (approved):

runtime-sdk/src/modules/consensus_accounts/mod.rs
ptrus: hm but in that case this will emit 2 times when not in check tx (once here, and once in the consensus method invocation)

kostko: Oh I see, ok. Maybe then just add a comment why this is so.

ptrus: makes sense, will add

PR #147, opened 2021-05-26
runtime-sdk: pass dispatcher to method handlers

pro-wh:
cross reference https://github.com/oasisprotocol/oasis-sdk/pull/145#discussion_r638707458

on top of #145

also of interest: #148 -- with static dispatch, even looking up a method needs the runtime type

this takes the "dispatcher has runtime type and module methods take a runtime type parameter" approach. now we give an entire dispatcher ref to query methods. this replaces having a copy of the method registry in the context.

we can do a similar conversion when we want to access the dispatcher from calls.

### DIFF
--- a/client-sdk/go/modules/core/core.go
+++ b/client-sdk/go/modules/core/core.go
@@ -1,0 +1,34 @@
+package core
+
+import (
+	"context"
+
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/client"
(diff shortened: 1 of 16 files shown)


pro-wh:
changing base to main to enable ci

pro-wh:
this is now inextricable from #150. there are no unsettle comments here, so closing. if I've missed any, please transfer them

PR #148, opened 2021-05-26
runtime-sdk: Static call/query/message result handler dispatch

kostko:
Refactors the way method dispatch is performed by removing the method
registries and replacing them with static dispatch in the form of a
MethodHandler trait which needs to be provided by modules.

Additionally the contexts are made more composable by making sure that
the we take Context as a generic type argument everywhere and improving
the way simulation mode is handled.

Each context is now also bound to a specific Runtime type to allow for
nested method dispatch.

### DIFF
--- a/runtime-sdk/src/context.rs
+++ b/runtime-sdk/src/context.rs
@@ -1,6 +1,6 @@
 //! Execution context.
 use core::fmt;
-use std::{any::Any, collections::BTreeMap, sync::Arc};
+use std::{any::Any, collections::BTreeMap, marker::PhantomData, sync::Arc};
 
 use io_context::Context as IoContext;
 use slog::{self, o};
@@ -15,9 +15,8 @@ use oasis_core_runtime::{
 
 use crate::{
     event::Event,
-    module::MethodRegistry,
     modules::core::Error,
-    storage,
+    runtime, storage,
     types::{address::Address, message::MessageEventHookInvocation, transaction},
 };
 
@@ -51,8 +50,10 @@ impl From<&Mode> for &'static str {
 
 /// Runtime SDK context.
 pub trait Context {
+    /// Runtime that the context is being invoked in.
+    type Runtime: runtime::Runtime;
     /// Runtime state output type.
-    type S: storage::Store;
+    type Store: storage::Store;
 
     /// Returns a logger.
     fn get_logger(&self, module: &'static str) -> slog::Logger;
@@ -77,7 +78,7 @@ pub trait Context {
     fn runtime_round_results(&self) -> &roothash::RoundResults;
 
     /// Runtime state store.
-    fn runtime_state(&mut self) -> &mut Self::S;
+    fn runtime_state(&mut self) -> &mut Self::Store;
 
     /// Consensus state.
     fn consensus_state(&self) -> &consensus::state::ConsensusState;
@@ -157,48 +158,91 @@ pub trait Context {
     fn take_tx_value<V>(&mut self, key: &'static str) -> Option<Box<V>>
     where
         V: Any + Default;
+
+    /// Executes a function with the transaction-specific context set.
+    fn with_tx<F, Rs>(&mut self, tx: transaction::Transaction, f: F) -> Rs
+    where
+        F: FnOnce(TxContext<'_, '_, Self::Runtime, Self::Store>, transaction::Call) -> Rs;
+
+    /// Executes a function in a simulation context.
+    ///
+    /// The simulation context collects its own messages and starts with an empty set of context
+    /// values.
+    fn with_simulation<F, Rs>(&mut self, f: F) -> Rs
+    where
+        F: FnOnce(
+            DispatchContext<'_, Self::Runtime, storage::OverlayStore<&mut Self::Store>>,
+        ) -> Rs;
 }
 
 /// Dispatch context for the whole batch.
-pub struct DispatchContext<'a> {
-    pub(crate) mode: Mode,
+pub struct DispatchContext<'a, R: runtime::Runtime, S: storage::Store> {
+    mode: Mode,
 
-    pub(crate) runtime_header: &'a roothash::Header,
-    pub(crate) runtime_round_results: &'a roothash::RoundResults,
-    pub(crate) runtime_storage: storage::MKVSStore<&'a mut dyn mkvs::MKVS>,
+    runtime_header: &'a roothash::Header,
+    runtime_round_results: &'a roothash::RoundResults,
+    runtime_storage: S,
     // TODO: linked consensus layer block
-    pub(crate) consensus_state: &'a consensus::state::ConsensusState,
-    pub(crate) epoch: consensus::beacon::EpochTime,
-    pub(crate) io_ctx: Arc<IoContext>,
-    pub(crate) logger: slog::Logger,
+    consensus_state: &'a consensus::state::ConsensusState,
+    epoch: consensus::beacon::EpochTime,
+    io_ctx: Arc<IoContext>,
+    logger: slog::Logger,
 
-    /// The runtime's methods, in case you need to look them up for some reason.
-    pub(crate) methods: &'a MethodRegistry,
-
-    pub(crate) block_tags: Tags,
+    block_tags: Tags,
 
     /// Maximum number of messages that can be emitted.
-    pub(crate) max_messages: u32,
+    max_messages: u32,
     /// Emitted messages.
-    pub(crate) messages: Vec<(roothash::Message, MessageEventHookInvocation)>,
+    messages: Vec<(roothash::Message, MessageEventHookInvocation)>,
 
     /// Per-context values.
-    pub(crate) values: BTreeMap<&'static str, Box<dyn Any>>,
+    values: BTreeMap<&'static str, Box<dyn Any>>,
+
+    _runtime: PhantomData<R>,
 }
 
-impl<'a> DispatchContext<'a> {
+impl<'a, R: runtime::Runtime, S: storage::Store> DispatchContext<'a, R, S> {
+    /// Create a new dispatch context.
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(
+        mode: Mode,
+        runtime_header: &'a roothash::Header,
+        runtime_round_results: &'a roothash::RoundResults,
+        runtime_storage: S,
+        consensus_state: &'a consensus::state::ConsensusState,
+        epoch: consensus::beacon::EpochTime,
+        io_ctx: Arc<IoContext>,
+        max_messages: u32,
+    ) -> Self {
+        Self {
+            mode,
+            runtime_header,
+            runtime_round_results,
+            runtime_storage,
+            consensus_state,
+            epoch,
+            io_ctx,
+            logger: get_logger("runtime-sdk")
+                .new(o!("ctx" => "dispatch", "mode" => Into::<&'static str>::into(&mode))),
+            block_tags: Tags::new(),
+            max_messages,
+            messages: Vec::new(),
+            values: BTreeMap::new(),
+            _runtime: PhantomData,
+        }
+    }
+
     /// Create a new dispatch context from the low-level runtime context.
     pub(crate) fn from_runtime(
         ctx: &'a RuntimeContext<'_>,
         mkvs: &'a mut dyn mkvs::MKVS,
-        methods: &'a MethodRegistry,
-    ) -> Self {
+    ) -> DispatchContext<'a, R, storage::MKVSStore<&'a mut dyn mkvs::MKVS>> {
         let mode = if ctx.check_only {
             Mode::CheckTx
         } else {
             Mode::ExecuteTx
         };
-        Self {
+        DispatchContext {
             mode,
             runtime_header: ctx.header,
             runtime_round_results: ctx.round_results,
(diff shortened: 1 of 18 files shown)


runtime-sdk/src/module.rs
```diff
@@ -8,137 +8,113 @@ use oasis_core_runtime::common::cbor;
...
+#[derive(Clone, Debug)]
+pub struct MessageResult {
+    pub event: MessageEvent,
+    pub context: cbor::Value,
+}
 
-    /// Looks up a previously registered callable method.
-    pub fn lookup_callable(&self, name: &str) -> Option<&CallableMethodInfo> {
-        self.callable_methods.get(name)
+pub trait Dispatcher {
```
pro-wh, line 41: could we consider a different name, Dispatcher -> MethodHandler
- for some kind of consistency with AuthHandler/BlockHandler/etc.
- to make the codebase easier to navigate by not duplicating dispatcher::Dispatcher's name

kostko: Agreed.

runtime-sdk/src/module.rs
```diff
@@ -1,153 +1,128 @@
...
-}
-
-/// Registry of methods exposed by the modules.
-pub struct MethodRegistry {
-    callable_methods: BTreeMap<&'static str, CallableMethodInfo>,
-    query_methods: BTreeMap<&'static str, QueryMethodInfo>,
+/// Result of invoking the method handler.
+pub enum DispatchResult<B, R> {
+    Handled(R),
+    Unhandled(B),
```
pro-wh, line 22: :information_source: `B` is used to move the body back to the caller

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -185,34 +186,19 @@ impl Module {
...
         args: transaction::Transaction,
     ) -> Result<u64, Error> {
-        let mi = ctx
-            .methods
-            .lookup_callable(&args.call.method)
-            .ok_or(Error::InvalidMethod)?;
-        ctx.with_tx_simulation(args, |mut tx_ctx, call| {
-            (mi.handler)(&mi, &mut tx_ctx, call.body);
+        ctx.with_simulation(|mut ctx| {
+            let _ = dispatcher::Dispatcher::<C::Runtime>::dispatch_tx(&mut ctx, args);
```
pro-wh, line 194: fully static dispatch! :raised_hands:

runtime-sdk/src/modules/core/test.rs
```diff
@@ -62,33 +62,76 @@ fn test_use_gas() {
...
+
+impl module::MethodHandler for GasWasterModule {
+    fn dispatch_call<C: Context>(
+        ctx: &mut C,
+        method: &str,
+        body: cbor::Value,
+    ) -> module::DispatchResult<cbor::Value, transaction::CallResult> {
+        match method {
+            Self::METHOD_WASTE_GAS => {
+                Core::use_gas(ctx, Self::MAX_GAS).expect("use_gas should succeed");
```
pro-wh, line 88: enterprise u64 addition :sunglasses:

runtime-sdk/src/testing/mock.rs
```diff
@@ -1,22 +1,34 @@
...
     types::transaction,
 };
 
+/// A mock runtime that only has the core module.
+pub struct EmptyRuntime;
+
+impl Runtime for EmptyRuntime {
+    const VERSION: Version = Version::new(0, 0, 0);
+
+    type Modules = modules::core::Module;
```
pro-wh, line 25: with, #147 I think we'll be okay to use `()` here, without even needing `core`

runtime-sdk/src/testing/mock.rs
```diff
@@ -25,32 +37,32 @@ pub struct Mock {
...
-            epoch: self.epoch,
-            io_ctx: IoContext::background().freeze(),
-            methods: &self.methods,
-            logger: get_logger("mock"),
-            block_tags: Tags::new(),
-            messages: Vec::new(),
-            max_messages: self.max_messages,
-            values: BTreeMap::new(),
-        }
+    pub fn create_ctx(
```
pro-wh, line 45: :information_source: existing uses of this don't re-dispatch anything, so they can use this empty-runtime context without any changes

pro-wh (commented):
looking good so far. I'm going to try merging this with my current work to see if I come up with any other feedback

runtime-sdk/src/module.rs
pro-wh: actually how much utility do we get from moving the body/message result into the handler? could consider converting to references while we have our sledgehammers out

pro-wh:
my merge in #150

runtime-sdk/src/module.rs
kostko: Yeah wanted to do that as well, but `cbor::from_value` wants to own the value so passing a reference won't work without a clone which I wanted to avoid.

kostko:
Cherry-picked some improvements from @nhynes where we have separate traits for `TxContext` and `BatchContext` to make things more clear and avoid unnecessary options/unwraps.

runtime-sdk/src/modules/accounts/mod.rs
```diff
@@ -353,85 +323,68 @@ impl Module {
...
-            handler: Self::_callable_transfer_handler,
-        });
+impl module::MethodHandler for Module {
+    fn dispatch_call<C: Context>(
+        ctx: &mut C,
+        method: &str,
+        body: cbor::Value,
+    ) -> module::DispatchResult<cbor::Value, CallResult> {
+        match method {
+            "accounts.Transfer" => {
```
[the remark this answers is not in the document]
kostko: The plan is to autogenerate these via a derive macro, see #139.

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -129,51 +132,50 @@ const CONTEXT_KEY_GAS_USED: &str = "core.GasUsed";
...
-
+    fn use_batch_gas<C: Context>(ctx: &mut C, gas: u64) -> Result<(), Error> {
         let batch_gas_limit = Self::params(ctx.runtime_state()).max_batch_gas;
-        let batch_gas_used = ctx.value::<u64>(CONTEXT_KEY_GAS_USED);
+        let batch_gas_used = ctx.value::<u64>(CONTEXT_KEY_GAS_USED).or_default();
         let batch_new_gas_used = batch_gas_used
             .checked_add(gas)
             .ok_or(Error::BatchGasOverflow)?;
         if batch_new_gas_used > batch_gas_limit {
+            println!("batch gas limit: {:?}", batch_gas_limit);
```
pro-wh, line 155: do we have some logging facility more formal than this

runtime-sdk/src/context.rs
```diff
@@ -130,75 +120,116 @@ pub trait Context {
...
+            RuntimeBatchContext<'_, Self::Runtime, storage::OverlayStore<&mut Self::Store>>,
+        ) -> Rs;
+}
 
-    /// Fetches or sets a value associated with the transaction.
-    ///
-    /// Only present if this is a transaction processing context.
-    fn tx_value<V>(&mut self, key: &'static str) -> Option<&mut V>
+/// Runtime SDK batch-wide context.
+pub trait BatchContext: Context {
```
pro-wh, line 138: what other things do you envision will impl this and TxContext?

pro-wh (commented):
cool. in practice we had several places that would unwrap tx-specific operations, so those ones might as well be separate traits

but I thought we were trying to unify the context so that we could have operations that we could use in and out of transactions. we might reasonably use gas in these operations, and now Core::use_gas is bifurcated.

runtime-sdk/src/module.rs
pro-wh: o

kostko:
> but I thought we were trying to unify the context so that we could have operations that we could use in and out of transactions. we might reasonably use gas in these operations, and now Core::use_gas is bifurcated.

That's a fair point, @nhynes any thoughts on this? I agree that having a unified context could be useful. @pro-wh do you have any specific use cases in mind?

runtime-sdk/src/context.rs
nhynes: I don't think anything will. The trait could be turned into an `impl` if you don't think there'll be mocked versions ever (I think the MKVS is a better mock target, though).

runtime-sdk/src/modules/core/mod.rs
nhynes: https://crates.io/crates/tracing is good

nhynes:
> unified context could be useful

I agree that it'd be clean, but is it really true that once you downcast into a `TxContext` you can't do batch contextual operations? If you `use_tx_gas` within a transaction, does it always `use_batch_gas`? Alternatively, you can have `tx_ctx.batch_context().use_gas(...)`. All I know is that:
* static types > runtime types
* this is a low-level SDK (by virtue of there being no others) and it might be useful to be able to off-road batch vs tx operations (big maybe)
* unification is backwards-compatible

> operations that we could use in and out of transactions

That describes `trait Context` in this PR. The gas accounting thing is regrettable, but so is code that does a dynamic check for being in a transaction context.

runtime-sdk/src/modules/core/mod.rs
kostko: Yes, the context has a `slog` logging facility, see the `get_logger` method.

kostko: The `tracing` crate looks nice though, maybe we should use it instead of `slog`.

kostko:
> but is it really true that once you downcast into a TxContext you can't do batch contextual operations

The only batch operation is entering a transaction context via `with_tx`. I think that it doesn't make much sense to allow `with_tx` inside a transaction context unless you use `with_simulation` which gives you a simulated batch context.

runtime-sdk/src/context.rs
ptrus: It looks to me that that the goal here is to have have a trait (`TxContext`) and struct (`RuntimeTxContext`) be tightly coupled, and then we can use the trait where it fits nicely (static dispatch, generic functions - e.g. all the generic module transaction methods which accept `TxContext`) and the struct in places where we want to avoid dynamic dispatch/trait objects which tend to be somewhat limited/awkward to work with? e.g. bellow in the `with_tx` `FnOnce` argument.
Is this somewhat correct?

ptrus (commented):
Looks good to me

runtime-sdk/src/context.rs
kostko: Actually as @nhynes mentioned we could even get rid of the more specific tx/batch traits and just use concrete types instead. The only reason why you would want traits is if you had different implementations for them which seems unlikely in this case (e.g., you wouldn't mock contexts).

kostko: Oh but these not being traits makes it much less convenient as then you need to transfer around even more generic arguments that associated types conveniently hide for you.

ptrus: I think what i'm asking is why does the `with_tx` method bellow use `RuntimeTxContext` and not the `TxContext` trait in the `FnOnce`? 
I think it's cause using a trait object would be quite inconvenient, or is there any other reason

kostko: > I think it's cause using a trait object would be quite inconvenient, or is there any other reason?

Yeah, so a trait object would be impossible due to generic type parameters and closures cannot have generic type parameters AFAIK.

runtime-sdk/src/modules/core/mod.rs
pro-wh, line 194: https://github.com/oasisprotocol/oasis-sdk/pull/150#discussion_r641992976

runtime-sdk/src/dispatcher.rs
```diff
@@ -62,29 +61,19 @@ impl From<types::transaction::CallResult> for DispatchResult {
...
-
     _runtime: PhantomData<R>,
 }
 
 impl<R: Runtime> Dispatcher<R> {
-    pub(super) fn new(
-        methods: MethodRegistry,
-        consensus_message_handlers: MessageHandlerRegistry,
-    ) -> Self {
+    pub(super) fn new() -> Self {
```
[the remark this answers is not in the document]
kostko: Done.

runtime-sdk/src/modules/core/mod.rs
nhynes:
```suggestion
          dispatcher::Dispatcher::<C::Runtime>::dispatch_tx(&mut ctx, args).ok();
```

kostko:
Any further blockers before merging?

pro-wh (approved):
About the context split: I'm fine with this at this time, no disrespect to our previous attempt to unify them. If we later want a unified use_gas_whichever method, we're one step away from adding that to the Context trait and implementing that in the classes.

About the single-implementation TxContext and BatchContext traits: Thanks for bringing up that it makes it simpler to write. At this moment I would still prefer to save the compile time indirection for the benefit of human readability at the cost of fiddliness for writers. But it'll be good for the future where we can add stuff to the context without adding more and more generic parameters everywhere. Kinda lame how `with_tx` prevents us from actually mocking these traits though.

So yeah let's go with this

PR #193, opened 2021-06-07
test artifacts changes

### LINKED ISSUE #186: tests: switch to buildkite artifacts
github actions requires authentication to download artifacts. this change would make testing easier

### LINKED ISSUE #187: tests: explicitly fail when we need dev artifact but github token is not available
currently in this case we try to download a "-dev" release from github releases, which fails due to that release not existing. we can make the diagnostics better by detecting this condition

### LINKED ISSUE #188: tests: version the artifacts we download
without this, over time local tests will start breaking because all we do is check that oasis-node et al. exist, not that they are the right version. short of calling into them to ask them what version they are, we can leave the oasis_core_xx_linux_amd64 dir segment intact. we can add something similar for dev artifacts

### LINKED ISSUE #189: tests: unify artifact store for go e2e tests and ts-web tests
they both download copies of oasis-node and other software

### LINKED ISSUE #192: tests: default artifact paths for run-e2e.sh
would be nice to use `TEST_NODE_BINARY` et al. from where download-artifacts.sh puts things when no custom path is set

pro-wh:
fixes #186 
fixes #187 
fixes #188 
fixes #189 
fixes #192 

Consolidate our various consts.sh and download-artifacts.sh. We now have a single location for oasis-core artifacts, in tests/untracked. The ts-web/core test is updated to work with the version we use in other tests. Going forward, we should try to make sure the entire SDK works with some version of oasis-core, not for different parts to require different versions.

Save these assets into directories indicating what version they are, e.g. github-<artifact id> or buildkite-<build number>. Previously, we had to clear out the old artifacts manually when we change the version under test. And we didn't have a medium to tell developers to perform this step. Now the script will see that the desired version is missing and download it.

Experimental: resolving Buildkite artifact URLs is moved to be done at download-artifacts time, which makes it less tedious to bump the version used to test. As a trade-off, developers need to have `jq` installed.

Requesting comments: a commit in this PR changes the tests to use artifacts from Buildkite instead of from GitHub Actions. See the rationale in #186. We didn't record any rationale when we switched from Buildkite to GitHub Actions, so feel free to bring that up here, and we can decide.


kostko (approved):
Makes sense to me.

kostko:
@pro-wh merging this early as it helps making the Oasis Core bump easier.

PR #198, opened 2021-06-08
ts-web/core: update to oasis-core v21.2

pro-wh:
unfinished


pro-wh:
image

image

let's make a breaking change and switch over our capitalization ):

pro-wh:
:shrug: `cargo test` works locally. what else could it be?

kostko (approved):

kostko:
>  `cargo test` works locally. what else could it be?

Weird, looks like a buggy action?

kostko:
Oh right, this is usually the error when it can't find the specified release for tarpaulin, but [`0.18.0-alpha3` does exist](https://github.com/xd009642/tarpaulin/releases/tag/0.18.0-alpha3) :thinking:

kostko:
Seems to be working now, maybe it was due to the CDN issues yesterday.

PR #204, opened 2021-06-09
runtime-sdk: Add invariant checker

### LINKED ISSUE #18: Invariant checker
Each Runtime SDK module should be able to define an invariant checker (e.g., by introducing a new `InvariantHandler` trait) that can optionally be invoked on the state checking for any invariant violations. To invoke the invariant checker one would use a special query defined by the `core` module.

Examples of invariants for the accounts module include:
* All account balances should sum up to the total supply.

Examples of invariants for the consensus layer interaction module (#11) include:
* Total supply of the designated consensus layer token denomination should be less than or equal to the balance of the general runtime's account in the consensus layer.

In E2E tests we should then invoke the invariant checker query.

abukosek:
Closes #18.

### DIFF
--- a/runtime-sdk/src/context.rs
+++ b/runtime-sdk/src/context.rs
@@ -11,9 +11,10 @@ use io_context::Context as IoContext;
 use slog::{self, o};
 
 use oasis_core_runtime::{
-    common::logger::get_logger,
+    common::{logger::get_logger, namespace::Namespace},
     consensus,
     consensus::roothash,
+    protocol::HostInfo,
     storage::mkvs,
     transaction::{context::Context as RuntimeContext, tags::Tags},
 };
@@ -74,6 +75,14 @@ pub trait Context {
     /// Whether the transaction is just being simulated.
     fn is_simulation(&self) -> bool {
         self.mode() == Mode::SimulateTx
+    }
+
+    /// Information about the host environment.
+    fn host_info(&self) -> &HostInfo;
+
+    /// Runtime ID.
+    fn runtime_id(&self) -> &Namespace {
+        &self.host_info().runtime_id
     }
 
     /// Last runtime block header.
@@ -159,6 +168,7 @@ pub trait TxContext: Context {
 pub struct RuntimeBatchContext<'a, R: runtime::Runtime, S: storage::Store> {
     mode: Mode,
 
+    host_info: &'a HostInfo,
     runtime_header: &'a roothash::Header,
     runtime_round_results: &'a roothash::RoundResults,
     runtime_storage: S,
@@ -186,6 +196,7 @@ impl<'a, R: runtime::Runtime, S: storage::Store> RuntimeBatchContext<'a, R, S> {
     #[allow(clippy::too_many_arguments)]
     pub fn new(
         mode: Mode,
+        host_info: &'a HostInfo,
         runtime_header: &'a roothash::Header,
         runtime_round_results: &'a roothash::RoundResults,
         runtime_storage: S,
@@ -196,6 +207,7 @@ impl<'a, R: runtime::Runtime, S: storage::Store> RuntimeBatchContext<'a, R, S> {
     ) -> Self {
         Self {
             mode,
+            host_info,
             runtime_header,
             runtime_round_results,
             runtime_storage,
@@ -216,6 +228,7 @@ impl<'a, R: runtime::Runtime, S: storage::Store> RuntimeBatchContext<'a, R, S> {
     pub(crate) fn from_runtime(
         ctx: &'a RuntimeContext<'_>,
         mkvs: &'a mut dyn mkvs::MKVS,
+        host_info: &'a HostInfo,
     ) -> RuntimeBatchContext<'a, R, storage::MKVSStore<&'a mut dyn mkvs::MKVS>> {
         let mode = if ctx.check_only {
             Mode::CheckTx
@@ -224,6 +237,7 @@ impl<'a, R: runtime::Runtime, S: storage::Store> RuntimeBatchContext<'a, R, S> {
         };
         RuntimeBatchContext {
             mode,
+            host_info,
             runtime_header: ctx.header,
             runtime_round_results: ctx.round_results,
             runtime_storage: storage::MKVSStore::new(ctx.io_ctx.clone(), mkvs),
@@ -251,6 +265,10 @@ impl<'a, R: runtime::Runtime, S: storage::Store> Context for RuntimeBatchContext
 
     fn mode(&self) -> Mode {
         self.mode
+    }
+
+    fn host_info(&self) -> &HostInfo {
+        &self.host_info
     }
 
     fn runtime_header(&self) -> &roothash::Header {
@@ -300,6 +318,7 @@ impl<'a, R: runtime::Runtime, S: storage::Store> Context for RuntimeBatchContext
 
         let sim_ctx = RuntimeBatchContext {
             mode: Mode::SimulateTx,
+            host_info: self.host_info,
             runtime_header: self.runtime_header,
             runtime_round_results: self.runtime_round_results,
             runtime_storage: store,
@@ -333,6 +352,7 @@ impl<'a, R: runtime::Runtime, S: storage::Store> BatchContext for RuntimeBatchCo
 
         let tx_ctx = RuntimeTxContext {
             mode: self.mode,
+            host_info: self.host_info,
             runtime_header: self.runtime_header,
             runtime_round_results: self.runtime_round_results,
             consensus_state: self.consensus_state,
@@ -372,6 +392,7 @@ impl<'a, R: runtime::Runtime, S: storage::Store> BatchContext for RuntimeBatchCo
 pub struct RuntimeTxContext<'round, 'store, R: runtime::Runtime, S: storage::Store> {
     mode: Mode,
 
+    host_info: &'round HostInfo,
     runtime_header: &'round roothash::Header,
     runtime_round_results: &'round roothash::RoundResults,
     consensus_state: &'round consensus::state::ConsensusState,
@@ -413,6 +434,10 @@ impl<'round, 'store, R: runtime::Runtime, S: storage::Store> Context
 
     fn mode(&self) -> Mode {
         self.mode
+    }
+
+    fn host_info(&self) -> &HostInfo {
+        &self.host_info
     }
 
     fn runtime_header(&self) -> &roothash::Header {
@@ -463,6 +488,7 @@ impl<'round, 'store, R: runtime::Runtime, S: storage::Store> Context
 
         let sim_ctx = RuntimeBatchContext {
             mode: Mode::SimulateTx,
+            host_info: self.host_info,
             runtime_header: self.runtime_header,
             runtime_round_results: self.runtime_round_results,
             runtime_storage: store,
--- a/runtime-sdk/src/module.rs
+++ b/runtime-sdk/src/module.rs
@@ -256,6 +256,24 @@ impl BlockHandler for Tuple {
     }
 }
 
+/// Invariant handler.
+pub trait InvariantHandler {
+    /// Check invariants.
+    fn check_invariants<C: Context>(_ctx: &mut C) -> Result<(), modules::core::Error> {
+        // Default implementation doesn't do anything.
+        Ok(())
+    }
+}
+
+#[impl_for_tuples(30)]
+impl InvariantHandler for Tuple {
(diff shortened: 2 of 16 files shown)


runtime-sdk/src/context.rs
```diff
@@ -318,6 +334,10 @@ impl<'a, R: runtime::Runtime, S: storage::Store> Context for RuntimeBatchContext
         };
         f(sim_ctx)
     }
+
+    fn check_invariants(&mut self) -> bool {
```
ptrus, line 338: Would it be nicer if this returned a `Result<(), Error>`. The invariant checks can then return more descriptive errors.

abukosek: Yeah, that was my intention initially as well... but, how do you properly merge the Errors of all InvariantHandlers for all modules? (see `impl InvariantHandler for Tuple` in `runtime-sdk/src/module.rs`)
If I had used the same trick as was used a few lines above for the Genesis parameters in the migration handler, the error codes would be mixed up (e.g. two modules can define error code 2 to mean different errors).
I suppose we could have errors without codes for this or come up with some other scheme?

pro-wh (commented):
the runtime ID is added to the dispatcher and contexts. this is so that the runtime can ask the host for its own balance.

it makes the dispatcher non-empty. so now there's a little weirdness where some methods need the `&self` (e.g. the great big `execute_batch` entry point) and some that don't (e.g. actually dispatching).

runtime-sdk/src/context.rs
abukosek: OK, I'll try it out.

runtime-sdk/src/modules/accounts/mod.rs
```diff
@@ -160,6 +166,31 @@ pub static ADDRESS_COMMON_POOL: Lazy<Address> =
...
+    MalformedDenomination,
+}
+
+impl std::convert::TryFrom<&[u8]> for AddressWithDenomination {
+    type Error = AWDError;
+
+    fn try_from(bytes: &[u8]) -> Result<Self, Self::Error> {
+        let address = Address::try_from(&bytes[..address::ADDRESS_SIZE])
+            .map_err(|_| AWDError::MalformedAddress)?;
+        let denomination = token::Denomination::try_from(&bytes[address::ADDRESS_SIZE..])
```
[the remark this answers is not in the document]
abukosek: No :)

runtime-sdk/src/modules/accounts/mod.rs
```diff
@@ -602,3 +675,32 @@ impl module::BlockHandler for Module {
...
+            for (den, amt) in bals {
+                computed_ts
+                    .entry(den.clone())
+                    .and_modify(|a| *a += amt)
+                    .or_insert_with(|| amt.clone());
+            }
+        }
+
+        for (den, ts) in &total_supplies {
+            let computed = computed_ts.remove(&den).expect("unexpected total supply");
```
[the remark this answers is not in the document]
abukosek: Ah, you're right, I'll fix this, thanks.

runtime-sdk/src/module.rs
```diff
@@ -256,6 +258,25 @@ impl BlockHandler for Tuple {
     }
 }
 
+/// Invariant handler.
+pub trait InvariantHandler {
+    /// Check invariants.
+    fn check_invariants<C: Context>(_ctx: &mut C) -> Result<(), CoreError> {
```
kostko, line 264:
```suggestion
    fn check_invariants<C: Context>(_ctx: &mut C) -> Result<(), modules::core::Error> {
```
(To be consistent with other places.)

runtime-sdk/src/modules/accounts/mod.rs
```diff
@@ -160,6 +166,31 @@ pub static ADDRESS_COMMON_POOL: Lazy<Address> =
...
+pub struct AddressWithDenomination(Address, token::Denomination);
+
+#[derive(Error, Debug)]
+pub enum AWDError {
+    #[error("malformed address")]
+    MalformedAddress,
+
+    #[error("malformed denomination")]
+    MalformedDenomination,
+}
```
kostko, line 180: This is only used internally, make it private.

runtime-sdk/src/types/address.rs
```diff
@@ -72,6 +77,15 @@ impl Address {
         )
     }
 
+    /// Creates a new runtime address.
+    pub fn from_runtime_id(id: &Namespace) -> Self {
```
kostko, line 81: Add tests to ensure consistency with consensus layer address derivation.

runtime-sdk/src/runtime.rs
```diff
@@ -46,6 +48,12 @@ pub trait Runtime {
         }
     }
 
+    /// Check invariants in all modules.
+    /// Returns true if everything is OK, false otherwise.
+    fn check_invariants<C: Context>(ctx: &mut C) -> Result<(), CoreError> {
```
kostko, line 53: Why do we need this forward function?

runtime-sdk/src/modules/accounts/mod.rs
```diff
@@ -300,6 +364,15 @@ impl API for Module {
             balances: account.iter().collect(),
         })
     }
+
+    fn get_total_supplies<S: storage::Store>(
```
kostko, line 368: Add some unit tests for this.

runtime-sdk/src/modules/accounts/mod.rs
```diff
@@ -602,3 +675,54 @@ impl module::BlockHandler for Module {
         }
     }
 }
+
+impl module::InvariantHandler for Module {
+    /// Check invariants.
+    fn check_invariants<C: Context>(ctx: &mut C) -> Result<(), CoreError> {
```
kostko, line 681: Test this in a unit test.

runtime-sdk/src/modules/accounts/mod.rs
```diff
@@ -228,6 +259,39 @@ impl Module {
         total_supplies.insert(amount.denomination(), &total_supply);
         Ok(())
     }
+
+    /// Get all balances.
+    fn get_all_balances<S: storage::Store>(
```
kostko, line 264: Add unit test for this.

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -83,6 +83,10 @@ pub enum Error {
     #[error("multisig too many signers")]
     #[sdk_error(code = 16)]
     MultisigTooManySigners,
+
+    #[error("invariants check failed: {0}")]
+    #[sdk_error(code = 17)]
+    InvChkFailed(&'static str),
```
kostko, line 89:
```suggestion
    #[error("invariant violation: {0}")]
    #[sdk_error(code = 17)]
    InvariantViolation(String),
```
(Use `String` so it is possible to use something like `format!`.)

runtime-sdk/src/types/address.rs
```diff
@@ -5,15 +5,16 @@ use bech32::{self, FromBase32, ToBase32, Variant};
...
+    consensus::address::Address as ConsensusAddress,
 };
 
 use crate::crypto::{multisig, signature::PublicKey};
 use oasis_core_runtime::common::cbor;
 
 const ADDRESS_VERSION_SIZE: usize = 1;
 const ADDRESS_DATA_SIZE: usize = 20;
-const ADDRESS_SIZE: usize = ADDRESS_VERSION_SIZE + ADDRESS_DATA_SIZE;
+pub const ADDRESS_SIZE: usize = ADDRESS_VERSION_SIZE + ADDRESS_DATA_SIZE;
```
kostko, line 17: Could instead expose this as `Address::SIZE` for convenience.

runtime-sdk/src/context.rs
```diff
@@ -76,6 +76,9 @@ pub trait Context {
         self.mode() == Mode::SimulateTx
     }
 
+    /// Runtime ID.
+    fn runtime_id(&self) -> &Namespace;
```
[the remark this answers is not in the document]
abukosek: I can add this, but I'd still like to keep the `runtime_id()` method, since it will probably see much more use than the rest of the host info.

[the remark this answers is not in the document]
kostko: Sure, that's fine, it should just be a default implementation on the trait that uses `self.host_info()` for doing the actual work. And so the contexts only need to store a reference to `host_info` instead of a separate one to runtime ID.

abukosek:
Thanks for the review -- I've added the unit tests and other things, so this is ready for another look :)

runtime-sdk/src/context.rs
```diff
@@ -76,6 +77,14 @@ pub trait Context {
         self.mode() == Mode::SimulateTx
     }
 
+    /// Host info.
```
kostko, line 80:
```suggestion
    /// Information about the host environment.
```

runtime-sdk/src/modules/accounts/mod.rs
```diff
@@ -602,3 +674,58 @@ impl module::BlockHandler for Module {
...
+            match computed_ts.remove(&den) {
+                Some(c) => computed = c,
+                // Total supplies have a denomination that we didn't encounter
+                // when computing total supplies based on account balances.
+                None => {
+                    return Err(CoreError::InvariantViolation(
+                        "unexpected denomination".to_string(),
+                    ))
+                }
+            }
```
kostko, line 711: You could do something like (preserve the comment ofc):
```suggestion
            let computed = computed_ts.remove(&den).ok_or(CoreError::InvariantViolation(...))?;
```

runtime-sdk/src/modules/consensus_accounts/mod.rs
```diff
@@ -396,3 +396,35 @@ impl<Accounts: modules::accounts::API, Consensus: modules::consensus::API> modul
...
+    for Module<Accounts, Consensus>
+{
+    /// Check invariants.
+    fn check_invariants<C: Context>(ctx: &mut C) -> Result<(), CoreError> {
+        // Total supply of the designated consensus layer token denomination
+        // should be less than or equal to the balance of the runtime's general
+        // account in the consensus layer.
+
+        let den = Consensus::consensus_denomination(ctx).unwrap();
+        let ts = Accounts::get_total_supplies(ctx.runtime_state()).unwrap();
```
kostko, line 410: This shouldn't panic, map to an error instead.

tests/e2e/runtime.go
```diff
@@ -319,6 +339,14 @@ func (sc *RuntimeScenario) Run(childEnv *env.Env) error {
 			return testErr
 		}
 		sc.Logger.Info("test passed", "test", testName)
+
+		// Do an invariants check after each test.
```
kostko, line 343: Can you also make this part of the txgen fuzzing test(s) to make sure invariants are being maintained?

runtime-sdk/src/testing/mock.rs
```diff
@@ -1,9 +1,11 @@
 //! Mock dispatch context for use in tests.
 use io_context::Context as IoContext;
+use std::collections::BTreeMap;
```
kostko, line 3: Std imports should be in its own group at the top.

tests/e2e/txgen/txgen.go
```diff
@@ -195,6 +209,11 @@ func Generate(ctx context.Context, rtc client.RuntimeClient, rng *rand.Rand, acc
 						atomic.AddUint64(&okCount, 1)
 					}
 				}
+
+				// Perform an invariants check.
+				if err := CheckInvariants(ctx, rtc); err != nil {
+					atomic.AddUint64(&invErrCount, 1)
```
kostko, line 215: An invariant check failing should halt the test with an error immediately.

abukosek: Fixed! :)

tests/e2e/txgen/txgen.go
```diff
@@ -195,6 +212,11 @@ func Generate(ctx context.Context, rtc client.RuntimeClient, rng *rand.Rand, acc
 						atomic.AddUint64(&okCount, 1)
 					}
 				}
+
+				// Perform an invariants check.
+				if err := CheckInvariants(ctx, rtc); err != nil {
+					errCh <- err
```
kostko, line 218: Maybe do something like this so that if multiple goroutines report an error, this doesn't result in blocked goroutines:
```suggestion
					select {
					case errCh <- err:
					default:
					}
```

abukosek: Good catch, thanks!  I've fixed it now.

kostko (approved):

PR #208, opened 2021-06-11
ts-web/core: update to oasis-core v21.2.1

pro-wh:
new staking debonding_start event


ptrus (approved):
Should also update dependencies and test const files?

kostko (approved):

kostko:
> Should also update dependencies and test const files?

Yeah and in this case we can get rid of the `oasis-net-runner` "hack" for disabling runtimes as this is fixed in 21.2.1.

pro-wh:
thanks, I've filed #210 to track updating the other parts

PR #215, opened 2021-06-13
Add context to error messages

nhynes:
This was useful for me to determine that txs require at least one signature, but I don't really know why that's true.


kostko:
This also seems to include parts of #216? Can you remove that part?

kostko (approved):

PR #216, opened 2021-06-13
Allow runtimes to accept unsigned transactions

nhynes:
Say the runtime is inside of a VPC with the gateway. Decentralized? No, but at least I don't need to worry about managing useless secrets. Also, this is useful for development.

re: #215

### DIFF
--- a/runtime-sdk/src/dispatcher.rs
+++ b/runtime-sdk/src/dispatcher.rs
@@ -103,7 +103,7 @@ impl<R: Runtime> Dispatcher<R> {
 
         // Verify transaction signatures.
         // TODO: Support signature verification of the whole transaction batch.
-        utx.verify()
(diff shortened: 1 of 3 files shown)


pro-wh:
I think we have some code that reaches for `signer_info[0]`, e.g. the part that takes the fee. should we do something about something? least work idea: we leave a vague warning on `ALLOW_UNSIGNED_TXS` flag that some stuff might break

nhynes:
> I think we have some code that reaches for signer_info[0], e.g. the part that takes the fee. should we do something about something?

afaik that's only if you're choosing to use the accounts module. perhaps I can add `#[doc(hidden)]` so that it's available, but not obvious? I'll add an explanatory comment, too.

kostko:
With #441 this can now be provided by a module that can support an arbitrary transaction encoding and signature scheme.

PR #218, opened 2021-06-13
runtime-sdk: begin and end block during check batch

nhynes:
This PR makes it more straightforward to do setup/teardown actions without knowing whether the runtime is being invoked in check vs execute mode.

### DIFF
--- a/runtime-sdk/src/dispatcher.rs
+++ b/runtime-sdk/src/dispatcher.rs
@@ -361,11 +361,17 @@ impl<R: Runtime> transaction::dispatcher::Dispatcher for Dispatcher<R> {
             // Perform state migrations if required.
             R::migrate(&mut ctx);
 
+            // Run begin block hooks.
+            R::Modules::begin_block(&mut ctx);
+
             // Check the batch.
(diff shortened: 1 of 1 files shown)


kostko (commented):
This doesn't seem right, can you explain what the use case is?

nhynes:
> can you explain what the use case is?

I load a database before a batch, and sync the db after a batch. It's not feasible to create the database in the middle of the batch, and is actually unsafe to do so because the state being passed in lives no longer than the batch, so must be destroyed afterward.

It doesn't need to be `begin_block` but it needs to happen in about the same place. This patch is working well for me so far, though.

kostko:
My concern is that this could result in erroneous execution due to the handlers being invoked multiple times on the same state for the same round.

nhynes:
Shall I add another lifecycle hook roughly named `before_handlers/after_handlers`? I needed to add the same begin/end_batch invocation to the queries handler, too, for the same reason.

kostko:
Yeah I think a separate hook would make sense with the explicit mention what the difference to the block hooks is. Because this seems more like some "prepare context" and "teardown context" or something?

PR #220, opened 2021-06-14
ts-web: add browser extension utilities

pro-wh:
summary

a browser extension "signer interface." where a browser extension can serve the role of a ContextSigner implementation.

---

update zero

Turns out there's a feature `web_accessible_resources`, which will allow us to make a webpage that sites can load into an iframe and then message. I've written up some code to verify that this works.

Next I'll be writing some helper functions to interact with the extension within the `ContextSigner` interface.


pro-wh:
I've extracted some code into a library. It's missing some things we usually expect--tests, package name. But it's at a point where reviewers can take a look to get an impression of the overall design

client-sdk/ts-web/ext-utils/src/protocol.ts
```diff
@@ -0,0 +1,69 @@
...
+
+// methods
+
+export const METHOD_CONTEXT_SIGNER_PUBLIC = 'context-signer-public';
+
+export interface ContextSignerPublicRequest {
+    method: typeof METHOD_CONTEXT_SIGNER_PUBLIC;
+    /**
+     * An extra parameter, suggested to be used to help select which key to
+     * use. This lirbary passes it through verbatim; it's up to the extension
```
kostko, line 38:
```suggestion
     * use. This library passes it through verbatim; it's up to the extension
```

client-sdk/ts-web/ext-utils/src/protocol.ts
```diff
@@ -0,0 +1,69 @@
...
+     * to define how it's interpreted.
+     */
+    which: any;
```
kostko, line 41: Could be useful to have a method for listing keys as well?

client-sdk/ts-web/ext-utils/src/protocol.ts
```diff
@@ -0,0 +1,69 @@
...
+    public_key: Uint8Array;
+}
+
+export const METHOD_CONTEXT_SIGNER_SIGN = 'context-signer-sign';
+
+export interface ContextSignerSignRequest {
+    method: typeof METHOD_CONTEXT_SIGNER_SIGN;
+    /**
+     * An extra parameter, suggested to be used to help select which key to
+     * use. This lirbary passes it through verbatim; it's up to the extension
```
kostko, line 54:
```suggestion
     * use. This library passes it through verbatim; it's up to the extension
```

kostko (commented):
Cool, I like the general idea of how this works. For the API is there anything else that might be useful to have? Also should there be support for versioning the protocol?

client-sdk/ts-web/ext-utils/src/protocol.ts
pro-wh: o good idea

pro-wh: ok I tried adding one

pro-wh:
I tried adding tests but the headless version can't click the `window.confirm`s :laughing:

pro-wh:
could maybe do some versioning via method names. maybe we'd like to version the overall transport where "it makes an iframe and postmessage with type=MESSAGE_REQUEST etc" is version 1 as well

client-sdk/ts-web/ext-utils/src/protocol.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export interface MessageRequest {
+    type: typeof MESSAGE_TYPE_REQUEST;
+    id: number;
+    body: unknown;
+}
+
+// methods
+
+export const METHOD_KEYS_LIST = 'list-keys';
```
kostko, line 32:
```suggestion
export const METHOD_KEYS_LIST = 'keys-list';
```
Either that or rename all the types/vars.

client-sdk/ts-web/ext-utils/src/protocol.ts
kostko: Cool, looks good, thanks!

pro-wh:
I added some logic to make it not open any alerts and confirms under a special testing mode. Under this mode it also uses a fixed key from the ADR 0008 test vectors.

I added some `-v1` suffixes to message/method constants.

kostko (approved):
I see this is shared with #233.

pro-wh:
correction to cypress screenshot upload step. gonna merge

Issue #96, opened 2021-04-19
Benchmarks


kostko:
We should implement some end-to-end benchmarks for various modules that are part of the SDK and run them periodically (e.g. daily) on CI.

PR #223, opened 2021-06-15
Benchmarks

ptrus:
Closes: https://github.com/oasisprotocol/oasis-sdk/issues/96
Requires: https://github.com/oasisprotocol/oasis-core/pull/4025

Adds test benchmarking runtime, benchmarking client (mostly reused https://github.com/oasislabs/oasis-ethwasi-runtime/tree/master/benchmark), and initial account transfer benchmark.

TODO:
- [x] setup CI (dependabot, run on PRs)
- [x] setup daily tests on multiple instances (similar as old ethwasi benchmarks)
  - [x] need new dockerhub repo `oasisprotocol/oasis-sdk-benchmark`
  - [x] add dockerhub credentials to the added docker build CI job
- [x] ~add some more benchmarks~ (separate PRs, todo open issue to add benchmarks after this is merged)


tests/benchmark/runtime/client.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	tx := types.NewTransaction(nil, methodAccoutsMint, mint)
+	tx.AppendAuthSignature(signer.Public(), nonce)
+	stx := tx.PrepareForSigning()
+	if err = stx.AppendSign(info.ChainContext, signer); err != nil {
+		return err
+	}
+
+	if _, err = a.rc.SubmitTx(ctx, stx.UnverifiedTransaction()); err != nil {
```
kostko, line 40: Can you change these to use the transaction builder (just to be consistent with other modules)?

runtime-sdk/src/modules/accounts/mod.rs
```diff
@@ -91,6 +91,8 @@ pub struct Parameters {
     pub transfers_disabled: bool,
     #[serde(rename = "gas_costs")]
     pub gas_costs: GasCosts,
+    #[serde(rename = "debug_disable_nonce_check")]
+    pub debug_disable_nonce_check: bool,
```
kostko, line 95: Maybe configure serialization such that these debug params are not serialized if unset? Maybe also gate them under an "unsafe-allow-debug" feature that needs to be enabled during build time.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_ACCESS_TOKEN }}
+      - name: "Rebuild oasisprotocol/oasis-sdk-benchmark:${{ steps.determine_tag.outputs.tag }}"
+        uses: docker/build-push-action@v2.5.0
+        with:
+          context: .
+          file: tests/benchmark/Dockerfile
```
kostko, line 66: Could we make it so that the Oasis Core version comes from the const file that is also used for tests? To avoid having the version hardcoded in too many places.

ptrus: good idea, fixed

runtime-sdk/src/modules/accounts/mod.rs
ptrus: yeah good idea, done both

tests/benchmark/runtime/client.go
ptrus: changed

tests/benchmark/cmd/fixture.go
```diff
@@ -0,0 +1,167 @@
...
+			Values: []string{"100_000"},
+		},
+		{
+			Name:   p2p.CfgP2PValidateThrottle,
+			Values: []string{"100_000"},
+		},
+	}
+	// TODO: runtime.
+	clientExtraArgs := []oasis.Argument{
+		// We manually set the client runtime, as we only set the cupproted
```
kostko, line 62:
```suggestion
		// We manually set the client runtime, as we only set the supported
```

tests/benchmark/cmd/fixture.go
```diff
@@ -0,0 +1,167 @@
...
+		{
+			Name:   p2p.CfgP2PValidateConcurrency,
+			Values: []string{"100_000"},
+		},
+		{
+			Name:   p2p.CfgP2PValidateThrottle,
+			Values: []string{"100_000"},
+		},
+	}
+	// TODO: runtime.
```
kostko, line 60: What is missing here?

tests/benchmark/cmd/fixture.go
```diff
@@ -0,0 +1,167 @@
...
+				Executor: registry.ExecutorParameters{
+					GroupSize:       2,
+					GroupBackupSize: 1,
+					RoundTimeout:    5,
+					MaxMessages:     128,
+				},
+				TxnScheduler: registry.TxnSchedulerParameters{
+					Algorithm:         registry.TxnSchedulerSimple,
+					MaxBatchSize:      10_000,
+					MaxBatchSizeBytes: 10 * 16 * 1024 * 1024, // 160 MiB
```
kostko, line 123: Whoa, these are some large batches!

kostko (approved):

tests/benchmark/cmd/fixture.go
ptrus: was already done

tests/benchmark/cmd/fixture.go
ptrus: hm yeah i think i copied this from some other benchmarks, the batches aren't close this big though, the batch size is reached before.

pro-wh:
see #265 for test breakages

PR #241, opened 2021-06-23
runtime-sdk: bump oasis-core-runtime version

nhynes:
To prevent confusion about which version works on the testnet.

### DIFF
--- a/tests/consts.sh
+++ b/tests/consts.sh
@@ -6,7 +6,7 @@
 
 # Released version from GitHub Releases.
 # e.g. '21.1.2'
-OASIS_CORE_VERSION='21.2.3'
+OASIS_CORE_VERSION='21.2.4'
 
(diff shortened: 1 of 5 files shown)


kostko (commented):
Both versions should work as there have been no changes in the runtime interface. Feel free to also update the version in the README and the Go packages.

tests/consts.sh
```diff
@@ -6,7 +6,7 @@
 
 # Released version from GitHub Releases.
 # e.g. '21.1.2'
-OASIS_CORE_VERSION='21.2.3'
+OASIS_CORE_VERSION='21.2.4'
```
kostko, line 9: You should also update the `BUILD_NUMBER` below to correspond to the Buildkite build of 21.2.4.

kostko (approved):

PR #244, opened 2021-06-24
runtime-sdk: Add state migration helpers

### LINKED ISSUE #240: Convenience helpers for global state migrations
Besides per-module migrations we should also support for global runtime migrations that could be used to change runtime parameters. This should be part of the existing `migrate` method implementation and use the runtime metadata.

kostko:
Fixes #240.

### DIFF
--- a/runtime-sdk/src/modules/core/types.rs
+++ b/runtime-sdk/src/modules/core/types.rs
@@ -2,6 +2,9 @@ use std::collections::BTreeMap;
 
 use serde::{Deserialize, Serialize};
 
+/// Key in the versions map used for the global state version.
+pub const VERSION_GLOBAL_KEY: &str = "";
+
 #[derive(Clone, Debug, Default, Serialize, Deserialize)]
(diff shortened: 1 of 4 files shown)


pro-wh (approved):
modules had their own versions, right? now the runtime also has a version, which describes the runtime's "global state."

migrations can only happen one version at a time. this should be fine as long as the runtime developer and deployer are working together in lock step. it could be trouble otherwise.

it looks from the examples that we envision using this to update modules' parameters. I guess that makes sense, as the modules themselves are more like libraries that are unopinionated about their parameters.

kostko:
> migrations can only happen one version at a time. this should be fine as long as the runtime developer and deployer are working together in lock step. it could be trouble otherwise.

Yeah this is similar to how the genesis state is also part of the runtime itself. And the runtime is just a collection of modules, all the advanced logic should live in the modules anyway. So the runtime is just configuring a bunch of modules.

ptrus (approved):

PR #248, opened 2021-06-28
ts-web/ext-utils: add instructions

pro-wh:
Part of this is imported from some materials we shared privately. I've adjusted it to be more general. Then, I added instructions for the dapp side as well.


kostko (approved):
Nice! Would be useful if all these docs were consolidated under the top-level `docs` somehow.

pro-wh:
ah I see you filed #251

PR #283, opened 2021-07-14
runtime-sdk: Add EVM module

abukosek:
This PR implements a basic EVM module for the Runtime SDK.
Additional work has been filed under issue #351.


runtime-sdk/src/modules/evm/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+
+        if exit_reason.is_succeed() {
+            Ok(exit_value)
+        } else {
+            Err(Error::EVMError(format!("{:?}", exit_reason)))
+        }
+    }
+
+    fn tx_create<C: TxContext>(ctx: &mut C, body: types::CreateTx) -> Result<Vec<u8>, Error> {
+        Self::create(ctx, H160::from_slice(&body.caller), body.init_code)
```
kostko, line 152: Hm, so caller should be derived from the first transaction signer (similarly to how we define the caller), maybe by only working for simple Secp256k1 keys for now (returning an error otherwise), using the same derivation scheme as Ethereum.

runtime-sdk/Cargo.toml
```diff
@@ -25,6 +25,8 @@ impl-trait-for-tuples = "0.2.1"
 base64 = "0.13.0"
 once_cell = "1.8.0"
 slog = "2.7.0"
+evm = { version = "0.28.0", features = ["with-serde"] }
```
kostko, line 28: Given these dependencies, maybe we should move this module into its own crate (still in the same repo, just maybe under `runtime-sdk/modules/evm`) to avoid bloating the core SDK? I'll probably do the same with the `contracts` module.

runtime-sdk/modules/evm/src/evm_backend.rs
```diff
@@ -0,0 +1,224 @@
...
+pub const STORAGES: &[u8] = &[0xe5];
+
+impl<'c, C: oasis_runtime_sdk::Context> EVMBackend for Backend<'c, C> {
+    fn gas_price(&self) -> primitive_types::U256 {
+        self.vicinity.gas_price.into()
+    }
+    fn origin(&self) -> primitive_types::H160 {
+        self.vicinity.origin.into()
+    }
+    fn block_hash(&self, _number: primitive_types::U256) -> primitive_types::H256 {
```
synote, line 49: I suppose the runtime requires grpc to acccess to consensus state and get the block info

runtime-sdk/modules/evm/src/evm_backend.rs
```diff
@@ -0,0 +1,224 @@
...
+                    let mut accounts =
+                        storage::TypedStore::new(storage::PrefixStore::new(store, &ACCOUNTS));
+
+                    accounts.remove(&addr);
+                }
+            }
+        }
+
+        // TODO: What to do with logs, emit them as events?
+        for log in logs {
```
synote, line 220: The block tags may store log topics and data.

runtime-sdk/modules/evm/src/evm_backend.rs
kostko: It's not gRPC but rather there is the [Runtime Host Protocol](https://docs.oasis.dev/oasis-core/high-level-components/index-1/runtime-host-protocol) which provides all the necessary information to runtimes/ParaTimes. Note that the SDK already provides more high-level wrappers for accessing this information, so things like the block information is already available in the [`Context`](https://api.docs.oasis.dev/oasis-sdk/oasis_runtime_sdk/context/trait.Context.html#tymethod.runtime_header) that is passed to all transaction handlers etc. The [`Header`](https://github.com/oasisprotocol/oasis-core/blob/master/runtime/src/consensus/roothash.rs#L143) is of course slightly different from Ethereum's so we should figure out how to properly translate this over.

runtime-sdk/modules/evm/src/evm_backend.rs
```diff
@@ -0,0 +1,224 @@
...
+        self.vicinity.gas_price.into()
+    }
+    fn origin(&self) -> primitive_types::H160 {
+        self.vicinity.origin.into()
+    }
+    fn block_hash(&self, _number: primitive_types::U256) -> primitive_types::H256 {
+        primitive_types::H256::default()
+    }
+    fn block_number(&self) -> primitive_types::U256 {
+        primitive_types::U256::default()
```
kostko, line 53: We could just use the round number here (e.g., from `self.ctx.runtime_header().round`).

runtime-sdk/modules/evm/src/evm_backend.rs
```diff
@@ -0,0 +1,224 @@
...
+    }
+    fn block_coinbase(&self) -> primitive_types::H160 {
+        primitive_types::H160::default()
+    }
+    fn block_timestamp(&self) -> primitive_types::U256 {
+        primitive_types::U256::zero()
```
kostko, line 59: We could just use the runtime block timestamp here (e.g., from `self.ctx.runtime_header().timestamp`).

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -0,0 +1,361 @@
...
+                Ok(cbor::to_value(Self::q_peek_storage(ctx, args)?))
+            })()),
+            "evm.PeekCode" => module::DispatchResult::Handled((|| {
+                let args = cbor::from_value(args).map_err(|_| Error::InvalidArgument)?;
+                Ok(cbor::to_value(Self::q_peek_code(ctx, args)?))
+            })()),
+            _ => module::DispatchResult::Unhandled(args),
+        }
+    }
+}
```
nhynes, line 334: sourcecode boil, sourcecode bubble, something plate-y this way comes. I wonder if #139 will ever land?

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -0,0 +1,361 @@
...
+    ) -> bool {
+        let version = meta.versions.get(Self::NAME).copied().unwrap_or_default();
+        if version == 0 {
+            // Initialize state from genesis.
+            Self::set_params(ctx.runtime_state(), genesis.parameters);
+            meta.versions.insert(Self::NAME.to_owned(), Self::VERSION);
+            return true;
+        }
+
+        // Migrations are not supported.
```
nhynes, line 352:
```suggestion
        // Migrations are not used.
```

runtime-sdk/modules/evm/src/types.rs
```diff
@@ -0,0 +1,168 @@
+//! EVM module types.
+#[derive(Clone, Debug, cbor::Encode, cbor::Decode)]
+pub struct CreateTx {
+    pub value: Vec<u8>, // U256
```
nhynes, line 4: can you not make it a `U256` directly? it's just a `[u8; N]`. It'll save you a few copies (or unsafe code) if it's direct.

runtime-sdk/modules/evm/src/types.rs
```diff
@@ -0,0 +1,168 @@
...
+
+impl H160 {
+    pub fn from_slice(src: &[u8]) -> Self {
+        H160(primitive_types::H160::from_slice(src))
+    }
+
+    pub fn as_bytes(&self) -> &[u8] {
+        self.0.as_bytes()
+    }
+}
```
nhynes, line 168: this is a tonne of code that you don't need. Try this instead:

```rust
// This `mod` exists solely to place an `#[allow(warnings)]` around the generated code.
#[allow(warnings)]
mod eth {
    use super::*;

    macro_rules! construct_fixed_hash {
        ($name:ident($num_bytes:literal)) => {
            fixed_hash::construct_fixed_hash! {
                #[derive(cbor::Encode, cbor::Decode)
               #[cbor(transparent)
                pub struct $name($num_bytes);
            }
        };
    }

    macro_rules! construct_uint {
        ($name:ident($num_words:tt)) => {
            uint::construct_uint! {
                #[derive(cbor::Encode, cbor::Decode)]
                #[cbor(transparent)]
                pub struct $name($num_words);
            }
        };
    }

    construct_fixed_hash!(H160(20));
    construct_fixed_hash!(H256(32));
    construct_uint!(U256(4));

    macro_rules! impl_upstream_conversions {
        ($($ty:ident => $upstream:ident),* $(,)?) => {
            $(
                impl From<$ty> for primitive_types::$upstream {
                    fn from(t: $ty) -> Self {
                        Self(t.0)
                    }
                }

                impl From<primitive_types::$upstream> for $ty {
                    fn from(t: primitive_types::$upstream) -> Self {
                        Self(t.0)
                    }
                }
            )*
        }
    }

    impl_upstream_conversions! {
        H160, H256, U256
    }
}
pub use eth::{H160, H256, U256};
```
and then you can remove all of the converters, and also use strongly-typed values in the RPC structs

tests/runtimes/simple-evm/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+name = "test-runtime-simple-evm"
+version = "0.1.0"
+authors = ["Oasis Protocol Foundation <info@oasisprotocol.org>"]
+edition = "2018"
+license = "Apache-2.0"
+
+[dependencies]
+oasis-runtime-sdk = { path = "../../../runtime-sdk" }
+oasis-runtime-sdk-evm = { path = "../../../runtime-sdk/modules/evm" }
+cbor = { version = "0.1.3", package = "oasis-cbor" }
```
nhynes, line 11: sort

runtime-sdk/modules/evm/src/evm_backend.rs
```diff
@@ -0,0 +1,224 @@
...
+    pub fn new(vicinity: Vicinity, ctx: &'c mut C) -> Self {
+        Self {
+            vicinity,
+            ctx: RefCell::new(ctx),
+        }
+    }
+}
+
+/// Prefix for Ethereum accounts in our storage (maps H160 -> Account).
+pub const ACCOUNTS: &[u8] = &[0xea];
```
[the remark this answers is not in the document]
kostko: Yeah, any reason this doesn't start with `0x01`?

runtime-sdk/modules/evm/src/evm_backend.rs
```diff
@@ -0,0 +1,224 @@
...
+        primitive_types::U256::zero()
+    }
+    fn block_difficulty(&self) -> primitive_types::U256 {
+        primitive_types::U256::zero()
+    }
+    fn block_gas_limit(&self) -> primitive_types::U256 {
+        primitive_types::U256::zero()
+    }
+    fn chain_id(&self) -> primitive_types::U256 {
+        primitive_types::U256::default()
```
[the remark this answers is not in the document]
kostko: Yeah I already suggested above that we use our chain context here (which happens to also be 256-bit).

Oh it seems I didn't submit my review :P

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -0,0 +1,361 @@
...
+                // Caller address is derived by doing Keccak-256 on the
+                // secp256k1 public key and taking the last 20 bytes
+                // of the result.
+                let mut k = Keccak::v256();
+                let mut out = [0u8; 32];
+                k.update(pk.as_bytes());
+                k.finalize(&mut out);
+                Ok(H160::from_slice(&out[32 - 20..]))
+            }
+            _ => Err(Error::InvalidSignerType),
```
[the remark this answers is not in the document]
kostko: Not sure how that would work with Ethereum addresses unless we introduce a custom scheme for other signers?

runtime-sdk/modules/evm/src/lib.rs
kostko: Would be great if it did ;-)

runtime-sdk/modules/evm/src/evm_backend.rs
```diff
@@ -0,0 +1,224 @@
...
+
+#[derive(Clone, Eq, PartialEq, cbor::Encode, cbor::Decode, Default)]
+pub struct Account {
+    pub nonce: U256,
+    pub balance: U256,
+}
+
+pub struct Backend<'c, C: oasis_runtime_sdk::Context> {
+    vicinity: Vicinity,
+    ctx: RefCell<&'c mut C>,
```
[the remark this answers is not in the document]
kostko: Definitely first within runtimes.

runtime-sdk/modules/evm/src/evm_backend.rs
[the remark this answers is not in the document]
kostko: You should use `crypto::signature::get_chain_context_for with` a base like `oasis-runtime-sdk/evm: chain id` to derive the chain id to use here.

runtime-sdk/modules/evm/src/evm_backend.rs
```diff
@@ -0,0 +1,224 @@
...
+
+impl<'c, C: oasis_runtime_sdk::Context> EVMBackend for Backend<'c, C> {
+    fn gas_price(&self) -> primitive_types::U256 {
+        self.vicinity.gas_price.into()
+    }
+    fn origin(&self) -> primitive_types::H160 {
+        self.vicinity.origin.into()
+    }
+    fn block_hash(&self, _number: primitive_types::U256) -> primitive_types::H256 {
+        primitive_types::H256::default()
```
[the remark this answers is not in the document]
anonymousGiga: use self.ctx.runtime_header().encoded_hash() can only get the current block's hash,  but the hash of the block of a given block number cannot be obtained.

runtime-sdk/modules/evm/src/evm_backend.rs
```diff
@@ -0,0 +1,224 @@
+//! EVM backend.
+use std::cell::RefCell;
+
+use oasis_runtime_sdk::storage::{self, Store as _};
+
+use crate::types::{H160, H256, U256};
+use evm::backend::{Apply, ApplyBackend, Backend as EVMBackend, Basic, Log};
+
+#[derive(Clone, Eq, PartialEq, cbor::Encode, cbor::Decode, Default)]
+pub struct Vicinity {
```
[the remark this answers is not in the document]
abukosek: `Vicinity` sounds dumb to me as well, but that is what the `evm` crate uses throughout its code for this struct (see e.g. https://github.com/rust-blockchain/evm/blob/master/src/backend/memory.rs), so I didn't want to invent a new word for it.

runtime-sdk/modules/evm/src/evm_backend.rs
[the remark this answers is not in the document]
abukosek: It doesn't seem like we have a guide for choosing prefixes, so I chose `0xea` as shorthand for **E**thereum **A**ccounts.

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -0,0 +1,361 @@
...
+const MODULE_NAME: &str = "evm";
+
+pub struct Module;
+
+/// Errors emitted by the EVM module.
+#[derive(Error, Debug, oasis_runtime_sdk_macros::Error)]
+pub enum Error {
+    #[error("invalid argument")]
+    #[sdk_error(code = 1)]
+    InvalidArgument,
```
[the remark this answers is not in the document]
abukosek: We should change this in the rest of the runtime-sdk as well.

runtime-sdk/modules/evm/src/types.rs
abukosek: Nice! Thanks!

runtime-sdk/modules/evm/src/evm_backend.rs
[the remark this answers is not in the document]
kostko: That's true, is there a specific requirement why this information on past blocks would be useful? It definitely doesn't seem wise to allow arbitrary historic horizons as this imposes additional requirements on nodes. I would just return the current block hash here for now.

runtime-sdk/modules/evm/src/evm_backend.rs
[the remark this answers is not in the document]
kostko: Yeah in all other modules we just start with `0x01` :-) Also the constants are usually placed inside a `mod state` inside the top-level module, let's do the same here so they are in one place and it is easy to see which are in use.

tests/e2e/simpleevmtest.go
```diff
@@ -0,0 +1,463 @@
...
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/crypto/signature"
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/testing"
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/types"
+
+	"github.com/oasisprotocol/oasis-sdk/tests/e2e/txgen"
+)
+
+// The evmCreateTx type must match the CreateTx type from the evm module types
+// in runtime-sdk/src/modules/evm/types.rs.
+type evmCreateTx struct {
```
kostko, line 24: Could you move these to the client-sdk, e.g. to `client-sdk/go/modules/evm/types.go`.

tests/e2e/simpleevmtest.go
```diff
@@ -0,0 +1,463 @@
...
+	Index   []byte `json:"index"`
+}
+
+// The evmPeekCodeQuery type must match the PeekCodeQuery type from the
+// evm module types in runtime-sdk/src/modules/evm/types.rs.
+type evmPeekCodeQuery struct {
+	Address []byte `json:"address"`
+}
+
+func evmCreate(ctx context.Context, rtc client.RuntimeClient, signer signature.Signer, tx evmCreateTx) ([]byte, error) {
```
kostko, line 54: Also these functions would be useful to have in the client-sdk, see how it's done in the contracts module PR.

tests/e2e/simpleevmtest.go
```diff
@@ -0,0 +1,463 @@
...
...02fc565b6001600160a01b038085166000908152602081905260408082208585039055918516815290812080548492906106a790849061082d565b92505081905550826001600160a01b0316846001600160a01b03167fddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef846040516106f391815260200190565b60405180910390a350505050565b80356001600160a01b038116811461071857600080fd5b919050565b60006020828403121561072f57600080fd5b61073882610701565b9392505050565b6000806040838503121561075257600080fd5b61075b83610701565b915061076960208401610701565b90509250929050565b60008060006060848603121561078757600080fd5b61079084610701565b925061079e60208501610701565b9150604084013590509250925092565b600080604083850312156107c157600080fd5b6107ca83610701565b946020939093013593505050565b600060208083528351808285015260005b81811015610805578581018301518582016040015282016107e9565b81811115610817576000604083870101525b50601f01601f1916929092016040019392505050565b6000821982111561084e57634e487b7160e01b600052601160045260246000fd5b500190565b600181811c9082168061086757607f821691505b6020821081141561088857634e487b7160e01b600052602260045260246000fd5b5091905056fea264697066735822122057fae6e23c9b696979cb61373ad6bb8f5e6f3dd858b98a3b12e629cd6536fa5764736f6c63430008060033")
```
kostko, line 359: Would be better to move this into a separate file and use [`go:embed`](https://pkg.go.dev/embed).

runtime-sdk/modules/evm/src/evm_backend.rs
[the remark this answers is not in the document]
anonymousGiga: Yes, it is described in this document（https://docs.soliditylang.org/en/latest/yul.html?highlight=blockhash#evm-dialect）, this instruction needed from Frontier.

runtime-sdk/modules/evm/src/evm_backend.rs
```diff
@@ -0,0 +1,253 @@
...
+        self.ctx.borrow().runtime_header().timestamp.into()
+    }
+    fn block_difficulty(&self) -> primitive_types::U256 {
+        primitive_types::U256::zero()
+    }
+    fn block_gas_limit(&self) -> primitive_types::U256 {
+        primitive_types::U256::zero()
+    }
+    fn chain_id(&self) -> primitive_types::U256 {
+        crypto::signature::context::get_chain_context_for(b"oasis-runtime-sdk/evm: chain id")[..32]
```
kostko, line 83: Move this domain separation context into a constant.

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -0,0 +1,382 @@
...
+            return Err(Error::GasPriceTooLow);
+        }
+
+        let vicinity = evm_backend::Vicinity {
+            gas_price: gas_price_in,
+            origin: source,
+        };
+
+        let mut backend = evm_backend::Backend::<'_, C>::new(vicinity, ctx);
+        let metadata = StackSubstateMetadata::new(u64::max_value(), &Self::EVM_CONFIG);
```
kostko, line 226: Why do you set the gas limit to `u64::max_value()` here? Shouldn't it be set to `gas_limit`? Because currently you only do the check after executing the transaction which opens up a DoS vector where people can execute stuff for which they have not paid.

abukosek: Fixed, thanks :)

tests/e2e/evm_erc20_test_compiled.hex
```diff
@@ -0,0 +1 @@
...6102fc565b6001600160a01b038085166000908152602081905260408082208585039055918516815290812080548492906106a790849061082d565b92505081905550826001600160a01b0316846001600160a01b03167fddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef846040516106f391815260200190565b60405180910390a350505050565b80356001600160a01b038116811461071857600080fd5b919050565b60006020828403121561072f57600080fd5b61073882610701565b9392505050565b6000806040838503121561075257600080fd5b61075b83610701565b915061076960208401610701565b90509250929050565b60008060006060848603121561078757600080fd5b61079084610701565b925061079e60208501610701565b9150604084013590509250925092565b600080604083850312156107c157600080fd5b6107ca83610701565b946020939093013593505050565b600060208083528351808285015260005b81811015610805578581018301518582016040015282016107e9565b81811115610817576000604083870101525b50601f01601f1916929092016040019392505050565b6000821982111561084e57634e487b7160e01b600052601160045260246000fd5b500190565b600181811c9082168061086757607f821691505b6020821081141561088857634e487b7160e01b600052602260045260246000fd5b5091905056fea264697066735822122057fae6e23c9b696979cb61373ad6bb8f5e6f3dd858b98a3b12e629cd6536fa5764736f6c63430008060033
```
kostko, line 1: In the WASM contracts PR I've created a `tests/e2e/contracts` directory for these to avoid cluttering the E2E dir.

client-sdk/go/modules/evm/evm.go
```diff
@@ -0,0 +1,76 @@
...
+	methodPeekStorage = "evm.PeekStorage"
+	methodPeekCode    = "evm.PeekCode"
+)
+
+// V1 is the v1 EVM module interface.
+type V1 interface {
+	Create(value []byte, initCode []byte, gasPrice []byte, gasLimit uint64) *client.TransactionBuilder
+	Call(address []byte, value []byte, data []byte, gasPrice []byte, gasLimit uint64) *client.TransactionBuilder
+	PeekStorage(ctx context.Context, address []byte, index []byte) ([]byte, error)
+	PeekCode(ctx context.Context, address []byte) ([]byte, error)
```
kostko, line 24: Add documentation comments similar to other modules.

tests/e2e/evm_erc20_test_compiled.hex
abukosek: Good idea, I've moved these contracts to that directory too.

client-sdk/go/modules/evm/evm.go
abukosek: Done :)

runtime-sdk/modules/evm/src/evm_backend.rs
```diff
@@ -0,0 +1,254 @@
...
+    storage::{self, Store as _},
+};
+
+use crate::{
+    state,
+    types::{H160, H256, U256},
+};
+use evm::backend::{Apply, ApplyBackend, Backend as EVMBackend, Basic, Log};
+
+const EVM_CHAIN_CONTEXT: &[u8] = b"oasis-runtime-sdk/evm: chain id";
```
kostko, line 15: Add documentation comment.

runtime-sdk/modules/evm/src/evm_backend.rs
```diff
@@ -0,0 +1,254 @@
...
+use oasis_runtime_sdk::{
+    crypto,
+    storage::{self, Store as _},
+};
+
+use crate::{
+    state,
+    types::{H160, H256, U256},
+};
+use evm::backend::{Apply, ApplyBackend, Backend as EVMBackend, Basic, Log};
```
kostko, line 13: Move this into its own import group above the `oasis_runtime_sdk` group.

kostko (approved):
Ok, let's merge this and we can do follow up improvements (e.g., #351, #376) in a separate PR.

runtime-sdk/modules/evm/src/evm_backend.rs
```diff
@@ -0,0 +1,224 @@
...
+                            let mut codes = storage::TypedStore::new(storage::PrefixStore::new(
+                                c_store, &CODES,
+                            ));
+                            codes.insert(&addr, code);
+                        }
+
+                        if reset_storage {
+                            let state = self.ctx.get_mut().runtime_state();
+                            let s_store = storage::PrefixStore::new(state, &crate::MODULE_NAME);
+                            let mut storages = storage::PrefixStore::new(s_store, &STORAGES);
```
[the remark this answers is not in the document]
abukosek: This was surprisingly too much of a pain to get working in this PR on time, but I'll get it done in the next one :)

PR #285, opened 2021-07-15
 ts-web/ext-utils: add keys change event 

### LINKED ISSUE #284: ts-web/ext-utils: notify dapp when available keys changes
requested by an extension wallet developer. let's add a message type and broadcast it to `'*'` origin. no data with the event, have the dapp call `keys.list()` again if it cares

pro-wh:
Here's `oasisExt.keys.setKeysChangeHandler(conn, handler)`. When the extension calls `oasisExt.ext.keysChanged()`, ext-utils calls `handler` for the dapp.

Some notes, decide if you're okay with these:

- The list of keys is not included in the event. Call keys.list to get the new list.
- The dapp can have at most one handler for the keys change event per connection.
- There's only one kind of keys change event. The dapp may be able to diff against previous keys lists.

fixes #284


kostko (approved):

pro-wh:
I'm going to merge this, given that we've taken the time to review it. I'll split off the additional sample ext and sample page changes into a new PR.

pro-wh:
additional changes moved/added to #301

PR #300, opened 2021-07-20
ci: Add workflow for runtime build env Docker images

jberci:


.github/workflows/runtime-builder-image.yml
```diff
@@ -0,0 +1,66 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          # Check out pull request's HEAD commit instead of the merge commit.
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Determine tag name
+        id: determine-tag
+        uses: ./.github/actions/determine-tag
```
ptrus, line 46: Can you also update https://github.com/oasisprotocol/oasis-sdk/blob/main/.github/workflows/docker.yml to use this internal action for determining tag?

---

Also maybe rename the existing `docker` workflow to `benchmarking-image` (or something similar) now that we have multiple "docker" workflows

docker/runtime-builder/Dockerfile
```diff
@@ -0,0 +1,41 @@
+FROM ubuntu:20.04
+
+# Package versions.
+ARG GO_VERSION=1.16.4
+ARG RUST_NIGHTLY_VERSION=2021-05-07
+ARG OASIS_CORE_VERSION=21.2.7
```
ptrus, line 6: Can you remove default value for `OASIS_CORE_VERSION` and require passing it in as build argument.

In the github action you can then set it as we do it in the benchmarks workflow: https://github.com/oasisprotocol/oasis-sdk/blob/e7f68932321cff1ff0304839630784e7b6f1fccc/.github/workflows/docker.yml#L76-L77

(maybe also extract `determine_oasis_core_version` as you did for determining tag).

So that there's not one more file where the version needs to be updated.

---

Same thing can be done for `RUST_NIGHTLY_VERSION` (can get it automatically by reading https://github.com/oasisprotocol/oasis-sdk/blob/main/rust-toolchain).

docker/runtime-builder/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+    git gcc curl jq pkg-config libssl-dev protobuf-compiler && \
+    apt-get autoclean && apt-get autoremove && rm -rf /var/cache/apt/archives/* && \
+    mkdir -p /src /cargo/.rustup
+
+ENV HOME="/root"
+ENV GOPATH="/go"
+ENV CARGO_HOME="/cargo"
+ENV RUSTUP_HOME="/cargo/.rustup"
+ENV PATH="${CARGO_HOME}/bin:${GOPATH}/bin:/usr/local/go/bin:${PATH}"
+ENV SDK_SRC="/src"
```
ptrus, line 20: This seems unused within the build script, so unless the user of the image will mount the sdk folder at `/src`, this will be wrong. Maybe just remove the environment variable here.

docker/runtime-builder/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+
+# Install Rust.
+RUN curl "https://sh.rustup.rs" -sfo rustup.sh && \
+    sh rustup.sh -y --default-toolchain nightly-${RUST_NIGHTLY_VERSION} && \
+    rustup target add x86_64-fortanix-unknown-sgx && \
+    cargo install --version 0.4.0 fortanix-sgx-tools && \
```
ptrus, line 26: extract `0.4.0` (and `0.8.2` bellow) to build args

.github/actions/oasis-core-version/action.yml
```diff
@@ -0,0 +1,14 @@
+name: Determine Oasis Core Version
```
ptrus, line 1: Minor (feel free to ignore): Should this be renamed to "Determine Oasis Core Commit Reference" or something in those lines? (also in other places)

ptrus (approved):
Looks good, thanks!

.github/actions/oasis-core-version/action.yml
ptrus: Thanks, looks good now!

ptrus (approved):

kostko (approved):

PR #306, opened 2021-07-21
Add support for sr25519 signatures

Yawning:
* [x] runtime-sdk
 * [x] client-sdk (Go)

Someone that cares can deal with typescript, because I do not.


runtime-sdk/Cargo.toml
```diff
@@ -15,6 +15,7 @@ serde_bytes = "0.11.5"
 curve25519-dalek = "3.0.0"
 sha2 = "0.9.3"
 k256 = { version = "0.9.4" }
+schnorrkel = { version = "0.10.1", features = ["serde"] }
```
kostko, line 18: I believe the serde feature is not needed now as you switched to `Vec<u8>`?

kostko (commented):
Would be good to add some (E2E) tests for this.

runtime-sdk/src/crypto/signature/sr25519.rs
```diff
@@ -0,0 +1,81 @@
...
+            .verify(transcript, &signature)
+            .map_err(|_| Error::VerificationFailed)
+    }
+
+    fn decompress_public_key(bytes: &[u8]) -> Result<schnorrkel::PublicKey, Error> {
+        schnorrkel::PublicKey::from_bytes(bytes).map_err(|_| Error::MalformedPublicKey)
+    }
+}
+
+impl cbor::Encode for PublicKey {
```
kostko, line 66: These are actually not needed with the new cbor library. You can use `derive(..., cbor::Encode, cbor::Decode)` and the `cbor(transparent)` attribute on the struct to get the same result.

pro-wh (approved):

kostko (approved):

Issue #278, opened 2021-07-09
Add support for prefetching from storage


ptrus:
Add support for prefetching and for modules to specify what to prefetch before starting to process a batch. This can avoid many serial round trips form storage when processing large batches (especially while https://github.com/oasisprotocol/oasis-core/issues/3972 is not done).

PR #320, opened 2021-07-26
runtime-sdk: add prefetching support

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-sdk/issues/278

This is ready for some reviews, before i finish by implementing prefetch in other modules.

TODO:
- [x] add missing prefetch implementations to modules
- [x] tests

### DIFF
--- a/runtime-sdk/src/dispatcher.rs
+++ b/runtime-sdk/src/dispatcher.rs
@@ -1,6 +1,6 @@
 //! Transaction dispatcher.
 use std::{
-    collections::BTreeMap,
+    collections::{BTreeMap, BTreeSet},
     marker::PhantomData,
     sync::{atomic::AtomicBool, Arc},
 };
@@ -28,8 +28,10 @@ use crate::{
     modules,
     modules::core::API as _,
     runtime::Runtime,
-    storage, types,
-    types::transaction::TransactionWeight,
+    storage,
+    storage::Prefix,
+    types,
+    types::transaction::{Transaction, TransactionWeight},
 };
 
 /// Unique module name.
@@ -46,6 +48,10 @@ pub enum Error {
     #[error("malformed transaction in batch: {0}")]
     #[sdk_error(code = 2)]
     MalformedTransactionInBatch(#[source] modules::core::Error),
+
+    #[error("prefetch failed: {0}")]
+    #[sdk_error(code = 3)]
+    PrefetchFailed(#[source] RuntimeError),
 }
 
 /// Result of dispatching a transaction.
@@ -169,21 +175,7 @@ impl<R: Runtime> Dispatcher<R> {
     }
 
     /// Check whether the given transaction is valid.
-    pub fn check_tx<C: BatchContext>(ctx: &mut C, tx: &[u8]) -> Result<CheckTxResult, Error> {
-        let tx = match Self::decode_tx(ctx, &tx) {
-            Ok(tx) => tx,
-            Err(err) => {
-                return Ok(CheckTxResult {
-                    error: RuntimeError {
-                        module: err.module_name().to_string(),
-                        code: err.code(),
-                        message: err.to_string(),
-                    },
-                    meta: None,
-                })
-            }
-        };
-
+    pub fn check_tx<C: BatchContext>(ctx: &mut C, tx: Transaction) -> Result<CheckTxResult, Error> {
         let dispatch = Self::dispatch_tx(ctx, tx)?;
         match dispatch.result {
             types::transaction::CallResult::Ok(_value) => Ok(CheckTxResult {
@@ -210,20 +202,27 @@ impl<R: Runtime> Dispatcher<R> {
     }
 
     /// Execute the given transaction.
-    pub fn execute_tx<C: BatchContext>(ctx: &mut C, tx: &[u8]) -> Result<ExecuteTxResult, Error> {
-        // It is an error to include a malformed transaction in a batch. So instead of only
-        // reporting a failed execution result, we fail the whole batch. This will make the compute
-        // node vote for failure and the round will fail.
-        //
-        // Correct proposers should only include transactions which have passed check_tx.
-        let tx = Self::decode_tx(ctx, &tx).map_err(Error::MalformedTransactionInBatch)?;
-
+    pub fn execute_tx<C: BatchContext>(
+        ctx: &mut C,
+        tx: Transaction,
+    ) -> Result<ExecuteTxResult, Error> {
         let dispatch_result = Self::dispatch_tx(ctx, tx)?;
 
         Ok(ExecuteTxResult {
             output: cbor::to_vec(dispatch_result.result),
             tags: dispatch_result.tags,
         })
+    }
+
+    /// Prefetch prefixes for the given transaction.
+    pub fn prefetch_tx(
+        prefixes: &mut BTreeSet<Prefix>,
+        tx: types::transaction::Transaction,
+    ) -> Result<(), RuntimeError> {
+        match R::Modules::prefetch(prefixes, &tx.call.method, tx.call.body, &tx.auth_info) {
+            module::DispatchResult::Handled(r) => r,
+            module::DispatchResult::Unhandled(_) => Ok(()), // Unimplemented prefetch is allowed.
+        }
     }
 
     fn handle_last_round_messages<C: Context>(ctx: &mut C) -> Result<(), modules::core::Error> {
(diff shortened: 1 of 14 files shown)


runtime-sdk/src/modules/accounts/mod.rs
```diff
@@ -422,6 +426,41 @@ impl module::Module for Module {
...
+    fn prefetch(
+        prefixes: &mut BTreeSet<Prefix>,
+        method: &str,
+        body: cbor::Value,
+        auth_info: &AuthInfo,
+    ) -> module::DispatchResult<cbor::Value, Result<(), error::RuntimeError>> {
+        match method {
+            "accounts.Transfer" => {
+                module::DispatchResult::Handled(|| -> Result<(), error::RuntimeError> {
+                    let args: types::Transfer =
```
kostko, line 438: One annoying thing seems to be that we will now be deserializing the body twice, once during `prefetch` and again in `dispatch_call`. Not sure how hard it would be to avoid double deserialize, maybe as a future improvement :-)

runtime-sdk/src/storage/mkvs.rs
```diff
@@ -21,6 +21,13 @@ impl<M: mkvs::MKVS> MKVSStore<M> {
     fn create_ctx(&self) -> Context {
         Context::create_child(&self.ctx)
     }
+
+    // TODO: fix the type in oasis-core.
```
kostko, line 25: What is this about? Let's file an issue in Oasis Core and link it here.

runtime-sdk/src/dispatcher.rs
```diff
@@ -28,13 +28,17 @@ use crate::{
...
+    storage,
+    storage::Prefix,
+    types,
+    types::transaction::{Transaction, TransactionWeight},
 };
 
 /// Unique module name.
 const MODULE_NAME: &str = "dispatcher";
 
+const PREFETCH_LIMIT: u16 = std::u16::MAX;
```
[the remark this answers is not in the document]
ptrus: Hm maybe. So I was thinking about making prefetch itself configurable. Not sure if it makes more sense for this to be configurable by the runtime (via the runtime trait), or by the nodes running runtimes (via local config)?

Can then do the same for this limit.

runtime-sdk/src/storage/mkvs.rs
ptrus: Oh it's clippy complaining about to replacing the argument `&Vec` with a slice. I actually fixed it for now by not using a reference for `prefetch_prefixes` in oasis-sdk.

Can open an issue in oasis-core for this, but actually i think that since clippy is not setup in oasis-core, there might be bunch of lint fixes needed - should just setup clippy in oasis-core and address all of the issues found, so i'll open issue for that. Probably should be done only after cleaning up/removing all of the deprecated stuff.

kostko: Yeah let's open an issue to setup clippy in Core.

ptrus: https://github.com/oasisprotocol/oasis-core/issues/4170

runtime-sdk/src/dispatcher.rs
```diff
@@ -28,13 +28,20 @@ use crate::{
...
+
+/// Local config key for disabling prefetch.
+const CFG_PREFETCH_DISABLE: &str = "prefetch_disable";
```
ptrus, line 43: Does this make sense, or would it made more sense for this to be configurable by the runtime?

kostko: The only problem can be if the value of this could affect security and/or determinism in any way. On the other hand making it only configurable by the runtime would make it impossible for node operators to turn this off and would ensure that all runtime instances have it configured the same.

What is the reason one would want to disable prefetching?

ptrus: Yeah actually maybe indeed there's no reason for disabling prefetching. It was mostly useful for me to comapre benchmarks with and without it - but that can also be done by making it configurable by the runtime. It affects the account transfer benchmarks for about 5% (prefetch worsens it).

ptrus: I made `PREFETCH_LIMIT` a Runtime trait constant, and if set to `0` it will disable prefetch. I think this makes most sense (makes the limit configurable by runtimes and also can be used to disable prefetch).

runtime-sdk/src/modules/accounts/mod.rs
ptrus: Hm yeah, i'll open an issue for this once merged.

runtime-sdk/src/dispatcher.rs
```diff
@@ -365,13 +392,45 @@ impl<R: Runtime> transaction::dispatcher::Dispatcher for Dispatcher<R> {
...
+            for tx in batch.iter() {
+                let res = match Self::decode_tx(&mut ctx, tx) {
+                    Ok(tx) => {
+                        if prefetch_enabled {
+                            Self::prefetch_tx(&mut prefixes, tx.clone()).map(|_| tx)
+                        } else {
+                            Ok(tx)
+                        }
+                    }
+                    Err(err) => Err(RuntimeError {
```
kostko, line 411: The error derive macro now provides a conversion to `RuntimeError` so you should be able to just do `err.into()` here.

runtime-sdk/src/modules/consensus_accounts/test.rs
```diff
@@ -389,3 +391,71 @@ fn test_consensus_withdraw_handler() {
...
+            &mut prefixes,
+            &call.method,
+            call.body,
+            &auth_info,
+        )
+        .ok_or(anyhow!("dispatch failure"))
+        .expect("prefetch should succeed");
+
+        assert!(matches!(result, Ok(())));
+        assert_eq!(prefixes.len(), 0, "there should be 1 prefix to be fetched");
```
kostko, line 459:
```suggestion
        assert_eq!(prefixes.len(), 0, "there should be 0 prefixes to be fetched");
```

runtime-sdk/src/storage/mkvs.rs
```diff
@@ -21,6 +21,11 @@ impl<M: mkvs::MKVS> MKVSStore<M> {
     fn create_ctx(&self) -> Context {
         Context::create_child(&self.ctx)
     }
+
+    pub fn prefetch_prefixes(&mut self, prefixes: Vec<mkvs::Prefix>, limit: u16) {
```
kostko, line 25: Add documentation comment.

runtime-sdk/src/runtime.rs
```diff
@@ -19,6 +19,9 @@ pub trait Runtime {
     /// State version.
     const STATE_VERSION: u32 = 0;
 
+    /// Prefetch limit. To disable prefetch set it to 0.
+    const PREFETCH_LIMIT: u16 = std::u16::MAX;
```
kostko, line 23: I remember reading in one of your comments that enabling prefetch worsens benchmark results in some cases? Does it still make sense to enable it by default?

tests/benchmark/cmd/fixture.go
```diff
@@ -67,6 +64,10 @@ func fixture() *oasis.NetworkFixture {
 	}
 	clientExtraArgs = append(clientExtraArgs, computeExtraArgs...)
 
+	runtimeConfig := make(map[int]map[string]interface{})
+	runtimeConfig[0] = make(map[string]interface{})
+	runtimeConfig[0]["prefetch_disable"] = 1
```
kostko, line 69: This does not seem to be used anymore?

runtime-sdk/src/runtime.rs
ptrus: By about 5% on the accounts-transfer benchmark yes. Agreed it makes more sense to disable it by default for now.

runtime-sdk/src/dispatcher.rs
ptrus: fixed

runtime-sdk/src/modules/consensus_accounts/test.rs
ptrus: fixed

runtime-sdk/src/storage/mkvs.rs
ptrus: done

runtime-sdk/src/modules/consensus_accounts/test.rs
```diff
@@ -389,3 +391,75 @@ fn test_consensus_withdraw_handler() {
...
+        version: 1,
+        call: transaction::Call {
+            method: "consensus.Deposit".to_owned(),
+            body: cbor::to_value(Deposit {
+                amount: BaseUnits::new(1_000, Denomination::NATIVE),
+            }),
+        },
+        auth_info: auth_info.clone(),
+    };
+    // Deposit should result in zero prefixes..
```
kostko, line 446:
```suggestion
    // Deposit should result in zero prefixes.
```

runtime-sdk/src/runtime.rs
```diff
@@ -19,6 +19,9 @@ pub trait Runtime {
     /// State version.
     const STATE_VERSION: u32 = 0;
 
+    /// Prefetch limit. To enable prefetch set it to non zero value.
```
kostko, line 22:
```suggestion
    /// Prefetch limit. To enable prefetch set it to a non-zero value.
```

kostko (approved):

PR #331, opened 2021-07-30
ci: Add check for build reproducibility

### LINKED ISSUE #136: Deterministic build process
We should have a documented way to build runtimes based on the SDK deterministically, for both SGX and default targets (with priority on SGX targets). We should run daily tests on CI to make sure that at least the example runtimes in this repository can be built deterministically using the documented process.

jberci:
Closes #136

### DIFF
--- a/.github/workflows/ci-reproducibility.yml
+++ b/.github/workflows/ci-reproducibility.yml
@@ -1,0 +1,44 @@
+# NOTE: This name appears in GitHub's Checks API and in workflow's status badge.
+name: ci-reproducibility
+
+# Trigger the workflow when:
+on:
+  push:
+    branches:
+      - main
+      - stable/*
+  # Besides pushes on the branches above, also check every day at 00:00 UTC.
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+
+  check-reproducible-build:
+    # NOTE: This name appears in GitHub's Checks API.
+    name: check-reproducibility
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Determine tag name
+        id: determine-tag
+        uses: ./.github/actions/determine-tag
+
+      - name: Build test binaries, 1.
+        id: build-1
+        uses: ./.github/actions/hash-rust
+        with:
+          binaries: test-runtime-simple-consensus,test-runtime-simple-keyvalue
+          image: oasisprotocol/runtime-builder:${{ steps.determine-tag.outputs.tag }}
+
+      - name: Build test binaries, 2.
+        id: build-2
+        uses: ./.github/actions/hash-rust
+        with:
+          binaries: test-runtime-simple-consensus,test-runtime-simple-keyvalue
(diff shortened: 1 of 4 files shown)


.github/workflows/ci-reproducibility.yml
```diff
@@ -0,0 +1,38 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Build test binaries, 1.
+        id: build-1
+        uses: ./.github/actions/hash-rust
+        with:
+          binaries: test-runtime-simple-consensus,test-runtime-simple-keyvalue
```
ptrus, line 28: This uses the default  `oasisprotocol/runtime-builder:v0.0` image right?
 
Should instead this use the image `oasisprotocol/runtime-builder:<determine_tag_name.output>`instead? So that on `main` it will use the `main` image and on `stable/*` branches the `stable/*` images build by `runtime-builder-image` action.

docs/guide/reproducing.md
```diff
@@ -0,0 +1,66 @@
...
+following command:
+
+```bash
+cargo elf2sgxs --release
+```
+
+It is necessary to change directories first because the tool does not
+currently support cargo workspaces.
+
+The resulting binaries will have the `.sgxs`. extension.
```
kostko, line 66:
```suggestion
The resulting binaries will have the `.sgxs` extension.
```

docs/guide/reproducing.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Running the Image
+
+The images are available in the `oasisprotocol/runtime-builder`
+repository on Docker Hub and are tagged with the same version numbers as
+releases of the SDK. To pull the image and run a container with it, run
+the following:
+
+```bash
+docker run -t -i -v /home/user/oasis-sdk:/src oasisprotocol/runtime-builder:v22.1 /bin/bash
```
kostko, line 31: Where does this tag come from?

kostko (approved):

docs/guide/reproducing.md
```diff
@@ -0,0 +1,70 @@
...
+
+```bash
+git clone https://github.com/oasisprotocol/oasis-sdk.git
+```
+
+## Running the Image
+
+The images are available in the `oasisprotocol/runtime-builder`
+repository on Docker Hub and are tagged with the same version numbers as
+releases of the SDK. To pull the image and run a container with it, run
```
ptrus, line 27: But there is no `v0.2.0` release :-)

ptrus (approved):

docs/guide/reproducing.md
kostko: What will be the first actual tag?

ptrus: could use `master` for now, until there's a release that includes the runtime-builder? (since the `runtime-builder` builds a `master` docker image on every master merge)

kostko: Ah yes, in that case let's change it to `master`.

PR #334, opened 2021-08-03
ts-web/ext-utils: keys change event with keys list

pro-wh:
add a field for the new keys. the extension now has to provide this when it tells the web content that the keys list has changed

### DIFF
--- a/client-sdk/ts-web/ext-utils/docs/changelog.md
+++ b/client-sdk/ts-web/ext-utils/docs/changelog.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased changes
+
+New features:
+
+- The keys change event now delivers the new keys list as part of the event.
+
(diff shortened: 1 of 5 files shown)


peterjgilbert (approved):

kostko (approved):

pro-wh:
thanks for the quick review

posting some context for the public:

we had found out that popular wallet extensions for other networks do not persist a user's decision not to share keys with a site. thus the next time the site asks, the extensions would prompt the user again. we have some dissenting opinions on whether that's a good thing, but we'll be supporting extensions that want to match that behavior.

in order to support that design, we'll need to rework the "dapp receives keys changed event and then asks for the keys list" concept. this change makes it so that the keys list is delivered as part of the event, so there is no need for an additional request for the keys list, which eliminates an extra unwanted popup from the extension in some circumstances.

PR #373, opened 2021-08-23
Add keymanager interface

### LINKED ISSUE #255: Key manager interface
The SDK should provide a nice way to interact with the key manager for confidential runtimes that have a key manager set.

jberci:
Closes #255


runtime-sdk/src/context.rs
```diff
@@ -65,6 +66,8 @@ pub trait Context {
     /// Returns a logger.
     fn get_logger(&self, module: &'static str) -> slog::Logger;
 
+    fn tokio_runtime(&self) -> &TokioRuntime;
```
kostko, line 69: Maybe just call it `tokio`?

tests/runtimes/simple-keyvalue/src/keyvalue.rs
```diff
@@ -180,6 +192,14 @@ impl Module {
         Ok(())
     }
 
+    fn tx_getcreatekey<C: TxContext>(ctx: &mut C, body: types::Key) -> Result<(), Error> {
+        let key_result = ctx.key_manager().unwrap().get_or_create_keys(KeyPairId::from(b"01234567890123456789012345678901".as_ref()));
+        match ctx.tokio_runtime().block_on(key_result) {
+            Ok(_) => Ok(()),
+            Err(err) => Err(Error::Core(core::Error::MalformedTransaction(err.into()))),
```
kostko, line 199: Add a new `Error` variant for this instead of abusing the `Core` one :-)

tests/runtimes/simple-keyvalue/src/lib.rs
```diff
@@ -38,6 +40,27 @@ impl sdk::Runtime for Runtime {
...
+                    "ekiden key manager test multisig key 2",
+                ]
+                .iter()
+                {
+                    let pk = PrivateKey::from_test_seed(seed.to_string());
+                    set.insert(pk.public_key());
+                }
+                set
+            },
+            threshold: 9001,
```
kostko, line 59: Could just use 2.

kostko (approved):

ptrus:
Btw this seems to have broke the [`oasis-core-commit-ref`](https://github.com/oasisprotocol/oasis-sdk/blob/5e48b25fba78335613b5b472f62497138685de5c/.github/actions/oasis-core-commit-ref/action.yml) action.

Since the `REF` is now:
https://github.com/oasisprotocol/oasis-sdk/blob/5e48b25fba78335613b5b472f62497138685de5c/.github/actions/oasis-core-commit-ref/action.yml#L13-L14

```
echo $REF                                   
a1c9fe982c32d550e8eac530bf8e79402728d64a
a509afa
```

Probably cause there are now multiple oasis-core dependencies in Cargo.lock

jberci:
Looks like I forgot to change the dependency for the simple-keyvalue runtime when rebasing. #403 should fix that.

PR #393, opened 2021-08-27
ts-web/rt: doc comments for signing

pro-wh:
this adds doc comments to some ts-web-specific API elements. notably, with the signing design being different from how it works in go, we could use this extra info


pro-wh:
image


that's weird, are we using an unpinned version of a dependency or something?

kostko (approved):

PR #404, opened 2021-09-01
runtime-builder/Dockerfile: use --locked when installing oasis-core tools

ptrus:
This should fix https://github.com/oasisprotocol/oasis-sdk/runs/3487648229


pro-wh (approved):
cf https://github.com/oasisprotocol/oasis-core/pull/4232

PR #90, opened 2021-04-09
ts-web/core: reflect structs

pro-wh:
use Go reflection to walk the type system, generating typescript definitions. this will reduce the manual work we have to do when we update oasis-core

PR #405, opened 2021-09-01
ts-web/core: minor types fixes

pro-wh:
detected some mismatches from the generated types from #90 

- some `omitempty` fields aren't marked as optional
- consensus `Parameters.MaxEvidenceSize` is uint64, so it should be longnum. I probably missed this when I was replacing MaxEvidenceNum. in practice it's 51200, so the difference never came up
- the lexicographical ordering between Signature and SignatureMultiSigned was swapped. possibly related to Signature having been collapsed from SignatureSignature :shrug: 

~~also I'm changing `toarray` structs to be multiline (only `StorageLogEntry` at this time). lint, please don't kill me~~ nvm lint did kill me

### DIFF
--- a/client-sdk/ts-web/core/docs/changelog.md
+++ b/client-sdk/ts-web/core/docs/changelog.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased changes:
+
+Bug fixes:
+
+- Fixes to type declarations that were inconsistent with the Go types.
+
 ## v0.1.0-alpha5
 
 Spotlight change:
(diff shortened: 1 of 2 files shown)


pro-wh:
aw prettier come on

kostko (approved):

pro-wh:
removing the multiline array change. they'll probably wrap it automatically if it gets longer. it'll be fine.

thanks for reviewing

pro-wh:
and adding a changelog item

PR #410, opened 2021-09-04
ts-web/core: minor types sync

pro-wh:
further discrepancies with a newly expanded generated types set

- `OpenExecutorCommitment` is serialized as `ExecutorCommitment`, which we further lower to `Signed`. not `ComputeBody` as incorrectly specified before
- a couple of structs were out of order


pro-wh:
oh no the branch name is all wrong

PR #419, opened 2021-09-09
OAS20 spec

ptrus:
ERC-20/CW-20 like spec for contracts-sdk with an example implementation

TODO:
- [x] do we want so support Allowances and/or any additional token metadata (e.g. see CW-20 for examples), or anything else?
  - probably not needed for now, can add it later in a non-breaking way (as optional extensions)
- [x] cleanup the example code, currently it was more of a playground
- [x] add some more tests

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ members = [
     "contract-sdk/types",
     "contract-sdk/storage",
     "contract-sdk-macros",
+    "contract-sdk/specs/oas20/types",
 
     # Test runtimes.
     "tests/runtimes/benchmarking",
@@ -22,6 +23,9 @@ members = [
 ]
 exclude = [
     # Test contracts.
-    "tests/contracts"
+    "tests/contracts",
+
+    # OAS contracts.
+    "contract-sdk/specs/oas20/contract",
 ]
 resolver = "2"
--- a/client-sdk/go/modules/contracts/oas20/oas20.go
+++ b/client-sdk/go/modules/contracts/oas20/oas20.go
@@ -1,0 +1,95 @@
+package oas20
+
+import (
+	"github.com/oasisprotocol/oasis-core/go/common/quantity"
+
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/contracts"
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/types"
+)
+
+// XXX: create a more highlevel helper wrapping Contracts client.
+
+// InitialBalance is the OAS20 contract initial balance information.
+type InitialBalance struct {
+	Address types.Address     `json:"address"`
+	Amount  quantity.Quantity `json:"amount"`
+}
+
+// MintingInformation is the OAS20 contract minting information.
+type MintingInformation struct {
+	Minter types.Address      `json:"minter"`
+	Cap    *quantity.Quantity `json:"cap,omitempty"`
+}
+
+// Instantiate is the OAS20 contract's initial state.
+type Instantiate struct {
+	// Name is the name of the token.
+	Name string `json:"name"`
+	// Symbol is the token symbol.
+	Symbol string `json:"symbol"`
+	// Decimals is the number of token decimals.
+	Decimals uint8 `json:"decimals"`
+	// InitialBalances are the initial balances of the token.
+	InitialBalances []InitialBalance `json:"initial_balances,omitempty"`
+	// Minting is the information about minting in case the token supports minting.
+	Mintting *MintingInformation `json:"minting,omitempty"`
+}
+
+// Transfer is the OAS20 contract's transfer request.
+type Transfer struct {
+	To     types.Address     `json:"to"`
+	Amount quantity.Quantity `json:"amount"`
+}
+
+// Send is the OAS20 contract's send request.
+type Send struct {
+	To     contracts.InstanceID `json:"to"`
+	Amount quantity.Quantity    `json:"amount"`
+	Data   interface{}          `json:"data"`
+}
+
+// Balance is the OAS20 contract's balance query request.
+type Balance struct {
+	Address types.Address `json:"address"`
+}
+
+// TokenInformation is the OAS20 contract's token information request.
+type TokenInformation struct{}
+
+// Request is an OAS20 contract request.
+type Request struct {
+	Instantiate      *Instantiate      `json:"instantiate,omitempty"`
+	Transfer         *Transfer         `json:"transfer,omitempty"`
+	Send             *Send             `json:"send,omitempty"`
+	TokenInformation *TokenInformation `json:"token_information,omitempty"`
+	Balance          *Balance          `json:"balance,omitempty"`
+}
+
+// TokenInformationResponse is the token information response.
+type TokenInformationResponse struct {
+	// Name is the name of the token.
+	Name string `json:"name"`
+	// Symbol is the token symbol.
+	Symbol string `json:"symbol"`
+	// Decimals is the number of token decimals.
+	Decimals uint8 `json:"decimals"`
+	// TotalSupply is the total supply of the token.
+	TotalSupply quantity.Quantity `json:"total_supply"`
+	// Minting is the information about minting in case the token supports minting.
+	Mintting *MintingInformation `json:"minting,omitempty"`
+}
+
+// Empty is the empty response.
+type Empty struct{}
+
+// Response is an OAS20 contract response.
+type Response struct {
+	TokenInformation *TokenInformationResponse `json:"token_information,omitempty"`
+	Balance          *BalanceResponse          `json:"balance,omitempty"`
+	Empty            *Empty                    `json:"empty,omitempty"`
+}
+
+// BalanceResponse is the OAS20 balance response.
+type BalanceResponse struct {
+	Balance quantity.Quantity `json:"balance"`
+}
--- a/client-sdk/go/modules/contracts/types.go
+++ b/client-sdk/go/modules/contracts/types.go
@@ -1,6 +1,8 @@
 package contracts
 
 import (
+	"encoding/binary"
+
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/hash"
 
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/types"
@@ -11,6 +13,13 @@ type CodeID uint64
 
 // InstanceID is the unique deployed code instance identifier.
 type InstanceID uint64
+
+// Address returns address for the InstanceID.
+func (i *InstanceID) Address() types.Address {
+	id := make([]byte, 8)
+	binary.BigEndian.PutUint64(id, uint64(*i))
+	return types.NewAddressForModule("contracts", id)
+}
 
 // Policy is a generic policy that specifies who is allowed to perform an action.
 type Policy struct {
--- a/contract-sdk/specs/oas20/README.md
+++ b/contract-sdk/specs/oas20/README.md
@@ -1,0 +1,330 @@
+# OAS-20 spec: Fungible Tokens
+
+Specification for standardization of fungible token contracts within the Oasis contracts-sdk framework. Roughly based on [ERC20] and [CW20] specs.
+
+[ERC20]: https://ethereum.org/en/developers/docs/standards/tokens/erc-20/
+[CW20]: https://github.com/CosmWasm/cw-plus/blob/main/packages/cw20/README.md
+
+## Requests
+
+### Instantiate
+
+```rust
+#[cbor(rename = "instantiate")]
+Instantiate(TokenInstantiation),
+
+/// OAS20 token instantiation information.
+#[derive(Debug, Default, Clone, PartialEq, Eq, cbor::Decode, cbor::Encode)]
+pub struct TokenInstantiation {
+    /// Name of the token.
+    pub name: String,
+    /// Token symbol.
+    pub symbol: String,
+    /// Number of decimals.
+    pub decimals: u8,
+    /// Initial balances of the token.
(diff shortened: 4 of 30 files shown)


Cargo.toml
```diff
@@ -19,8 +19,11 @@ members = [
     "tests/runtimes/simple-keyvalue",
     "tests/runtimes/simple-consensus",
     "tests/runtimes/simple-evm",
+
+    # OAS contracts specs.
+    "runtime-sdk/modules/contracts/specs/oas20",
```
kostko, line 24: It would probably make more sense to put these under `contract-sdk/contracts`?

ptrus: Hm yeah i actually had it there earlier, then I figured that the code actually cannot only depend on the `contract-sdk` crates, but also the `runtime-sdk/modules/contracts` due to the Id->Address derivation for contracts (which is defined in `modules/contracts` and also depends on MODULE_NAME).

However even with that dependency i think it does make more sense that it lives in `contract-sdk/contracts` for now. I'll move it back.

kostko: Hm the contract part should not depend on the runtime SDK as then it will not be usable from contracts. If needed, we should copy some (minimal) functionality into the `oasis-contract-sdk-types` crate which is shared between the runtime SDK module and the contract SDK. This is what has been done for things like the `Address` type.

ptrus: Oh. For context i needed for this:
https://github.com/oasisprotocol/oasis-sdk/blob/bf3d54dc7a3ec1e76994353be566cdf15a2489fd/runtime-sdk/modules/contracts/src/types.rs#L85-L87

so I could get a contract address from its instance id. One option would be moving this id->address mapping to the `oasis-contract-sdk-types` (and make it not rely on module name?).  Or having access to the id->address (and otherway) mapping available somewhere in the `context` or `env`.

contract-sdk/specs/oas20/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+`MinterInfo` - Returns the minter address and the minting cap.
+
+```rust
+/// Token mintable information.
+#[derive(Debug, Default, Clone, cbor::Decode, cbor::Encode, PartialEq, Eq)]
+pub struct MintableInformation {
+    /// Caller address which is allowed to mint new tokens.
+    pub minter: Address,
+    /// Maximum number of minted tokens.
```
pro-wh, line 79: I checked the sample implementation, and the cap is a limit for the total supply. that might be subtly different from 'number of minted tokens' when (i) the token starts out with some nonzero supply and/or (ii) someone burns tokens

ptrus: That's true, i'll update the docs, since cap being the total supply limit is what makes more sense. It's also what related specs do (CW20).

Cargo.toml
```diff
@@ -22,5 +23,8 @@ members = [
 ]
 exclude = [
     # Test contracts.
-    "tests/contracts"
+    "tests/contracts",
+
+    # OAS contracts.
+    "contract-sdk/specscontract-sdk/specs/oas20/contract",
```
kostko, line 29: Looks like the first part is duplicated.

client-sdk/go/modules/contracts/oas20/oas20.go
```diff
@@ -0,0 +1,78 @@
+package oas20
+
+import (
+	"github.com/oasisprotocol/oasis-core/go/common/quantity"
+
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/contracts"
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/types"
+)
+
+// XXX: create a more highlevel helper wrapping Contracts client.
```
kostko, line 10: Would be great to file an issue.

client-sdk/go/modules/contracts/oas20/oas20.go
```diff
@@ -0,0 +1,78 @@
...
+	Amount  quantity.Quantity `json:"amount"`
+}
+
+// MintableInformation is the OAS20 contract minting information.
+type MintableInformation struct {
+	Minter types.Address      `json:"minter"`
+	Cap    *quantity.Quantity `json:"cap,omitempty"`
+}
+
+// Instantiate is the OAS20 contract initiate data.
```
kostko, line 24:
```suggestion
// Instantiate is the OAS20 contract's initial state.
```

client-sdk/go/modules/contracts/oas20/oas20.go
```diff
@@ -0,0 +1,78 @@
...
+	Decimals uint8 `json:"decimals"`
+	// InitialBalances are the initial balances of the token.
+	InitialBalances []InitialBalance `json:"initial_balances,omitempty"`
+	// Mintable is the information about minting in case the token is mintable.
+	Mintable *MintableInformation `json:"mintable,omitempty"`
+}
+
+// InstantiateCall is the OAS20 contract instnatiate call.
+func InstantiateCall(instantiate *Instantiate) map[string]*Instantiate {
+	return map[string]*Instantiate{
```
kostko, line 40: You could define request/response types in Go as well?

client-sdk/go/modules/contracts/types.go
```diff
@@ -12,6 +14,13 @@ type CodeID uint64
 // InstanceID is the unique deployed code instance identifier.
 type InstanceID uint64
 
+// Address returns address for the InstanceID.
+func (i *InstanceID) Address() types.Address {
+	id := make([]byte, 8)
+	binary.BigEndian.PutUint64(id, uint64(*i))
+	return types.NewAddressForModule("contracts", id)
+}
```
kostko, line 22: Would be good to have a test case to make sure it is equal to the Rust side.

kostko (commented):
In progress (8/25).

client-sdk/go/modules/contracts/oas20/oas20.go
ptrus: Yeah i will in case it won't be done in this PR

client-sdk/go/modules/contracts/types.go
ptrus: It was implicitly tested via `NewAddressForModule` tests before. Added explicit tests for `InstanceId`->`Address` now to both rust and go sides.

contract-sdk/specs/oas20/types/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+pub enum Event {
+    #[sdk_event(code = 1)]
+    Oas20Instantiated { token_info: TokenInformation }, // TODO: include instance id?
+
+    #[sdk_event(code = 2)]
+    Oas20Transferer {
+        from: Address,
+        to: Address,
+        amount: u128,
+    }, // TODO: include token info, instance id?
```
kostko, line 112: Hm, should we include instance ID as part of all contract events (e.g. where we do the mapping in the contracts module)? Code ID is already included as part of the module (e.g. the module for these events is `contracts.<code-id>`).

And for token info, any reason why this would need to be included in the event instead of just being queryable once you have the instance ID?

contract-sdk/specs/oas20/types/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+    #[cbor(rename = "balance")]
+    Balance { address: Address },
+}
+
+/// All possible responses that the OAS20 contract can return.
+///
+/// This includes both calls and queries.
+#[derive(Clone, Debug, PartialEq, cbor::Encode, cbor::Decode)]
+pub enum Response {
+    #[cbor(rename = "token_information_response")]
```
kostko, line 166: Can probably drop the `_response`/`Response` suffix as these are all responses.

contract-sdk/specs/oas20/types/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+    #[cbor(rename = "balance_response")]
+    BalanceResponse { balance: u128 },
+
+    #[cbor(rename = "empty")]
+    Empty,
+}
+
+/// Oas20 receiver request. Contracts expecting to receive Oas20 tokens should
+/// implement his request.
+/// TODO: https://github.com/oasisprotocol/oasis-cbor/issues/19 would make embedding this nicer.
```
kostko, line 178: This is now available in oasis-cbor 0.2.0.

contract-sdk/types/src/address.rs
```diff
@@ -40,6 +40,20 @@ impl Address {
         Ok(Self(a))
     }
 
+    /// Tries to create a new address from Bech32-encoded string.
+    pub fn from_bech32(data: &str) -> Result<Self, Error> {
+        let (hrp, data, variant) = bech32::decode(data).map_err(|_| Error::MalformedAddress)?;
+        if hrp != ADDRESS_BECH32_HRP {
+            return Err(Error::MalformedAddress);
```
kostko, line 47: Would be great to have tests for these error cases.

tests/contracts/hello/src/lib.rs
```diff
@@ -83,6 +84,10 @@ pub enum Request {
 
     #[cbor(rename = "upgrade_fail_post")]
     UpgradeFailPost,
+
+    #[cbor(rename = "oas20_receive")]
+    // TODO: is there a nicer way to compose requests, maybe: https://github.com/oasisprotocol/oasis-cbor/issues/19
```
kostko, line 89: Now there is.

kostko (commented):
We should probably include the specs contracts tests in the CI test flow and coverage.

contract-sdk/specs/oas20/contract/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+    fn call<C: sdk::Context>(ctx: &mut C, request: Request) -> Result<Response, Error> {
+        // This method is called for each contracts.Call call. It is supposed to handle the request
+        // and return a response.
+        match request {
+            Request::Transfer { to, amount } => {
+                // Transfers the `amount` of funds from caller to `to` address.
+                let from = ctx.caller_address().to_owned();
+                Self::transfer(ctx, from, to, amount)?;
+
+                ctx.emit_event(Event::Oas20Transferer { from, to, amount });
```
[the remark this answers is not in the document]
ptrus: Yeah that was the plan, will fix the typo :grin:

contract-sdk/specs/oas20/README.md
```diff
@@ -0,0 +1,83 @@
+# OAS-20 spec: Fungible Tokens
+
+Specification for standardization of fungible token contracts within the Oasis contracts-sdk framework. Roughly based on [ERC20] and [CW20] specs.
+
+[ERC20]: https://ethereum.org/en/developers/docs/standards/tokens/erc-20/
+[CW20]: https://github.com/CosmWasm/cw-plus/blob/main/packages/cw20/README.md
+
```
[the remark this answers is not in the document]
ptrus: Makes sense to include yes, didn't thought about those until after i coded up the contract, then forgot to include events and errors in the spec :)

contract-sdk/specs/oas20/types/src/lib.rs
ptrus: Yeah if we have instance ID then the token info is not needed.  Will open an issue to include instance ID in all contract events, and remove the comments to include it here.

contract-sdk/specs/oas20/README.md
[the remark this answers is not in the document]
ptrus: Done, also refactored the structure of the document a bit, and updated the errors a bit (removed an unused one, and some renames).

contract-sdk/types/src/address.rs
ptrus: Added, also to the `runtime-sdk` address module

ptrus:
> We should probably include the specs contracts tests in the CI test flow and coverage.

Added

tests/e2e/contracts.go
```diff
@@ -89,9 +94,159 @@ func ContractsTest(sc *RuntimeScenario, log *logging.Logger, conn *grpc.ClientCo
...
+		[]types.BaseUnits{},
+	).
+		SetFeeGas(1_000_000).
+		AppendAuthSignature(signer.Public(), nonce+5)
+	_ = tb.AppendSign(ctx, signer)
+	if err := tb.SubmitTx(ctx, &rawResult); err != nil {
+		return fmt.Errorf("failed to call OAS20 transfer: %w", err)
+	}
+
+	// TODO: watch emitted event.
```
kostko, line 159: I guess it would be great to have some more high-level wrapper for processing events (as currently you just get the raw tags) in the Go client-sdk and then specifically for contract events. Probably would need to be something generic with some per-module helpers for those specific events and then something for contracts on top of that.

kostko (commented):
Looks nice!

tests/e2e/contracts.go
```diff
@@ -89,9 +94,159 @@ func ContractsTest(sc *RuntimeScenario, log *logging.Logger, conn *grpc.ClientCo
...
+
+	if err := cbor.Unmarshal(rawResult, &result); err != nil {
+		return fmt.Errorf("failed to decode contract result: %w", err)
+	}
+
+	if result["hello"]["greeting"] != fmt.Sprintf("hello e2e test (%d)", counter) {
 		return fmt.Errorf("unexpected result from contract: %+v", result)
 	}
+	// Calling say_hello bumps the counter.
+	// counter++
```
kostko, line 228: Is this commented out because this would otherwise throw the ineffective lint error?

kostko (approved):

tests/e2e/contracts.go
ptrus: yeah, i guess i could also use the nolint directive

PR #423, opened 2021-09-10
runtime-sdk/modules/evm: Add withdraw/deposit

### LINKED ISSUE #351: runtime-sdk/modules/evm: Add withdraw/deposit
We need `withdraw` and `deposit` API methods that will move tokens between SDK accounts and EVM accounts.  For SDK accounts, an additional denomination parameter will need to be specified, since they support multiple denominations.

To track the tokens that have been moved to EVM accounts inside the EVM module, a token pool for the EVM module needs to be created, e.g.:
```
pub static ADDRESS_EVM_TOKENS: Lazy<Address> =
    Lazy::new(|| Address::from_module(MODULE_NAME, "evm-tokens"));
```
When a `deposit` is made, tokens are moved to this address and also deposited to the given EVM account address.
When a `withdraw`-al is made, tokens are removed from the given EVM account address and moved from the EVM module token pool to the given SDK address.

An invariant check should be added to ensure that the total balance of EVM accounts is equal to the total balance of tokend in the EVM_TOKENS pool.

Tokens spent for EVM gas are moved into the fee accumulator of the SDK accounts module, so they get automatically disbursed to entities.

abukosek:
Closes #351.

### DIFF
--- a/client-sdk/go/modules/evm/evm.go
+++ b/client-sdk/go/modules/evm/evm.go
@@ -4,12 +4,15 @@ import (
 	"context"
 
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/client"
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/types"
 )
 
 const (
 	// Callable methods.
-	methodCreate = "evm.Create"
-	methodCall   = "evm.Call"
+	methodCreate   = "evm.Create"
+	methodCall     = "evm.Call"
+	methodDeposit  = "evm.Deposit"
+	methodWithdraw = "evm.Withdraw"
 
 	// Queries.
 	methodPeekStorage = "evm.PeekStorage"
@@ -19,10 +22,26 @@ const (
 // V1 is the v1 EVM module interface.
 type V1 interface {
 	// Create generates an EVM CREATE transaction.
-	Create(value []byte, initCode []byte, gasPrice []byte, gasLimit uint64) *client.TransactionBuilder
+	// Note that the transaction's gas limit should be set to cover both the
+	// SDK gas limit and the EVM gas limit.  The transaction fee should be
+	// high enough to cover the EVM gas price multiplied by the EVM gas limit.
+	Create(value []byte, initCode []byte) *client.TransactionBuilder
 
 	// Call generates an EVM CALL transaction.
-	Call(address []byte, value []byte, data []byte, gasPrice []byte, gasLimit uint64) *client.TransactionBuilder
+	// Note that the transaction's gas limit should be set to cover both the
+	// SDK gas limit and the EVM gas limit.  The transaction fee should be
+	// high enough to cover the EVM gas price multiplied by the EVM gas limit.
+	Call(address []byte, value []byte, data []byte) *client.TransactionBuilder
+
+	// Deposit generates a deposit transaction that moves tokens from the
+	// given SDK account into the given EVM account.  The denomination must
+	// be identical to the denomination set in the EVM module's parameters.
+	Deposit(from types.Address, to []byte, amount types.BaseUnits) *client.TransactionBuilder
+
+	// Withdraw generates a withdraw transaction that moves tokens from the
+	// given EVM account into the given SDK account.  The denomination must
+	// be identical to the denomination set in the EVM module's parameters.
+	Withdraw(from []byte, to types.Address, amount types.BaseUnits) *client.TransactionBuilder
 
 	// PeekStorage queries the EVM storage.
 	PeekStorage(ctx context.Context, address []byte, index []byte) ([]byte, error)
@@ -36,23 +55,37 @@ type v1 struct {
 }
 
 // Implements V1.
-func (a *v1) Create(value []byte, initCode []byte, gasPrice []byte, gasLimit uint64) *client.TransactionBuilder {
-	return client.NewTransactionBuilder(a.rtc, methodCreate, &CreateTx{
+func (a *v1) Create(value []byte, initCode []byte) *client.TransactionBuilder {
+	return client.NewTransactionBuilder(a.rtc, methodCreate, &Create{
 		Value:    value,
 		InitCode: initCode,
-		GasPrice: gasPrice,
-		GasLimit: gasLimit,
 	})
 }
 
 // Implements V1.
-func (a *v1) Call(address []byte, value []byte, data []byte, gasPrice []byte, gasLimit uint64) *client.TransactionBuilder {
-	return client.NewTransactionBuilder(a.rtc, methodCall, &CallTx{
-		Address:  address,
-		Value:    value,
-		Data:     data,
-		GasPrice: gasPrice,
-		GasLimit: gasLimit,
+func (a *v1) Call(address []byte, value []byte, data []byte) *client.TransactionBuilder {
+	return client.NewTransactionBuilder(a.rtc, methodCall, &Call{
+		Address: address,
+		Value:   value,
+		Data:    data,
+	})
+}
+
+// Implements V1.
+func (a *v1) Deposit(from types.Address, to []byte, amount types.BaseUnits) *client.TransactionBuilder {
+	return client.NewTransactionBuilder(a.rtc, methodDeposit, &Deposit{
+		From:   from,
+		To:     to,
+		Amount: amount,
+	})
+}
+
+// Implements V1.
+func (a *v1) Withdraw(from []byte, to types.Address, amount types.BaseUnits) *client.TransactionBuilder {
+	return client.NewTransactionBuilder(a.rtc, methodWithdraw, &Withdraw{
+		From:   from,
+		To:     to,
+		Amount: amount,
 	})
 }
 
--- a/client-sdk/go/modules/evm/types.go
+++ b/client-sdk/go/modules/evm/types.go
@@ -1,33 +1,46 @@
 package evm
 
-// The CreateTx type must match the CreateTx type from the evm module types
+import (
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/types"
+)
+
+// The types in this file must match the types from the evm module types
 // in runtime-sdk/modules/evm/src/types.rs.
-type CreateTx struct {
+
+// Create is an EVM CREATE transaction.
+type Create struct {
 	Value    []byte `json:"value"`
 	InitCode []byte `json:"init_code"`
-	GasPrice []byte `json:"gas_price"`
-	GasLimit uint64 `json:"gas_limit"`
 }
 
-// The CallTx type must match the CallTx type from the evm module types
-// in runtime-sdk/modules/evm/src/types.rs.
-type CallTx struct {
-	Address  []byte `json:"address"`
-	Value    []byte `json:"value"`
-	Data     []byte `json:"data"`
-	GasPrice []byte `json:"gas_price"`
-	GasLimit uint64 `json:"gas_limit"`
+// Call is an EVM CALL transaction.
+type Call struct {
+	Address []byte `json:"address"`
+	Value   []byte `json:"value"`
+	Data    []byte `json:"data"`
 }
 
-// The PeekStorageQuery type must match the PeekStorageQuery type from the
-// evm module types in runtime-sdk/modules/evm/src/types.rs.
+// Deposit is a transaction that deposits tokens into an EVM account.
+type Deposit struct {
+	From   types.Address   `json:"from"`
+	To     []byte          `json:"to"`
+	Amount types.BaseUnits `json:"amount"`
+}
+
+// Withdraw is a transaction that withdraws tokens from an EVM account.
+type Withdraw struct {
+	From   []byte          `json:"from"`
+	To     types.Address   `json:"to"`
+	Amount types.BaseUnits `json:"amount"`
+}
+
+// PeekStorageQuery queries the EVM storage.
 type PeekStorageQuery struct {
 	Address []byte `json:"address"`
 	Index   []byte `json:"index"`
 }
 
-// The PeekCodeQuery type must match the PeekCodeQuery type from the
-// evm module types in runtime-sdk/modules/evm/src/types.rs.
+// PeekCodeQuery queries the EVM code storage.
 type PeekCodeQuery struct {
 	Address []byte `json:"address"`
 }
--- a/runtime-sdk/modules/evm/src/lib.rs
+++ b/runtime-sdk/modules/evm/src/lib.rs
@@ -92,16 +92,12 @@ pub enum Error {
     #[sdk_error(code = 6)]
     GasLimitTooLow(u64),
 
-    #[error("gas price too low")]
-    #[sdk_error(code = 7)]
-    GasPriceTooLow,
-
     #[error("insufficient balance")]
-    #[sdk_error(code = 8)]
+    #[sdk_error(code = 7)]
     InsufficientBalance,
 
     #[error("invalid denomination")]
-    #[sdk_error(code = 9)]
+    #[sdk_error(code = 8)]
     InvalidDenomination,
 
     #[error("core: {0}")]
@@ -119,8 +115,6 @@ pub struct GasCosts {
 /// Parameters for the EVM module.
 #[derive(Clone, Default, Debug, cbor::Encode, cbor::Decode)]
 pub struct Parameters {
-    /// Minimum acceptable gas price for EVM transactions.
-    pub min_gas_price: U256,
     /// Token denomination used for the EVM token.
     pub token_denomination: token::Denomination,
     /// Gas costs.
@@ -349,15 +343,9 @@ impl<Cfg: Config> Module<Cfg> {
     {
         let params = Self::params(ctx.runtime_state());
         let den = params.token_denomination;
-        let min_gas_price: primitive_types::U256 = params.min_gas_price.into();
 
         let gas_limit: u64 = core::Module::remaining_tx_gas(ctx);
         let gas_price: primitive_types::U256 = ctx.tx_auth_info().fee.gas_price().into();
-
-        // Make sure that gas_price >= min_gas_price.
-        if gas_price < min_gas_price {
-            return Err(Error::GasPriceTooLow);
-        }
 
         let vicinity = evm_backend::Vicinity {
             gas_price: gas_price.into(),
--- a/runtime-sdk/modules/evm/src/lib.rs
+++ b/runtime-sdk/modules/evm/src/lib.rs
@@ -2,10 +2,13 @@
 pub mod evm_backend;
 pub mod types;
 
+use std::collections::BTreeMap;
+
 use evm::{
     executor::{MemoryStackState, StackExecutor, StackSubstateMetadata},
-    Config,
+    Config as EVMConfig,
 };
+use once_cell::sync::Lazy;
 use thiserror::Error;
 use tiny_keccak::{Hasher, Keccak};
 
@@ -14,8 +17,17 @@ use oasis_runtime_sdk::{
     crypto::signature::PublicKey,
     error,
     module::{self, CallResult, Module as _},
-    modules, storage,
-    types::transaction::AddressSpec,
+    modules::{
+        self,
+        accounts::API as _,
+        core::{self, Error as CoreError, API as _},
+    },
+    storage,
+    types::{
+        address::Address,
+        token,
+        transaction::{AddressSpec, AuthInfo, Transaction},
+    },
 };
 
 use evm::backend::ApplyBackend;
@@ -39,7 +51,19 @@ pub mod state {
     pub const BLOCK_HASH_WINDOW_SIZE: u64 = 256;
 }
 
-pub struct Module;
+/// Module configuration.
+pub trait Config: 'static {
+    /// Module that is used for accessing accounts.
+    type Accounts: modules::accounts::API;
+}
+
+pub struct Module<Cfg: Config> {
+    _cfg: std::marker::PhantomData<Cfg>,
+}
+
+/// EVM token pool address.
+pub static ADDRESS_EVM_TOKENS: Lazy<Address> =
+    Lazy::new(|| Address::from_module(MODULE_NAME, "evm-tokens"));
 
 /// Errors emitted by the EVM module.
 #[derive(Error, Debug, oasis_runtime_sdk::Error)]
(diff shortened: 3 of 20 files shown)


runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -241,41 +372,66 @@ impl Module {
...
             Err(Error::EVMError(format!("{:?}", exit_reason)))
         }
     }
 
-    fn derive_caller<C>(ctx: &mut C) -> Result<H160, Error>
-    where
-        C: TxContext,
-    {
-        match &ctx.tx_auth_info().signer_info[0].address_spec {
+    fn derive_caller_from_bytes(b: &[u8]) -> Result<H160, Error> {
```
kostko, line 405: Can this ever error? Just always return `H160`.

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -356,9 +522,103 @@ impl module::MigrationHandler for Module {
     }
 }
 
-impl module::AuthHandler for Module {}
+impl<Accounts: modules::accounts::API> module::AuthHandler for Module<Accounts> {
+    fn authenticate_tx<C: Context>(ctx: &mut C, tx: &Transaction) -> Result<(), CoreError> {
+        // We're only interested in transactions that can be paid with tokens
+        // from the corresponding EVM account.
+        if !(tx.call.method == "evm.Create"
```
kostko, line 529: Could you use a `match` instead?

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -356,9 +522,103 @@ impl module::MigrationHandler for Module {
...
+        // Take the tokens from the EVM account and move them into the
+        // fee accumulator.
+        let state = ctx.runtime_state();
+        let store = storage::PrefixStore::new(state, &MODULE_NAME);
+        let mut accounts =
+            storage::TypedStore::new(storage::PrefixStore::new(store, &state::ACCOUNTS));
+        let mut account: evm_backend::Account = accounts.get(&evm_acct_addr).unwrap_or_default();
+
+        if account.balance < total_fees.into() {
+            return Err(CoreError::InsufficientFeeBalance);
```
kostko, line 607: So currently, you must always first deposit enough into the EVM account to cover any gas fees for the EVM invocations, right?

runtime-sdk/src/module.rs
```diff
@@ -277,7 +277,7 @@ pub trait AuthHandler {
         _tx: &Transaction,
     ) -> Result<(), modules::core::Error> {
         // Default implementation doesn't do any checks.
```
kostko, line 279: Change this comment to say that the default implementation rejects all transactions.

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -98,6 +98,10 @@ pub enum Error {
     #[error("{0}")]
     #[sdk_error(transparent, abort)]
     Abort(#[source] dispatcher::Error),
+
+    #[error("authenticate_tx not handled")]
+    #[sdk_error(code = 19)]
+    AuthenticateTxNotHandled,
```
kostko, line 104: Maybe rename this to something like "no module could authenticate the transaction", `NotAuthenticated`.

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -356,9 +522,103 @@ impl module::MigrationHandler for Module {
...
+                if call.gas_price < min_gas_price {
+                    return Err(CoreError::MalformedTransaction(anyhow::Error::new(
+                        Error::GasPriceTooLow,
+                    )));
+                }
+                let evm_fee = call
+                    .gas_price
+                    .checked_mul(U256::from(call.gas_limit))
+                    .ok_or(CoreError::GasOverflow)?
+                    .checked_add(call.value)
```
kostko, line 569: Value should probably be separate from the gas fee as that is only moved in case the transaction is successful? Also below.

kostko (commented):
What happens if the Accounts module is defined before the EVM module (so that the accounts authentication handler runs first)? It seems like this could cause some problems as the SDK handler wouldn't know about the EVM part.

Could there be a design where the maximum fee was part of `auth_info.fee.amount` instead of being supplied separately as `gas_price * gas_limit` within the call? You could just use `auth_info.fee.gas` as the `gas_limit` (after reducing for any base SDK fees, you can use `Core::remaining_tx_gas`) and then `gas_price` is defined as usual.

This would make it possible to use either SDK or EVM accounts as source of funds.

abukosek:
Thanks for the comments!  I've started working on changing the gas payment as you've described.

Currently, the e2e tests fail because the accounts module's invariant checker finds that the gas fee gets lost somewhere -- I'll continue debugging this on Monday :)
(It's probably something really simple, but I've been looking at the same code all day and can't really notice anything that could cause this at the moment.)

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -356,9 +552,56 @@ impl module::MigrationHandler for Module {
...
 
-impl module::AuthHandler for Module {}
+impl<Accounts: modules::accounts::API> module::AuthHandler for Module<Accounts> {
+    fn authenticate_tx<C: Context>(ctx: &mut C, tx: &Transaction) -> Result<(), CoreError> {
+        // We're only interested in transactions that can be paid with tokens
+        // from the corresponding EVM account.
+        match tx.call.method.as_str() {
+            "evm.Create" => {}
+            "evm.Call" => {}
+            "evm.Withdraw" => {}
```
kostko, line 562:
```suggestion
            "evm.Create" | "evm.Call" | "evm.Withdraw" => {}
```

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -356,9 +552,56 @@ impl module::MigrationHandler for Module {
     }
 }
 
-impl module::AuthHandler for Module {}
+impl<Accounts: modules::accounts::API> module::AuthHandler for Module<Accounts> {
+    fn authenticate_tx<C: Context>(ctx: &mut C, tx: &Transaction) -> Result<(), CoreError> {
```
kostko, line 556: Don't forget to also do nonce checks/updates here.

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -241,41 +372,66 @@ impl Module {
...
         }
     }
 
-    fn derive_caller<C>(ctx: &mut C) -> Result<H160, Error>
-    where
-        C: TxContext,
-    {
-        match &ctx.tx_auth_info().signer_info[0].address_spec {
+    fn derive_caller_from_bytes(b: &[u8]) -> Result<H160, Error> {
+        // Caller address is derived by doing Keccak-256 on the
```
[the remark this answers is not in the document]
abukosek: I intend to do this in an upcoming PR, where I'll also add unit tests for the new EVM code.

client-sdk/go/modules/evm/types.go
```diff
@@ -1,33 +1,50 @@
...
-// The CallTx type must match the CallTx type from the evm module types
-// in runtime-sdk/modules/evm/src/types.rs.
-type CallTx struct {
+// Call is an EVM CALL transaction.
+type Call struct {
 	Address  []byte `json:"address"`
 	Value    []byte `json:"value"`
 	Data     []byte `json:"data"`
 	GasPrice []byte `json:"gas_price"`
 	GasLimit uint64 `json:"gas_limit"`
```
kostko, line 24: Are these fields still needed after the recent changes?

runtime-sdk/modules/evm/src/types.rs
```diff
@@ -11,14 +12,30 @@ pub struct CreateTx {
 
 /// Transaction body for calling an EVM contract.
 #[derive(Clone, Debug, cbor::Encode, cbor::Decode)]
-pub struct CallTx {
+pub struct Call {
     pub address: H160,
     pub value: U256,
     pub data: Vec<u8>,
     pub gas_price: U256,
     pub gas_limit: u64,
```
kostko, line 20: Are these fields still needed with the recent changes?

client-sdk/go/modules/evm/evm.go
```diff
@@ -24,6 +27,12 @@ type V1 interface {
 	// Call generates an EVM CALL transaction.
 	Call(address []byte, value []byte, data []byte, gasPrice []byte, gasLimit uint64) *client.TransactionBuilder
 
+	// Deposit generates a deposit transaction.
+	Deposit(from types.Address, to []byte, amount types.BaseUnits) *client.TransactionBuilder
+
+	// Withdraw generates a withdraw transaction.
+	Withdraw(from []byte, to types.Address, amount types.BaseUnits) *client.TransactionBuilder
```
kostko, line 34: Maybe use some more informative documentation comments (e.g. explain from/to where the deposit/withdraw happens).

runtime-sdk/src/module.rs
```diff
@@ -276,8 +276,8 @@ pub trait AuthHandler {
         _ctx: &mut C,
         _tx: &Transaction,
     ) -> Result<(), modules::core::Error> {
-        // Default implementation doesn't do any checks.
-        Ok(())
+        // Default implementation rejects all transactions.
+        Err(modules::core::Error::NotAuthenticated)
```
kostko, line 280: Add a quick unit test that a runtime without any auth handler modules will reject transactions.

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -174,11 +233,98 @@ impl API for Module {
...
+        if is_check_only {
+            return Ok(());
+        }
+
+        // Decrease EVM account's balance by the amount of tokens requested.
+        account.balance -= amount.amount().into();
+        accounts.insert(&from, account);
+
+        // Transfer tokens from EVM pool into SDK account.
+        if Cfg::Accounts::transfer(ctx, *ADDRESS_EVM_TOKENS, to, &amount).is_err() {
```
kostko, line 320: Use `.map_err(|_| Error::InsufficientBalance)?;` instead.

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -174,11 +233,98 @@ impl API for Module {
...
+                Cfg::Accounts::get_balance(ctx.runtime_state(), from, params.token_denomination)
+                    .unwrap();
+            if bal < amount.amount() {
+                return Err(Error::InsufficientBalance);
+            }
+            return Ok(());
+        }
+
+        // Transfer tokens from SDK account into EVM pool.
+        if Cfg::Accounts::transfer(ctx, from, *ADDRESS_EVM_TOKENS, &amount).is_err() {
```
kostko, line 269: Use `.map_err(|_| Error::InsufficientBalance)?;` instead.

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -212,71 +358,119 @@ impl Module {
...
+
+        let gas_price: primitive_types::U256 = gas_price_in.into();
+
+        // Make sure that gas_price >= min_gas_price.
         if gas_price < min_gas_price {
             return Err(Error::GasPriceTooLow);
         }
 
+        // Make sure we have enough core gas left.
+        if core::Module::remaining_tx_gas(ctx) < gas_limit {
```
kostko, line 375: Shouldn't it instead compute `gas_limit` from `remaining_tx_gas`? Or why is this separate `gas_limit` parameter needed?

runtime-sdk/modules/evm/src/lib.rs
kostko: With the recent change, should this return not authenticated as maybe the core account has enough funds?

runtime-sdk/src/modules/accounts/mod.rs
```diff
@@ -174,6 +174,20 @@ pub trait API {
...
+        from: Address,
+        amount: &token::BaseUnits,
+    ) -> Result<(), modules::core::Error>;
+
+    // Move amount from fee accumulator into address.
+    fn move_from_fee_accumulator<C: Context>(
+        ctx: &mut C,
+        to: Address,
+        amount: &token::BaseUnits,
+    ) -> Result<(), modules::core::Error>;
```
kostko, line 190: Exposing this externally can be quite dangerous. We should maybe consider just getting rid of this and transfer directly to the fee accumulator account (after measuring any potential performance impact, if it's not too big). Can you file an issue?

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -71,13 +95,27 @@ pub enum Error {
...
+    #[sdk_error(code = 8)]
+    InsufficientBalance,
+
+    #[error("invalid denomination")]
+    #[sdk_error(code = 9)]
+    InvalidDenomination,
+
+    #[error("core gas error: {0}")]
+    #[sdk_error(code = 10)]
+    CoreGasError(CoreError),
```
kostko, line 109: Should probably use the transparent modifier as this can be any core error:
```suggestion
    #[error("core: {0}")]
    #[sdk_error(transparent)]
    Core(#[from] CoreError),
```

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -71,13 +95,27 @@ pub enum Error {
...
 }
 
 /// Parameters for the EVM module.
 #[derive(Clone, Default, Debug, cbor::Encode, cbor::Decode)]
 pub struct Parameters {
     /// Minimum acceptable gas price for EVM transactions.
     pub min_gas_price: U256,
+    /// Allowed token denomination for deposits and withdrawals.
```
kostko, line 117:
```suggestion
    /// Token denominatiin used for the EVM token.
```

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -212,71 +358,119 @@ impl Module {
...
+        }
+
         let vicinity = evm_backend::Vicinity {
             gas_price: gas_price_in,
             origin: source,
         };
 
+        // The maximum gas fee has already been withdrawn in authenticate_tx().
+        let paid_fee = ctx.tx_auth_info().fee.amount.amount();
+        let max_gas = gas_price
```
kostko, line 386:
```suggestion
        let max_gas_fee = gas_price
```

runtime-sdk/src/modules/accounts/mod.rs
abukosek: Filed as https://github.com/oasisprotocol/oasis-sdk/issues/433 :)

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -197,104 +327,139 @@ impl API for Module {
...
             origin: source,
         };
 
+        // The maximum gas fee has already been withdrawn in authenticate_tx().
+        let paid_fee = ctx.tx_auth_info().fee.amount.amount();
+        let max_gas_fee = gas_price
+            .checked_mul(primitive_types::U256::from(gas_limit))
+            .ok_or(Error::FeeOverflow)?;
+
+        if paid_fee < max_gas_fee.as_u128() {
```
kostko, line 364: Why is this check needed, briefly it seems like this will never be true as on one hand we know that:
```
paid_fee = ai.fee.amount
max_gas_fee = gas_price * gas_limit
gas_price = ai.fee.amount / ai.fee.gas
gas_limit <= ai.fee.gas
```

And on the other:
```
paid_fee < max_gas_fee
ai.fee.amount < (ai.fee.amount / ai.fee.gas) * gas_limit
1 < gas_limit / ai.fee.gas
ai.fee.gas < gas_limit
```

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -197,104 +327,139 @@ impl API for Module {
...
+        let (exit_reason, exit_value) = f(&mut executor, gas_limit);
+
+        if !exit_reason.is_succeed() {
+            return Err(Error::EVMError(format!("{:?}", exit_reason)));
+        }
 
         let gas_used = executor.used_gas();
 
         if gas_used > gas_limit {
+            core::Module::use_tx_gas(ctx, gas_limit).map_err(Error::Core)?;
```
kostko, line 389: Hm, will this ever be triggered given that you have an `!exit_reason.is_succeed()` check and early return above?

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -197,104 +327,139 @@ impl API for Module {
...
-        // Make sure that gas_price >= min_gas_price.
         let params = Self::params(ctx.runtime_state());
-        let gas_price: primitive_types::U256 = gas_price_in.into();
+        let den = params.token_denomination;
         let min_gas_price: primitive_types::U256 = params.min_gas_price.into();
+
+        let gas_limit: u64 = core::Module::remaining_tx_gas(ctx);
+        let gas_price: primitive_types::U256 = ctx.tx_auth_info().fee.gas_price().into();
+
+        // Make sure that gas_price >= min_gas_price.
```
kostko, line 348: We could handle this at the core module level and optionally enforce a min gas price for all SDK transactions, not only EVM (since now the EVM gas price is basically the same as the SDK gas price)? In that case it could be part of core parameters and be handled in `before_handle_call` of the core module. Thoughts?

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -356,9 +523,75 @@ impl module::MigrationHandler for Module {
...
+        for si in tx.auth_info.signer_info.iter() {
+            let addr = si.address_spec.address();
+            let mut acct: modules::accounts::types::Account =
+                sdk_accounts.get(&addr).unwrap_or_default();
+
+            if acct.nonce != si.nonce {
+                return Err(CoreError::InvalidNonce);
+            }
+
+            acct.nonce += 1;
```
kostko, line 557: So it seems that we are also storing nonces for EVM accounts? Are those ever updated? Should they be synced? Maybe in `do_evm` the nonce of the caller's account should always be set to the nonce of the SDK account before executing the transaction to make sure these are aligned?

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -356,9 +523,75 @@ impl module::MigrationHandler for Module {
...
+        let params = Self::params(ctx.runtime_state());
+        let den = params.token_denomination;
+
+        let evm_acct_addr = Self::derive_caller_from_tx_auth_info(&tx.auth_info)
+            .map_err(|e| CoreError::MalformedTransaction(anyhow::Error::new(e)))?;
+
+        // Check and update nonces on all signer accounts.
+        let mut sdk_accounts_store =
+            storage::PrefixStore::new(ctx.runtime_state(), &modules::accounts::MODULE_NAME);
+        let mut sdk_accounts = storage::TypedStore::new(storage::PrefixStore::new(
```
kostko, line 544: This seems to rely on internal implementation details of the accounts module which is bad. Instead, this should be exposed as an API on the accounts module and called via `Cfg::Accounts`.

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -356,9 +523,75 @@ impl module::MigrationHandler for Module {
...
+        let store = storage::PrefixStore::new(state, &MODULE_NAME);
+        let mut accounts =
+            storage::TypedStore::new(storage::PrefixStore::new(store, &state::ACCOUNTS));
+        let mut account: evm_backend::Account = accounts.get(&evm_acct_addr).unwrap_or_default();
+
+        if account.balance < fee.into() {
+            return Err(CoreError::InsufficientFeeBalance);
+        }
+
+        account.nonce += 1.into();
```
kostko, line 582: What is the point of these separate EVM nonces given that they are not checked and only SDK nonces are checked? It seems like they could get out of sync when some calls are funded from SDK accounts and others from EVM. And then if someone queries the EVM account nonce to generate a transaction, she may get the wrong nonce? Should we make sure they are synced somehow?

kostko (commented):
Withdraw/deposit operations should probably charge some base gas (which should be a parameter similar to `gas_costs` in other modules).

runtime-sdk/modules/evm/src/lib.rs
abukosek: Oops, that's a leftover :)

runtime-sdk/modules/evm/src/lib.rs
abukosek: Probably not, but I'd still like to keep this here just in case (the comparison is cheap).

runtime-sdk/modules/evm/src/lib.rs
abukosek: That's a good idea!  I'll modify the code.

runtime-sdk/modules/evm/src/lib.rs
abukosek: I think these are only needed by the code running inside the EVM (in any case, the `evm` crate requires them).  We probably don't need to sync them.  We don't have a way to directly query the EVM nonce, so it shouldn't matter on our side?

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -116,6 +124,7 @@ pub struct Parameters {
     pub max_tx_signers: u32,
     pub max_multisig_signers: u32,
     pub gas_costs: GasCosts,
+    pub min_gas_price: u128,
```
kostko, line 127: Should this be per denomination? E.g., a `BTreeMap<token::Denomination, u128>`, since you can also specify a denomination while paying for gas. And if a denomination is not listed, then that denomination cannot be used to pay for gas.

abukosek: Good idea, I've changed it into a map.

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -343,6 +356,20 @@ impl module::AuthHandler for Module {
...
+        let fee = ctx.tx_auth_info().fee.clone();
+        let min_gas_price = params.min_gas_price.get(fee.amount.denomination());
+        if min_gas_price.is_none() {
+            return Err(Error::GasPriceTooLow);
+        }
+
+        // Check that the gas price is higher or equal than the set minimum.
+        if &fee.gas_price() < min_gas_price.unwrap() {
+            return Err(Error::GasPriceTooLow);
+        }
```
kostko, line 371: Could change these into a `match` having arms for `None`, `Some(min_gas_price) if &fee.gas_price() < min_gas_price` and `Some(_)`.

kostko (approved):

PR #437, opened 2021-09-14
ts-web/core: types fixes and sync chnages from oasis-core

pro-wh:
Here are the remaining discrepancies between the automatically generated types from #90 and our current type definitions. A this point, I'd like to apply all of them as changes.

- Breaking: NodeAddress is changed to NetTCPAddr, which is consistent with everywhere else where we use the least-derived equivalent type.
- We're synced up with the changes from the oasis-core side. These we don't include in the changelogs. Including but not limited to:
  - beacon consensus module removal
  - runtime client changes
- Slices of special `uint8` are corrected to be `Uint8Array` instead of `number[]`.
- Some mistakes in porting the types are corrected.
- Some mistakes in transferring the doc comments are corrected.
- Uppercase tuple element names from `toarray` structs are now preserved instead of being converted to lowercase.
- PVSS s11n field js doc is removed because it had originated not from proper godoc comments. if anyone would like to fix this in oasis-core, we'll automatically reinstate the docs here too.
- Divergences from alphabetical ordering, unintentional, are corrected.

### DIFF
--- a/client-sdk/ts-web/core/docs/changelog.md
+++ b/client-sdk/ts-web/core/docs/changelog.md
@@ -2,6 +2,10 @@
 
 ## Unreleased changes:
 
+Breaking changes:
+
+- The `NodeAddress` type is replaced with the equivalent `NetTCPAddr`.
+
 Bug fixes:
 
 - Fixes to type declarations that were inconsistent with the Go types.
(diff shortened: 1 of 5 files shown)


pro-wh:
oh this will need the corresponding updates to the client

kostko (approved):
Nice!

PR #438, opened 2021-09-15
contracts-sdk: add env.address_for_instance function

ptrus:
Adds `address_for_instance` function to the wasm contracts environment

### DIFF
--- a/contract-sdk/src/abi/env.rs
+++ b/contract-sdk/src/abi/env.rs
@@ -1,14 +1,24 @@
 //! Environment query ABI.
+use std::convert::TryFrom;
+
+use oasis_contract_sdk_types::address::Address;
+
 use crate::{
     env::Env,
     memory::{HostRegion, HostRegionRef},
-    types::env::{QueryRequest, QueryResponse},
+    types::{
+        env::{QueryRequest, QueryResponse},
+        InstanceId,
+    },
 };
 
 #[link(wasm_import_module = "env")]
 extern "wasm" {
     #[link_name = "query"]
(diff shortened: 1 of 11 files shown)


contract-sdk/src/abi/env.rs
```diff
@@ -28,4 +38,18 @@ impl Env for HostEnv {
     fn query<Q: Into<QueryRequest>>(&self, q: Q) -> QueryResponse {
         query(q.into())
     }
+
+    fn address_for_instance(&self, instance_id: InstanceId) -> Address {
+        // Prepare a region for response.
+        let dst = vec![0; 21];
```
kostko, line 44: Any reason for using vec instead of an array?

runtime-sdk/modules/contracts/src/abi/oasis/env.rs
```diff
@@ -50,6 +54,34 @@ impl<Cfg: Config> OasisV1<Cfg> {
...
+
+                // Charge base gas amount.
+                // TODO: probably separate gas cost.
+                gas::use_gas(ctx.instance, ec.params.gas_costs.wasm_env_query_base)?;
+
+                ctx.instance
+                    .runtime()
+                    .try_with_memory(|mut memory| -> Result<_, wasm3::Trap> {
+                        let instance_id: InstanceId = request.0.into();
+                        let dst: &mut [u8] =
```
kostko, line 73:
```suggestion
                        let dst =
```

kostko (approved):

contract-sdk/src/abi/env.rs
ptrus: no reason, fixed :-)

PR #441, opened 2021-09-16
runtime-sdk: add AuthProof::Module

### LINKED ISSUE #424: Support custom transaction decoding hooks
In order to support things like using raw Ethereum transactions, the ParaTime SDK should be extended with support for custom transaction decoders that are called as lifecycle hooks in the spot where `utx.verify()` is currently called.

The existing `UnverifiedTransaction` structure can be reused for arbitrary blobs, with the second field (list of `AuthProof`s) serving as the discriminator between the different formats. A new `AuthProof::Module(String)` variant would be added that would enable the new decoder hook to determine what format to decode. The name length should be limited to something sensible like 64 bytes and in case this auth proof variant is used, it needs to be the only one in the list.

pro-wh:
fixes #424 

Introducing **module-controlled decoding** (of transactions)

A new special AuthProof::Module(scheme) variant tells the dispatcher to ask the runtime's modules to decode and verify signatures on an incoming UnverifiedTransaction. This allows runtimes to have their own transaction signature scheme.


pro-wh:
oops I still need to write the impl for tuples

pro-wh:
ok there's the tuple implementation. next is to add a simple scheme to test this

kostko (commented):
Don't forget to also update the Go/TS client SDKs. Looks good otherwise (just add some tests).

pro-wh:
Option<Result> is not as good with `?` so I'm going to change it to Result<Option>

pro-wh:
```
{"caller":"simplekvtest.go:212","level":"info","module":"scenario/e2e/test-runtime-simple-keyvalue","msg":"inserting special greeting","ts":"2021-09-16T22:19:03.554655179Z"}
{"caller":"grpc.go:242","err":"client: transaction check failed: runtime error: module: core code: 1 message: malformed transaction: unexpected type","level":"error","method":"/oasis-core.RuntimeClient/SubmitTx","module":"grpc/client","msg":"request failed","req_seq":11,"rsp":null,"ts":"2021-09-16T22:19:03.561184113Z"}
{"caller":"runtime.go:363","err":"client: transaction check failed: runtime error: module: core code: 1 message: malformed transaction: unexpected type","level":"error","module":"scenario/e2e/test-runtime-simple-keyvalue","msg":"test failed","test":"main.SimpleKVTest","ts":"2021-09-16T22:19:03.561264401Z"}
{"caller":"root.go:434","err":"root: failed to run scenario: client: transaction check failed: runtime error: module: core code: 1 message: malformed transaction: unexpected type","level":"error","module":"test-runner","msg":"failed to run scenario","run_id":0,"scenario":"e2e/test-runtime-simple-keyvalue","ts":"2021-09-16T22:19:03.561318693Z"}
Error: root: failed to run scenario: root: failed to run scenario: client: transaction check failed: runtime error: module: core code: 1 message: malformed transaction: unexpected type
```

there's something I've done wrong here :thinking:

pro-wh:
this custom format in the test is going to need a nonce after all

client-sdk/go/types/transaction.go
```diff
@@ -22,6 +22,8 @@ type AuthProof struct {
 	Signature []byte `json:"signature,omitempty"`
 	// Multisig is for multisig authentication.
 	Multisig [][]byte `json:"multisig,omitempty"`
+	// Module is a flag to use module-controled decoding.
+	Module string `json:"module,omitempty"`
```
kostko, line 26: I guess this doesn't differentiate between a module being set to an empty string and the module not even being there? For example if you use an empty string then the check below will pass. You could maybe use a `*string` instead and check that it is `nil`.

tests/e2e/simplekvtest.go
```diff
@@ -147,6 +156,27 @@ func kvGet(rtc client.RuntimeClient, key []byte) ([]byte, error) {
...
+	if err != nil {
+		return fmt.Errorf("signing special greeting: %w", err)
+	}
+	utx := types.UnverifiedTransaction{
+		Body: cbor.Marshal(kvSpecialGreeting{
+			Greeting:  greeting,
+			From:      signer.Public().(ed25519.PublicKey),
+			Signature: sig,
+		}),
+		AuthProofs: nil,
```
kostko, line 171: Shouldn't this configure the module auth proof?

wh0: oh I must not have pushed that fix

pro-wh: yeah. I fixed that earlier. must not have pushed it. (wrong account, reposting this)

pro-wh:
oh yes, the other thing about publickey is still not pushed

pro-wh:
image

yeah see I "decode" it into a plain Transaction object but I didn't include a nonce in the test format so I just set some number, and the accounts module won't authenticate it. so going forward I'll just use a darned cbor-encoded thing that has a spot for the nonce

client-sdk/go/types/transaction.go
pro-wh: the empty string case! :laughing: 

Go is weird in this, but it ends up not being that bad. we can require that the encoding scheme name be non-empty. Similarly, that `Signature []byte` up there requires that the signature not be zero length, which would be encoded as nil.

pro-wh:
image
 noooooo

pro-wh:
oh because there are only e2e tests

client-sdk/go/types/transaction.go
pro-wh: ok I tried changing it to *string in a separate commit. although now you can't build AuthProofs in code as easily, with go not permitting `&"sceme-name"` address-of-constant antics

kostko: Oh I see, yeah i would be fine also with changing it back and not allowing empty strings as long as it is consistent with what is done in the runtime.

pro-wh: ok. adding a note about the non-empty requirement

PR #442, opened 2021-09-16
runtime-sdk/modules/evm: Use HashedStore for storage

### LINKED ISSUE #376: Use HashedStore for EVM contract storage
EVM contract storage should use `HashedStore` introduced in #263 (should be trivial to plug in after that is merged).

abukosek:
Closes #376.

### DIFF
--- a/runtime-sdk/modules/evm/src/evm_backend.rs
+++ b/runtime-sdk/modules/evm/src/evm_backend.rs
@@ -135,7 +135,9 @@ impl<'c, C: oasis_runtime_sdk::Context> EVMBackend for Backend<'c, C> {
 
         let store = storage::PrefixStore::new(state, &crate::MODULE_NAME);
         let storages = storage::PrefixStore::new(store, &state::STORAGES);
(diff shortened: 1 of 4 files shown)


runtime-sdk/modules/evm/src/evm_backend.rs
```diff
@@ -117,7 +117,9 @@ impl<'c, C: oasis_runtime_sdk::Context> EVMBackend for Backend<'c, C> {
         let state = ctx.runtime_state();
 
         let store = storage::PrefixStore::new(state, &crate::MODULE_NAME);
-        let codes = storage::TypedStore::new(storage::PrefixStore::new(store, &state::CODES));
+        let codes = storage::TypedStore::new(storage::HashedStore::<_, blake3::Hasher>::new(
```
kostko, line 120: I don't think this is needed here as the key is always an address which is already hashed?

abukosek: Fixed, thanks!

kostko (approved):

PR #452, opened 2021-09-22
ethereum transaction decoding

pro-wh:
This introduces a module-controlled transaction decoding scheme, `evm.ethereum.v0`. this parses and verifies an Ethereum "sendRawTransaction" compatible transaction.

For implementation, we're using some structs from the `ethereum` crate, which seems to be related to our choice of evm implementation. But that crate was buggy, and several things we ended up having to do ourselves.

### DIFF
--- a/runtime-sdk/modules/evm/src/lib.rs
+++ b/runtime-sdk/modules/evm/src/lib.rs
@@ -2,6 +2,7 @@
 pub mod derive_caller;
 pub mod evm_backend;
 pub mod precompile;
+pub mod raw_tx;
 pub mod types;
 
 use std::collections::BTreeMap;
@@ -531,6 +532,19 @@ impl<Cfg: Config> module::MigrationHandler for Module<Cfg> {
 }
 
 impl<Cfg: Config> module::AuthHandler for Module<Cfg> {
+    fn decode_tx<C: Context>(
+        _ctx: &mut C,
+        scheme: &str,
+        body: &[u8],
+    ) -> Result<Option<Transaction>, CoreError> {
+        match scheme {
+            "evm.ethereum.v0" => Ok(Some(
+                raw_tx::decode(body).map_err(CoreError::MalformedTransaction)?,
+            )),
+            _ => Ok(None),
+        }
+    }
+
     fn authenticate_tx<C: Context>(ctx: &mut C, tx: &Transaction) -> Result<(), CoreError> {
         // We're only interested in transactions that can be paid with tokens
         // from the corresponding EVM account.
--- a/runtime-sdk/modules/evm/src/lib.rs
+++ b/runtime-sdk/modules/evm/src/lib.rs
@@ -56,7 +56,8 @@ pub trait Config: 'static {
     /// Module that is used for accessing accounts.
     type Accounts: modules::accounts::API;
 
-    /// The chain ID to supply when a contract requests it.
+    /// The chain ID to supply when a contract requests it. Ethereum-format transactions must use
+    /// this chain ID.
     const CHAIN_ID: u64;
 }
 
@@ -539,7 +540,8 @@ impl<Cfg: Config> module::AuthHandler for Module<Cfg> {
     ) -> Result<Option<Transaction>, CoreError> {
         match scheme {
             "evm.ethereum.v0" => Ok(Some(
-                raw_tx::decode(body).map_err(CoreError::MalformedTransaction)?,
+                raw_tx::decode(body, Some(Cfg::CHAIN_ID))
+                    .map_err(CoreError::MalformedTransaction)?,
             )),
             _ => Ok(None),
         }
--- a/runtime-sdk/modules/evm/src/raw_tx.rs
+++ b/runtime-sdk/modules/evm/src/raw_tx.rs
@@ -1,0 +1,252 @@
+use std::convert::TryInto;
+
+use anyhow::{anyhow, Context as _};
+use ethereum;
+use k256;
+
+use oasis_runtime_sdk::{
+    crypto::signature,
+    types::{token, transaction},
+};
+
+use crate::types;
+
+pub fn decode(body: &[u8]) -> Result<transaction::Transaction, anyhow::Error> {
+    let (sig, sig_hash, eth_action, eth_value, eth_input, eth_nonce, eth_gas_price, eth_gas_limit) =
+        match rlp::decode::<ethereum::TransactionV2>(body)
+            .with_context(|| "decoding transaction rlp")?
+        {
+            ethereum::TransactionV2::Legacy(eth_tx) => {
+                let sig = k256::ecdsa::recoverable::Signature::new(
+                    &k256::ecdsa::Signature::from_scalars(
+                        eth_tx.signature.r().to_fixed_bytes(),
+                        eth_tx.signature.s().to_fixed_bytes(),
+                    )
+                    .with_context(|| "signature from_scalars")?,
+                    k256::ecdsa::recoverable::Id::new(eth_tx.signature.standard_v())
+                        .with_context(|| "recoverable id new")?,
+                )
+                .with_context(|| "recoverable signature new")?;
+                let message = ethereum::LegacyTransactionMessage::from(eth_tx);
+                let sig_hash = message.hash();
+                (
+                    sig,
+                    sig_hash,
+                    message.action,
+                    message.value,
+                    message.input,
+                    message.nonce,
+                    message.gas_price,
+                    message.gas_limit,
+                )
+            }
+            ethereum::TransactionV2::EIP2930(eth_tx) => {
+                let sig = k256::ecdsa::recoverable::Signature::new(
+                    &k256::ecdsa::Signature::from_scalars(
+                        eth_tx.r.to_fixed_bytes(),
+                        eth_tx.s.to_fixed_bytes(),
+                    )
+                    .with_context(|| "signature from_scalars")?,
+                    k256::ecdsa::recoverable::Id::new(eth_tx.odd_y_parity.into())
+                        .with_context(|| "recoverable id new")?,
+                )
+                .with_context(|| "recoverable signature new")?;
+                let message = ethereum::EIP2930TransactionMessage::from(eth_tx);
+                let sig_hash = message.hash();
+                (
+                    sig,
+                    sig_hash,
+                    message.action,
+                    message.value,
+                    message.input,
+                    message.nonce,
+                    message.gas_price,
+                    message.gas_limit,
+                )
+            }
+            ethereum::TransactionV2::EIP1559(eth_tx) => {
+                let sig = k256::ecdsa::recoverable::Signature::new(
+                    &k256::ecdsa::Signature::from_scalars(
+                        eth_tx.r.to_fixed_bytes(),
+                        eth_tx.s.to_fixed_bytes(),
+                    )
+                    .with_context(|| "signature from_scalars")?,
+                    k256::ecdsa::recoverable::Id::new(eth_tx.odd_y_parity.into())
+                        .with_context(|| "recoverable id new")?,
+                )
+                .with_context(|| "recoverable signature new")?;
+                let message = ethereum::EIP1559TransactionMessage::from(eth_tx);
+                let sig_hash = message.hash();
+                // Base fee is zero. Allocate only priority fee.
+                let resolved_gas_price =
+                    std::cmp::min(message.max_fee_per_gas, message.max_priority_fee_per_gas);
+                (
+                    sig,
+                    sig_hash,
+                    message.action,
+                    message.value,
+                    message.input,
+                    message.nonce,
+                    resolved_gas_price,
+                    message.gas_limit,
+                )
+            }
+        };
+    let (method, body) = match eth_action {
+        ethereum::TransactionAction::Call(eth_address) => (
+            "evm.Call",
+            cbor::to_value(types::Call {
+                address: eth_address.into(),
+                value: eth_value.into(),
+                data: eth_input,
+            }),
+        ),
+        ethereum::TransactionAction::Create => (
+            "evm.Create",
+            cbor::to_value(types::Create {
+                value: eth_value.into(),
+                init_code: eth_input,
+            }),
+        ),
+    };
+    let key = sig
+        .recover_verify_key_from_digest_bytes(sig_hash.as_bytes().into())
+        .with_context(|| "recover verify key from digest")?;
+    let nonce: u64 = eth_nonce
+        .try_into()
+        .map_err(|e| anyhow!("converting nonce: {}", e))?;
+    let gas_price: u128 = eth_gas_price
+        .try_into()
+        .map_err(|e| anyhow!("converting gas price: {}", e))?;
+    let gas_limit: u64 = eth_gas_limit
+        .try_into()
+        .map_err(|e| anyhow!("converting gas limit: {}", e))?;
+    let resolved_fee_amount = gas_price
+        .checked_mul(gas_limit as u128)
+        .ok_or_else(|| anyhow!("computing total fee amount"))?;
+    Ok(transaction::Transaction {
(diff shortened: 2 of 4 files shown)


pro-wh:
by the way, is `AddressSpec::Signature(PublicKey::Secp256k1)` the right thing to use here? or will we need something lower level due to some hashing difference that I vaguely remember hearing about?

kostko:
> by the way, is AddressSpec::Signature(PublicKey::Secp256k1) the right thing to use here

Yes this is ok, the EVM module will use the right derivation scheme for EVM account addresses.

pro-wh:
```
error[E0277]: the trait bound `str: std::error::Error` is not satisfied
  --> runtime-sdk/modules/evm/src/raw_tx.rs:70:10
   |
70 |         .with_context(|| "converting nonce")
   |          ^^^^^^^^^^^^ the trait `std::error::Error` is not implemented for `str`
   |
   = note: required because of the requirements on the impl of `std::error::Error` for `&str`
   = note: required because of the requirements on the impl of `anyhow::context::ext::StdError` for `&str`
   = note: required because of the requirements on the impl of `anyhow::context::private::Sealed` for `Result<u64, &str>`
```

what even is the connection with std::error::Error here? &str is just the context type, and we're creating an anyhow::Error

kostko:
Maybe the error type of the `try_into()` result is a `&str` (e.g. the `TryInto::Error` associated type) instead of a proper error type so `with_context` is only implemented for types that implement `std::error::Error`? Not sure what those types are, probably those U256 types? Just checked and [this is probably the reason](https://github.com/paritytech/parity-common/blob/master/uint/src/uint.rs#L197).

pro-wh:
oh cool, then we can throw more map_err at it :laughing:

kostko:
Looked into this a bit, here are some findings:
* Our address derivation scheme is incorrect as it needs to operate on the uncompressed point without the initial tag. That makes it derive addresses that agree with go-ethereum's raw address derivation. This needs to be fixed in `derive_caller_from_tx_auth_info` (which should probably be moved into its own module with some added tests).
* It also looks like the message hash is incorrectly computed due to two things:
  * The hashing scheme implemented in the Rust crate does not prepend a type prefix (`0x02`).
  * The RLP encoding in the Rust crate is broken, the serialization differs in the third byte for some reason.

runtime-sdk/modules/evm/src/raw_tx.rs
```diff
@@ -0,0 +1,176 @@
...
+pub fn decode(body: &[u8]) -> Result<transaction::Transaction, anyhow::Error> {
+    // TODO: verify signature
+    let (sig, sig_hash, eth_action, eth_value, eth_input, eth_nonce, eth_gas_price, eth_gas_limit) =
+        match rlp::decode::<ethereum::TransactionV2>(body)
+            .with_context(|| "decoding transaction rlp")?
+        {
+            ethereum::TransactionV2::Legacy(eth_tx) => {
+                let sig = k256::ecdsa::recoverable::Signature::new(
+                    &k256::ecdsa::Signature::from_scalars(
+                        eth_tx.signature.r().as_bytes(),
```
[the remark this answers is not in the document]
pro-wh: huh, I had written this back when it wasn't compiling but it was never posted

pro-wh:
>     * maybe there's something wrong with the recovery in the first place

>     * The hashing scheme implemented in the Rust crate does not prepend a type prefix (`0x02`).

there was. the message hash was wrong, with the transaction type byte left out. I've written some additional routines to do this correctly.

>     * it needs to operate on the uncompressed point without the initial tag

thankfully that was it. there was a `to_untagged_bytes` method to get it in this format

>     * The RLP encoding in the Rust crate is broken, the serialization differs in the third byte for some reason.

the rlp itself was working, but the transaction types library's use of it was incorrect. I've written a corrected

pro-wh:
oh and the changes to account derivation broke several other tests and I had to work on something else today instead of fixing those

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -429,18 +430,26 @@ impl<Cfg: Config> Module<Cfg> {
...
+            PublicKey::Ed25519(_) => Ok(Self::derive_caller_from_bytes(
                 &ai.signer_info[0].address_spec.address().as_ref()[1..],
             )),
             _ => Err(Error::InvalidSignerType),
         }
     }
 
+    fn derive_caller_from_tx_auth_info(ai: &AuthInfo) -> Result<H160, Error> {
+        match &ai.signer_info[0].address_spec {
+            AddressSpec::Signature(pk) => Self::derive_caller_from_public_key(pk),
```
kostko, line 448: For all non-secp256k1 specs the caller should be derived from the address regardless of what scheme it is using.

runtime-sdk/modules/evm/src/raw_tx.rs
```diff
@@ -0,0 +1,274 @@
...
+        ethereum::TransactionAction::Create => (
+            "evm.Create",
+            cbor::to_value(types::Create {
+                value: eth_value.into(),
+                init_code: eth_input,
+            }),
+        ),
+    };
+    let key = sig
+        .recover_verify_key_from_digest(sig_digest)
```
kostko, line 164: This doesn't actually verify the signature so signature verification needs to be performed separately.

runtime-sdk/modules/evm/src/raw_tx.rs
```diff
@@ -0,0 +1,274 @@
...
+    use oasis_runtime_sdk::types::token;
+
+    use crate::types;
+
+    use super::decode;
+
+    #[test]
+    fn test_decode() {
+        // Some recent transaction seen on Ethereum when we wrote this test.
+        // https://etherscan.io/tx/0x5e5cb07e74593198f66604b0dc4dae6187186c3fa94088f062cc51fc554141de
```
kostko, line 217: Also add tests for the other transaction types, so everything is covered.

runtime-sdk/src/crypto/signature/secp256k1.rs
```diff
@@ -17,6 +17,11 @@ impl PublicKey {
         self.0.as_bytes()
     }
 
+    /// Return an alternative byte representation used in deriving Ethereum-compatible addresses.
+    pub fn uncompressed_untagged_bytes(&self) -> Option<Vec<u8>> {
```
kostko, line 21:
```suggestion
    pub fn to_uncompressed_untagged_bytes(&self) -> Option<Vec<u8>> {
```

kostko (commented):
We probably need to change how we define chain ID as we use the full 256-bit identifier (derived from the consensus layer genesis and runtime id, see `evm_backend.rs`) but this is not supported everywhere (e.g. in transactions it is only an u64). So we could either make it a parameter (could even be an associated const in the `Config` trait as it is not expected to change) or truncate the derived chain ID.

Then we need to verify that the specified chain ID matches before signature verification.

kostko:
> oh also this uses the sha3 crate, to match what the evm and ethereum use. although I see we also use tiny-keccak in other places \:

Yeah using `sha3` is fine (main also uses it now with the precompiles PR getting merged). I would be in favor of replacing the use of `tiny_keccak` in the `evm` crate with `sha3` so we only use a single dependency.

runtime-sdk/modules/evm/src/raw_tx.rs
pro-wh: I've asked about this, and I've read some vaguely related things to the contrary. can you send a citation for this?

runtime-sdk/modules/evm/src/lib.rs
pro-wh: that's a divergence from what this previously did, but we can go ahead and do that as part of this PR

runtime-sdk/modules/evm/src/raw_tx.rs
pro-wh: ya should

runtime-sdk/src/crypto/signature/secp256k1.rs
pro-wh: ok

runtime-sdk/modules/evm/src/raw_tx.rs
pro-wh: nvm let's track that here

runtime-sdk/modules/evm/src/raw_tx.rs
kostko: Looking at the function source I don't see a verification step:

https://docs.rs/k256/0.9.6/src/k256/ecdsa/recoverable.rs.html#190

runtime-sdk/modules/evm/src/test.rs
```diff
@@ -47,8 +47,8 @@ fn load_erc20() -> Vec<u8> {
 #[test]
 fn test_evm_caller_addr_derivation() {
     let expected =
-        H160::from_slice(&Vec::<u8>::from_hex("89519d9720bbedf870ab6ae6fbd4bb8af92f4328").unwrap());
-    let derived = EVM::derive_caller_from_bytes(keys::dave::pk().as_bytes());
+        H160::from_slice(&Vec::<u8>::from_hex("dce075e1c39b1ae0b75d554558b6451a226ffe00").unwrap());
```
pro-wh, line 50: I've pasted the result over the old expected value

runtime-sdk/modules/evm/src/lib.rs
pro-wh: added as a commit near the beginning

runtime-sdk/modules/evm/src/raw_tx.rs
pro-wh: why does that have a question mark on it >_>

pro-wh: mathematically though, does the recover operation sometimes give public keys for which the given signature is not valid?

pro-wh:
I've rebased, and the rust tests are run, but codecov won't show the coverage info

pro-wh:
> Our address derivation scheme is incorrect as it needs to operate on the uncompressed point without the initial tag. That makes it derive addresses that agree with go-ethereum's raw address derivation. This needs to be fixed in `derive_caller_from_tx_auth_info` (which should probably be moved into its own module with some added tests).

moved to a module

pro-wh:
can someone provide a sample EIP-2930 transaction?

runtime-sdk/modules/evm/src/raw_tx.rs
kostko: Hm, so in case the signature doesn't verify it would return a bogus key, but it would still continue? Because looking at that function it doesn't seem like it would fail in that case? E.g. the only failure that I can see is if decompression of `r` fails. Can you check what happens given a bogus signature?

pro-wh: ok yeah

pro-wh: yup, recovers to a different address

pro-wh: I had to add a check for high `s` values ):

runtime-sdk/modules/evm/src/raw_tx.rs
pro-wh: added!

runtime-sdk/modules/evm/src/raw_tx.rs
pro-wh: tests added

runtime-sdk/modules/evm/src/test.rs
pro-wh: I added some test vectors from ethereum so we should be fine

pro-wh:
reworked this to use some vectors from ethereum tests

pro-wh:
added chain id check

pro-wh:
splitting prereq work into other PRs, with this one running so long

pro-wh:
aw now it won't run the tests

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -54,6 +55,9 @@ pub mod state {
 pub trait Config: 'static {
     /// Module that is used for accessing accounts.
     type Accounts: modules::accounts::API;
+
+    /// Ethereum-format transactions must use this chain ID.
+    const EXPECTED_CHAIN_ID: Option<u64>;
```
kostko, line 60: I would change this to just `CHAIN_ID`, make it non optional and update the evm backend to also return this when querying for `chain_id` instead of using that derivation scheme.

runtime-sdk/modules/evm/src/raw_tx.rs
```diff
@@ -0,0 +1,411 @@
...
+            )
+            .with_context(|| "recoverable signature new")?;
+            let message = ethereum::LegacyTransactionMessage::from(eth_tx);
+
+            // TODO: replace this with message.hash()
+            let mut sig_digest = sha3::Keccak256::new();
+            sig_digest.update(&rlp::encode(&message));
+
+            (
+                message.chain_id,
```
kostko, line 61: Oh so at some point there was no domain separation? We should just reject these cases without chain ID.

runtime-sdk/modules/evm/src/lib.rs
wh0: the rename is fine. as you observe below, none refers to legacy transactions with no chain separation, not "accept anything"

kostko: I don't think we want to support transactions without chain separation though? In any case the same `CHAIN_ID` should also be used in the `evm_backend.rs` for chain ID queries from EVM.

pro-wh: oh ok, that'll make things easier for the other places that'll use it

runtime-sdk/modules/evm/src/raw_tx.rs
pro-wh: can't, we use a bunch of sigs without chain id in the test suite

pro-wh:
gonna set the base branch to main so that we have CI. but y'all know that the earlier commits are from other PRs

runtime-sdk/modules/evm/src/lib.rs
pro-wh: changed

runtime-sdk/modules/evm/src/raw_tx.rs
kostko: Sounds good.

runtime-sdk/modules/evm/src/lib.rs
kostko: Also don't forget to update the chain ID query in the EVM backend.

pro-wh: oh yeah, that always sneaks around these refactors :laughing:

kostko (commented):
Can you also update this to use the new `CHAIN_ID` constant?
https://github.com/oasisprotocol/oasis-sdk/blob/a7abff8d43ac96c06345a939bf38b15816af7dde/runtime-sdk/modules/evm/src/evm_backend.rs#L87

kostko:
And just remove that `EVM_CHAIN_CONTEXT`. Since it seemed like it was an `U256` I figured it would be best to use a hash there, but it seems that in reality it is not an `U256` :-)

pro-wh:
> Can you also update this to use the new `CHAIN_ID` constant?
> 
> https://github.com/oasisprotocol/oasis-sdk/blob/a7abff8d43ac96c06345a939bf38b15816af7dde/runtime-sdk/modules/evm/src/evm_backend.rs#L87

was this previously using `b"oasis-runtime-sdk/evm: chain id "` constantly?

kostko:
Oh lol, this was missing a hashing step :facepalm:

kostko (approved):

PR #455, opened 2021-09-22
runtime-sdk: Bump Oasis Core to rev 6947544

kostko:


ptrus:
Can we wait to include https://github.com/oasisprotocol/oasis-core/pull/4278 as well please. To avoid bumping twice :-)

kostko:
Done.

ptrus (approved):
thanks

PR #458, opened 2021-09-23
runtime-sdk/modules/contracts: Cache and compress contract code

kostko:


runtime-sdk/modules/contracts/src/abi/oasis/test.rs
```diff
@@ -276,7 +276,10 @@ fn test_hello_contract_out_of_gas() {
 
     assert_eq!(result.module_name(), "core");
     assert_eq!(result.code(), 12);
-    assert_eq!(&result.to_string(), "core: out of gas");
+    assert_eq!(
+        &result.to_string(),
+        "core: out of gas (limit: 1000 wanted: 1000)"
```
ptrus, line 281: shouldn't `wanted` be > `limit` in all these cases where it fails with out of gas.

kostko: Yeah we don't actually know how much was wanted as the gas metering logic aborts before it would add the extra gas. But we could change it slightly to be able to recover this information.

ptrus: For this specific case, or in all cases? If for all cases then reporting both the "limit" and "wanted" in the error message is misleading/not really useful and just including the "limit" would be enough?

kostko: The error is the same for all modules, for the SDK we actually report this correctly, just WASM didn't report it correctly as there was not enough information available. I have it working locally, will push in a bit.

kostko: Check it out, it's in the last commit ;-)

ptrus: great, thanks :grin:

ptrus (approved):

PR #460, opened 2021-09-23
ci: check for consistent oasis-core version

pro-wh:
from a doc block in an added script:

> We have some various Go programs in multiple places. That results in
multiple specifications of the oasis-core dependency. This script checks
that they all use the same version.

preview:

image


kostko (approved):
Yeah good idea, it seems I forgot about the reflect one!

pro-wh:
it's buried so deep :cry:

PR #470, opened 2021-09-27
tests: fix syntax difference with single-dependency modules

pro-wh:
modules that only have one dependency get rewritten without the parentheses, so we can't look for a verbatim line match

I guess we'll start writing our own parser then :roll_eyes:

### DIFF
--- a/tests/check-core-versions.sh
+++ b/tests/check-core-versions.sh
@@ -4,9 +4,21 @@
 # multiple specifications of the oasis-core dependency. This script checks
 # that they all use the same version.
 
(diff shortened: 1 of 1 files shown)


kostko:
You can convert back to draft.

kostko:
It's slightly hidden though, do you not see this:
image

pro-wh:
oh thanks, I hadn't noticed that

pro-wh:
ok I've reimplemented this in awk

kostko (approved):

PR #489, opened 2021-10-01
contract-sdk: updates / improvements

ptrus:
-  add support for passing optional data to contract `reply_handlers`
- extract OAS20 methods logic to `types` crate helpers functions, so that these can be reused in contracts extending the OAS20-base (e.g.  creating a wrapped-oas20 token contract)
- add E2E test to `hello` contract for initializing a OAS20 token via the `hello` contract

### DIFF
--- a/contract-sdk/src/testing.rs
+++ b/contract-sdk/src/testing.rs
@@ -56,8 +56,15 @@ impl MockEnv {
 }
 
 impl Env for MockEnv {
-    fn query<Q: Into<QueryRequest>>(&self, _query: Q) -> QueryResponse {
-        unimplemented!()
+    fn query<Q: Into<QueryRequest>>(&self, query: Q) -> QueryResponse {
+        match query.into() {
+            QueryRequest::BlockInfo => QueryResponse::BlockInfo {
+                round: 42,
+                epoch: 2,
+                timestamp: 100_000,
+            },
+            _ => unimplemented!(),
+        }
     }
 
     fn address_for_instance(&self, instance_id: InstanceId) -> Address {
@@ -74,6 +81,8 @@ impl Env for MockEnv {
 pub struct MockContext {
     /// Execution context.
     pub ec: ExecutionContext,
+
+    pub deposited_tokens: Vec<token::BaseUnits>,
 
     /// Public store.
     pub public_store: MockStore,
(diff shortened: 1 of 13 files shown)


contract-sdk/types/src/message.rs
```diff
@@ -65,6 +71,13 @@ impl CallResult {
     }
 }
 
+/// Instantiate call result.
+#[derive(Clone, Debug, cbor::Encode, cbor::Decode)]
+pub struct InstantiateResult {
+    /// Assigned instance identifier.
+    pub id: InstanceId,
```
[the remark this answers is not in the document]
kostko: Oh this is the same thing, sorry nevermind.

contract-sdk/types/src/message.rs
```diff
@@ -65,6 +71,13 @@ impl CallResult {
     }
 }
 
+/// Instantiate call result.
+#[derive(Clone, Debug, cbor::Encode, cbor::Decode)]
+pub struct InstantiateResult {
```
kostko, line 76: So this is specific to the `contracts.Instantiate` module call. Maybe these should go into their own module instead of being part of `message`? It would make sense to have things namespaced by module name, e.g. `crate::modules::contracts::*`. And then in the documentation for the `modules` module explain that these are the types for easier calling into existing SDK modules.

contract-sdk/src/testing.rs
```diff
@@ -119,7 +129,7 @@ impl Context for MockContext {
     }
 
     fn deposited_tokens(&self) -> &[token::BaseUnits] {
-        &self.ec.deposited_tokens
```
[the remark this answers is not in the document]
ptrus: yeah true, this is not needed :grin:

contract-sdk/types/src/message.rs
ptrus: Done moved this type into `contract-sdk/modules/contracts.rs`

contract-sdk/src/lib.rs
```diff
@@ -9,6 +9,7 @@ pub mod env;
 pub mod error;
 pub mod event;
 pub mod memory;
+pub mod modules;
```
kostko, line 12: Oh I thought this would be added to `contract-sdk-types`?

kostko (approved):

contract-sdk/src/lib.rs
ptrus: Yeah of course that makes more sense, thanks :-)

PR #491, opened 2021-10-02
modules/evm: ethereum-compatible secp256k1 addresses

pro-wh:
hash it the way ethereum does, so we can get an ethereum address from a secp256k1 key

wanted in #452


kostko (approved):

pro-wh:
oh I need to fix up the test not to use the derive_caller from the next pr

PR #495, opened 2021-10-04
contract-sdk: add ecdsa_recover crypto helper

ptrus:
Adds `ecdsa_recover` helper to the contracts-sdk module which recovers the public key from a secp256k1 signature.

TODO:
- [ ] reuse code between the mock and non-mock versions (extract recover code to -types crate)
- [ ] add integration test to the hello wasm contract


contract-sdk/src/abi/env.rs
```diff
@@ -21,6 +21,12 @@ extern "wasm" {
     fn env_address_for_instance(instance_id: u64, dst_ptr: u32, dst_len: u32);
 }
 
+#[link(wasm_import_module = "crypto")]
```
kostko, line 24: Maybe move to `crypto.rs`?

contract-sdk/src/abi/env.rs
```diff
@@ -21,6 +21,12 @@ extern "wasm" {
...
+extern "wasm" {
+    #[link_name = "ecdsa_recover"]
+    fn env_ecdsa_recover(input_ptr: u32, input_len: u32) -> HostRegion;
```
kostko, line 27:
```suggestion
    fn crypto_ecdsa_recover(input_ptr: u32, input_len: u32) -> HostRegion;
```

pro-wh (commented):
`v` is in `{0,1}`? should document this, since ethereum uses different

contract-sdk/types/src/crypto.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    let mut msg = [0u8; 32];
+    let mut sig = [0u8; 64];
+    let v = input[96];
+
+    msg[0..32].copy_from_slice(&input[0..32]);
+    sig[0..].copy_from_slice(&input[32..96]);
+
+    let signature = Signature::try_from(&sig[..]).map_err(|_| Error::MalformedSignature)?;
```
pro-wh, line 34: there's also a `from_scalars` if you'd prefer to have the r and s explicit

pro-wh, line 34: are various other things like range checks on r and s up to the caller?

ptrus:
> `v` is in `{0,1}`? Hm yeah i see that ` should document this, since ethereum uses different

Hm yeah i see that `k256::recovery_id` only handles `0` and `1` cases. Should probably keep it the way ethereum does is. Do you have a ref?

contract-sdk/types/src/crypto.rs
ptrus: Does solidity's  `ecrecover` do it? If so, then yes we should probably do it here as well. I think this would ideally be in sync with that, since in most cases this method will be used due to interoperability with ethereum.

Also what are "other things"?


---

Also should probably keep in sync with the EVM's module precomiple: https://github.com/oasisprotocol/oasis-sdk/blob/a7abff8d43ac96c06345a939bf38b15816af7dde/runtime-sdk/modules/evm/src/precompile.rs#L40-L95

kostko:
AFAIK even Ethereum only supports a single bit, but for some stupid reason in the past they used some weird encoding where 0 encoded as 27 or something like that.

pro-wh:
https://ethereum.github.io/yellowpaper/paper.pdf

see appendix F in their yellowpaper. the text mentions "not only with a recovery identifier but with some other numbers," but I vaguely think that they're not supported by the actual instruction.

contract-sdk/types/src/crypto.rs
pro-wh: ok from my reading of the yellowpaper, at least the raw instruction does not. it says the ECRECOVER instruction calls ECDSARECOVER, and the check for canonical `s` is stated outside of that in appendix F.

'other things' I think are just these bounds checks

image

the `< secp256k1n` are handled by Signature::try_from I think

ptrus: Thanks, switched to using `from_scalars`

ptrus: Thanks, added the `s.is_high()` check i think it makes sense having it.  Should we update the evm precompile as well?

kostko: Yes I believe it makes sense to have it in the precompile as well, since we do the same check in Ethereum transaction signature verification.

contract-sdk/types/src/crypto.rs
```diff
@@ -0,0 +1,77 @@
...
+    #[error("malformed input")]
+    MalformedInput,
+    #[error("malformed signature")]
+    MalformedSignature,
+    #[error("signature verification failed")]
+    VerificationFailed,
+}
+
+/// ECDSA public key recovery method.
+pub fn ecdsa_recover(input: &[u8]) -> Result<Vec<u8>, Error> {
```
kostko, line 19: Any reason why this is part of `contract-sdk-types` crate, for tests? This pulls in the `k256` dependency into the WASM environment. Could gate it under an optional `crypto` feature and/or move it into a separate `contract-sdk-crypto` crate.

contract-sdk/types/src/crypto.rs
```diff
@@ -0,0 +1,77 @@
...
+    if signature.s().is_high().into() {
+        return Err(Error::MalformedSignature);
+    }
+
+    match signature.recover_verify_key_from_digest_bytes(&msg.into()) {
+        Ok(recovered_key) => Ok(EncodedPoint::from(&recovered_key)
+            .decompress()
+            .unwrap()
+            .as_bytes()
+            .to_vec()),
```
kostko, line 50: I believe the return type could be fixed size, right? This way we could avoid the need for allocations from the host as the caller could just preallocate an empty array and pass its address. I guess in this case we would also need something to signal an error.

contract-sdk/types/src/crypto.rs
ptrus: Yes, it's so that we can reuse the code in the `MockEnv` for tests. We'll probably want to do this for all crypto imports, so maybe having it in a separate `contract-sdk-crypto` crate makes snse.

contract-sdk/types/src/crypto.rs
ptrus: Yeah it could be done that way. So we could return a zero array in case something failed (that's what i currently do anyway in the "link function"), with a todo to propagate the error :-) - not sure if having the concrete error in the contract is all that useful anyway.

contract-sdk/types/src/crypto.rs
ptrus: done

contract-sdk/types/src/crypto.rs
ptrus: Only thing the error would really tell you if the signature is invalid, or if the whole input is malformed. I see that returning the zero array in case the verification failed is also what we do in the evm precompile.

contract-sdk/Cargo.toml
```diff
@@ -10,12 +10,16 @@ license = "Apache-2.0"
 cbor = { version = "0.2.1", package = "oasis-cbor" }
 oasis-contract-sdk-types = { path = "types" }
 oasis-contract-sdk-macros = { path = "../contract-sdk-macros", optional = true }
+oasis-contract-sdk-crypto = { path = "crypto" }
```
kostko, line 13: Should this only be a dependency on non-wasm32 since it is only used in tests? E.g.
```toml
[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
oasis-contract-sdk-crypto = { path = "crypto" }
```

contract-sdk/crypto/src/ecdsa.rs
```diff
@@ -0,0 +1,77 @@
...
+
+#[cfg(test)]
+mod test {
+    use hex;
+
+    use super::*;
+
+    #[test]
+    fn test_ecdsa_recover() {
+        // TODO: add some more test vectors.
```
kostko, line 63: Is the plan to add them in this PR?

contract-sdk/types/src/crypto.rs
kostko: Yeah this makes sense.

contract-sdk/crypto/src/ecdsa.rs
ptrus: Added

contract-sdk/types/src/crypto.rs
ptrus: Done: https://github.com/oasisprotocol/oasis-sdk/pull/505

kostko (approved):

PR #498, opened 2021-10-04
modules/evm: switch to sha3 crate for keccak

### LINKED ISSUE #497: modules/evm: migrate tiny_keccak -> sha3
we had been using tiny_keccak, but started using sha3's keccak for interop with some ecdsa libraries that expect a proper Digest trait implementation. so let's switch the rest over

it looks like we had originally started using tiny_keccak in the keymanager sdk for its `TupleHash` implementation. I guess let's keep using tiny_keccak there for now

pro-wh:
fixes #497

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1997,7 +1997,6 @@ dependencies = [
  "sha2",
  "sha3 0.9.1",
  "thiserror",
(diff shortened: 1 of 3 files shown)


pro-wh:
well that was bad timing

PR #499, opened 2021-10-04
modules/evm: switch to sha3 crate for keccak

### LINKED ISSUE #497: modules/evm: migrate tiny_keccak -> sha3

pro-wh:
fixes #497

### DIFF
--- a/runtime-sdk/modules/evm/src/derive_caller.rs
+++ b/runtime-sdk/modules/evm/src/derive_caller.rs
@@ -1,4 +1,4 @@
-use tiny_keccak::{Hasher, Keccak};
+use sha3::Digest as _;
 
 use oasis_runtime_sdk::{
     crypto::signature::{secp256k1, PublicKey},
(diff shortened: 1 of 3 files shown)


runtime-sdk/modules/evm/src/derive_caller.rs
```diff
@@ -1,7 +1,7 @@
-use tiny_keccak::{Hasher, Keccak};
+use sha3::Digest as _;
 
 use oasis_runtime_sdk::{
-    crypto::signature::{secp256k1, PublicKey},
+    crypto::signature::{PublicKey, secp256k1},
```
pro-wh, line 4: >_>

kostko (approved):

PR #507, opened 2021-10-06
grpc error details

pro-wh:
for convenience, I'll check in the compiled version

might be good to add a CI step to make sure they're in sync


pro-wh:
preview:
image

core playground bundle:

```
before
-rw-r--r-- 1 wh0 wh0 906878 Oct  6 16:10 main.js

after
-rw-r--r-- 1 wh0 wh0 1164833 Oct  6 16:11 main.js
```

webpack stats visualization:

image

this sucks, let's not merge it

pro-wh:
https://github.com/oasisprotocol/oasis-core/issues/4298 maybe we can improve this in oasis-core instead

PR #519, opened 2021-10-08
ci: explicit npm packages order

pro-wh:
I was way wrong about `--foreground-scripts`. It was coincidence that `core` sorted before the other directories that depended on it.

~~So here's a change to use `--ignore-scripts` and then execute the `prepare` scripts manually in an order that works.~~

So here's a special directory of numbered symlinks to the package dirs. this way we can have a super-explicit ordering and human readable names

npm workspaces bad

### DIFF
--- a/client-sdk/ts-web/package-lock.json
+++ b/client-sdk/ts-web/package-lock.json
@@ -5,15 +5,13 @@
     "packages": {
         "": {
             "workspaces": [
-                "core",
-                "ext-utils",
-                "rt",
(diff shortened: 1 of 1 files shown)


pro-wh:
probably not right to use `npm ci` multiple times. gonna switch to `npm install` ):

pro-wh:
> ```
> npm ERR! Cannot set property 'dev' of null
> ```

what

npm workspaces bad

PR #525, opened 2021-10-11
client-sdk/go: module events

ptrus:
- adds support for querying sdk events to the go client-sdk
- adds support for querying contract events

### DIFF
--- a/client-sdk/go/client/client.go
+++ b/client-sdk/go/client/client.go
@@ -71,14 +71,38 @@ type RuntimeClient interface {
 	// with their results and emitted events.
 	GetTransactionsWithResults(ctx context.Context, round uint64) ([]*TransactionWithResults, error)
 
-	// GetEvents returns all events emitted in a given block.
-	GetEvents(ctx context.Context, round uint64) ([]*coreClient.Event, error)
+	// GetEventsRaw returns all events emitted in a given block.
+	GetEventsRaw(ctx context.Context, round uint64) ([]*types.Event, error)
+
+	// GetEvents returns and decodes events emitted in a given block with the provided decoders.
+	GetEvents(ctx context.Context, round uint64, decoders []EventDecoder, includeUndecoded bool) ([]DecodedEvent, error)
 
 	// WatchBlocks subscribes to blocks for a specific runtimes.
 	WatchBlocks(ctx context.Context) (<-chan *roothash.AnnotatedBlock, pubsub.ClosableSubscription, error)
 
+	// WatchEvents subscribes and decodes runtime events.
+	WatchEvents(ctx context.Context, decoders []EventDecoder, includeUndecoded bool) (<-chan *BlockEvents, error)
+
 	// Query makes a runtime-specific query.
 	Query(ctx context.Context, round uint64, method string, args, rsp interface{}) error
+}
+
+// EventDecoder is an event decoder interface.
+type EventDecoder interface {
+	// DecodeEvent decodes an event. In case the event is not relevant, `nil, nil` should be returned.
+	DecodeEvent(*types.Event) (DecodedEvent, error)
+}
+
+// DecodedEvent is a decoded event.
+type DecodedEvent interface{}
+
+// BlockEvents are the events emitted in a block.
+type BlockEvents struct {
+	// Round is the round of the block.
+	Round uint64
+
+	// Events are the decoded events.
+	Events []DecodedEvent
 }
 
 // TransactionMeta are the metadata about transaction execution.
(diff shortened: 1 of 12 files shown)


client-sdk/go/modules/contracts/contracts.go
```diff
@@ -222,6 +228,43 @@ func (a *v1) Custom(ctx context.Context, round uint64, id InstanceID, data, rsp
...
+		}
+		evs = append(evs, ev.(*Event))
+	}
+
+	return evs, nil
+}
+
+// Implements client.EventDecoder.
+func (a *v1) DecodeEvent(event *types.Event) (client.DecodedEvent, error) {
+	if !strings.HasPrefix(event.Module, ModuleName) {
```
kostko, line 258: Probably should be that either it is `ModuleName` or the prefix is `ModuleName + "."`. So for example if there is a module called `contractsfoo` it doesn't clash.

client-sdk/go/client/client.go
```diff
@@ -318,11 +342,99 @@ func (rc *runtimeClient) GetTransactionsWithResults(ctx context.Context, round u
...
+		var ev types.Event
+		if err := ev.UnmarshalRaw(rawEv.Key, rawEv.Value); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal event '%v': %w", rawEv, err)
+		}
+		for _, decoder := range decoders {
+			decoded, err := decoder.DecodeEvent(&ev)
+			if err != nil {
+				return nil, fmt.Errorf("failed to decode event: %w", err)
+			}
+			if decoder != nil {
```
kostko, line 388: Should this be `decoded` not `decoder`?

client-sdk/go/client/client.go
```diff
@@ -318,11 +342,99 @@ func (rc *runtimeClient) GetTransactionsWithResults(ctx context.Context, round u
...
+
+	return evs, nil
+}
+
+// Implements RuntimeClient.
+func (rc *runtimeClient) WatchEvents(ctx context.Context, decoders []EventDecoder, includeUndecoded bool) (<-chan *BlockEvents, pubsub.ClosableSubscription, error) {
+	notifier := pubsub.NewBroker(true)
+	typedCh := make(chan *BlockEvents)
+	sub := notifier.Subscribe()
+	sub.Unwrap(typedCh)
```
kostko, line 406: Why are you creating a new broker here if you only have one subscription? A channel should be enough.

client-sdk/go/modules/contracts/oas20/oas20.go
```diff
@@ -93,3 +130,157 @@ type Response struct {
...
 }
+
+// EventDecoder is the OAS20 contract events decoder.
+type EventDecoder struct {
+	InstanceID contracts.InstanceID
+}
+
+// DecodeEvent implements client.EventDecoder.
+func (ed *EventDecoder) DecodeEvent(event *types.Event) (client.DecodedEvent, error) {
+	if !strings.HasPrefix(event.Module, contracts.ModuleName+".") {
```
kostko, line 141: Could also take a code ID and only decode events emitted by that code?

tests/e2e/simplekvtest.go
```diff
@@ -537,10 +537,35 @@ func KVTransferTest(sc *RuntimeScenario, log *logging.Logger, conn *grpc.ClientC
...
 	_ = tb.AppendSign(ctx, testing.Alice.Signer)
-	if err = tb.SubmitTx(ctx, nil); err != nil {
+	var meta *client.TransactionMeta
+	if meta, err = tb.SubmitTxMeta(ctx, nil); err != nil {
 		return err
 	}
 
+	evs, err := ac.GetEvents(ctx, meta.Round)
+	if err != nil {
+		return fmt.Errorf("fetching events:%w", err)
```
kostko, line 547:
```suggestion
		return fmt.Errorf("failed to fetch events: %w", err)
```

client-sdk/go/modules/contracts/oas20/oas20.go
ptrus: Yeah i think that makes sense. In the worst case, the user might need to do one more query to get the code ID in case they only have the instance ID.

client-sdk/go/client/client.go
ptrus: fixed

client-sdk/go/client/client.go
```diff
@@ -318,11 +342,96 @@ func (rc *runtimeClient) GetTransactionsWithResults(ctx context.Context, round u
...
+	ch := make(chan *BlockEvents)
+
+	blkCh, blkSub, err := rc.cc.WatchBlocks(ctx, rc.runtimeID)
+	if err != nil {
+		return nil, err
+	}
+
+	go func() {
+		defer blkSub.Close()
+
```
kostko, line 412: Also do `defer close(ch)`.

ptrus: done :+1:

PR #532, opened 2021-10-13
runtime-sdk/modules/evm: Add evm.SimulateCall query

### LINKED ISSUE #529: modules/evm: Add evm.SimulateCall query
We need to add an `evm.SimulateCall` query that simulates a call and returns the result. This will be needed to implement the [`eth_call`](https://www.quicknode.com/docs/ethereum/eth_call) Web3 endpoint.

abukosek:
Closes #529.

### DIFF
--- a/runtime-sdk/modules/evm/src/lib.rs
+++ b/runtime-sdk/modules/evm/src/lib.rs
@@ -15,7 +15,7 @@ use once_cell::sync::Lazy;
 use thiserror::Error;
 
 use oasis_runtime_sdk::{
-    context::{Context, TxContext},
+    context::{BatchContext, Context, TxContext},
     error,
     module::{self, CallResult, Module as _},
     modules::{
@@ -24,7 +24,7 @@ use oasis_runtime_sdk::{
         core::{self, Error as CoreError, API as _},
     },
     storage,
-    types::{address::Address, token, transaction::Transaction},
+    types::{address::Address, token, transaction, transaction::Transaction},
 };
 
 use evm::backend::ApplyBackend;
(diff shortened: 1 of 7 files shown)


runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -414,12 +482,14 @@ impl<Cfg: Config> Module<Cfg> {
...
         // Move the difference from the fee accumulator back into the
         // EVM token pool.
-        Cfg::Accounts::move_from_fee_accumulator(
-            ctx,
-            *ADDRESS_EVM_TOKENS,
-            &token::BaseUnits::new(return_fee.as_u128(), den),
-        )
-        .map_err(|_| Error::InsufficientBalance)?;
+        if !is_sim {
+            Cfg::Accounts::move_from_fee_accumulator(
```
kostko, line 486: Should `move_from_fee_accumulator` (and similar) check if ctx is in simulation mode and just not do the move?

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -345,6 +356,63 @@ impl<Cfg: Config> API for Module<Cfg> {
...
+                                .as_u128(),
+                            den,
+                        ),
+                        gas: gas_limit,
+                        consensus_messages: 0,
+                    },
+                },
+            };
+            sctx.with_tx(
+                cbor::to_vec(call_tx.clone()).len() as u32,
```
kostko, line 400: No need for that as you are not doing gas estimation, can just pass zero.

runtime-sdk/src/modules/accounts/mod.rs
```diff
@@ -496,6 +496,10 @@ impl API for Module {
         from: Address,
         amount: &token::BaseUnits,
     ) -> Result<(), modules::core::Error> {
+        if ctx.is_simulation() {
```
kostko, line 499: Add unit test to make sure that nothing is changed in simulation mode, but is changed outside simulation mode. Also for below.

PR #541, opened 2021-10-15
modules/evm: Unify EVM and SDK accounts

kostko:


runtime-sdk/modules/evm/src/backend.rs
```diff
@@ -0,0 +1,203 @@
+//! EVM backend.
```
ptrus, line 1: Could move the file rename in a separate commit, so that the actual changes can be reviewed more easily

client-sdk/go/types/address.go
```diff
@@ -143,6 +143,11 @@ func NewAddress(spec SignatureAddressSpec) (a Address) {
 	return (Address)(address.NewAddress(ctx, pkData))
 }
 
+// NewAddressRaw creates a new address from passed address context and data.
+func NewAddressRaw(ctx address.Context, data []byte) Address {
```
ptrus, line 147: Add a test case to `address_test.go`

runtime-sdk/modules/evm/src/backend.rs
```diff
@@ -0,0 +1,203 @@
...
+                    let addr: H160 = address.into();
+                    // Derive SDK account address from the Ethereum address.
+                    let address = Cfg::map_address(address);
+
+                    // Update account balance and nonce.
+                    let mut state = self.ctx.get_mut().runtime_state();
+                    let amount =
+                        token::BaseUnits::new(basic.balance.as_u128(), Cfg::TOKEN_DENOMINATION);
+                    Cfg::Accounts::set_balance(&mut state, address, &amount);
+                    Cfg::Accounts::set_nonce(state, address, basic.nonce.as_u64());
```
abukosek, line 159: These `{get,set}_{balance,nonce}` calls seem to always appear in pairs -- would it make more sense to join them into a `{get,set}_balance_and_nonce` call?  This would save one load and one store operation from/to storage.

runtime-sdk/modules/evm/src/backend.rs
```diff
@@ -0,0 +1,203 @@
...
+                    if let Some(code) = code {
+                        let state = self.ctx.get_mut().runtime_state();
+                        let mut store = state::codes(state);
+                        store.insert(&addr, code);
+                    }
+
+                    // Handle storage reset.
+                    if reset_storage {
+                        // NOTE: Storage cannot be efficiently reset as this would require iterating
+                        //       over all of the storage keys. We could add this is remove_prefix
```
abukosek, line 171:
```suggestion
                        //       over all of the storage keys. We could add this if remove_prefix
```

ptrus (approved):

runtime-sdk/modules/evm/src/evm_backend.rs
```diff
@@ -91,66 +84,55 @@ impl<'c, C: oasis_runtime_sdk::Context> EVMBackend for Backend<'c, C> {
...
 
     fn original_storage(
         &self,
-        address: primitive_types::H160,
-        index: primitive_types::H256,
+        _address: primitive_types::H160,
+        _index: primitive_types::H256,
     ) -> Option<primitive_types::H256> {
-        Some(self.storage(address, index))
+        None
```
pro-wh, line 130: need explanation for this change

runtime-sdk/modules/evm/src/evm_backend.rs
```diff
@@ -166,82 +148,45 @@ impl<'c, C: oasis_runtime_sdk::Context> ApplyBackend for Backend<'c, C> {
...
                 } => {
                     let addr: H160 = address.into();
-                    let is_empty = {
+                    // Derive SDK account address from the Ethereum address.
+                    let address = Cfg::map_address(address);
+
+                    // Update account balance and nonce.
+                    let mut state = self.ctx.get_mut().runtime_state();
+                    let amount =
+                        token::BaseUnits::new(basic.balance.as_u128(), Cfg::TOKEN_DENOMINATION);
```
pro-wh, line 157: these panic if the u256 doesn't fit. maybe there are total-supply-related reasons why this wouldn't happen. although I guess the api of this backend doesn't let us return an error anyway

runtime-sdk/modules/evm/src/evm_backend.rs
```diff
@@ -166,82 +148,45 @@ impl<'c, C: oasis_runtime_sdk::Context> ApplyBackend for Backend<'c, C> {
...
-                            let state = self.ctx.get_mut().runtime_state();
-                            let s_store = storage::PrefixStore::new(state, &crate::MODULE_NAME);
-                            let mut storages = storage::PrefixStore::new(s_store, &state::STORAGES);
-                            storages.remove(addr.as_bytes());
-                        }
+                    // Handle storage reset.
+                    if reset_storage {
+                        // NOTE: Storage cannot be efficiently reset as this would require iterating
+                        //       over all of the storage keys. We could add this if remove_prefix
+                        //       existed.
```
pro-wh, line 172: so we do nothing? is it correct for us to do that?

what did the removed code do?

runtime-sdk/modules/evm/src/evm_backend.rs
```diff
@@ -166,82 +148,45 @@ impl<'c, C: oasis_runtime_sdk::Context> ApplyBackend for Backend<'c, C> {
...
+                    for (index, value) in storage {
+                        let idx: H256 = index.into();
+                        let val: H256 = value.into();
 
-                            let state = self.ctx.get_mut().runtime_state();
-                            let s_store = storage::PrefixStore::new(state, &crate::MODULE_NAME);
-                            let storages = storage::PrefixStore::new(s_store, &state::STORAGES);
-                            let mut s = storage::TypedStore::new(storage::HashedStore::<
-                                _,
-                                blake3::Hasher,
```
pro-wh, line 209: what was this about a hasher in the previous implementation? is that removed in this pr? if so, what will that do?

runtime-sdk/modules/evm/src/evm_backend.rs
```diff
@@ -166,82 +148,45 @@ impl<'c, C: oasis_runtime_sdk::Context> ApplyBackend for Backend<'c, C> {
...
-                    let state = self.ctx.get_mut().runtime_state();
-                    let store = storage::PrefixStore::new(state, &crate::MODULE_NAME);
-                    let mut accounts = storage::TypedStore::new(storage::PrefixStore::new(
-                        store,
-                        &state::ACCOUNTS,
-                    ));
-
-                    accounts.remove(&addr);
+                Apply::Delete { .. } => {
+                    // Accounts cannot be deleted.
```
pro-wh, line 189: what from evm would attempt to do this?

runtime-sdk/modules/evm/src/evm_backend.rs
```diff
@@ -166,82 +148,45 @@ impl<'c, C: oasis_runtime_sdk::Context> ApplyBackend for Backend<'c, C> {
...
-                Apply::Delete { address } => {
-                    let addr: H160 = address.into();
-                    let state = self.ctx.get_mut().runtime_state();
-                    let store = storage::PrefixStore::new(state, &crate::MODULE_NAME);
-                    let mut accounts = storage::TypedStore::new(storage::PrefixStore::new(
-                        store,
-                        &state::ACCOUNTS,
-                    ));
-
-                    accounts.remove(&addr);
```
pro-wh, line 244: did we used to just do this and have the balance disappear?

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -38,17 +39,37 @@ const MODULE_NAME: &str = "evm";
...
+    ) -> storage::TypedStore<impl storage::Store + 'a> {
+        let store = storage::PrefixStore::new(state, &crate::MODULE_NAME);
+        let storages = storage::PrefixStore::new(store, &STORAGES);
+        storage::TypedStore::new(storage::HashedStore::<_, blake3::Hasher>::new(
+            storage::PrefixStore::new(storages, address),
+        ))
+    }
+
+    /// Get a typed store for codes of all contracts.
+    pub fn codes<'a, S: storage::Store + 'a>(
```
pro-wh, line 67: no similar one for block hashes?

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -59,16 +80,24 @@ pub trait Config: 'static {
...
+    }
 }
 
 pub struct Module<Cfg: Config> {
     _cfg: std::marker::PhantomData<Cfg>,
 }
 
-/// EVM token pool address.
-pub static ADDRESS_EVM_TOKENS: Lazy<Address> =
-    Lazy::new(|| Address::from_module(MODULE_NAME, "evm-tokens"));
```
pro-wh, line 70: we used to have this address do an additional layer of accounting to keep the evm itself honest. now we let the evm directly do set_balance on the accounts module.

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -80,43 +109,30 @@ pub enum Error {
     #[sdk_error(code = 2)]
     EVMError(String),
 
-    #[error("invalid signer type")]
-    #[sdk_error(code = 3)]
-    InvalidSignerType,
```
pro-wh, line 85: can you explain the removal of this? will the evm module now accept calls from all signers?

I do see that derive_caller supports other signer types. but the sdk account derivation wouldn't match between (i) ed25519 signature -> oasis-staking-compatible account and (ii) ed25519 signature -> H160 caller -> ethereum-compatible sdk account

pro-wh (commented):
looking through the evm backend changes :eyeglasses:

runtime-sdk/modules/evm/src/evm_backend.rs
```diff
@@ -91,66 +84,55 @@ impl<'c, C: oasis_runtime_sdk::Context> EVMBackend for Backend<'c, C> {
...
-        let store = storage::PrefixStore::new(state, &crate::MODULE_NAME);
-        let accounts = storage::TypedStore::new(storage::PrefixStore::new(store, &state::ACCOUNTS));
+        let mut state = ctx.runtime_state();
 
-        let a: Account = accounts.get(&addr).unwrap_or_default();
+        // Derive SDK account address from the Ethereum address.
+        let address = Cfg::map_address(address);
+        // Fetch balance and nonce from SDK accounts.
+        let balance = Cfg::Accounts::get_balance(&mut state, address, Cfg::TOKEN_DENOMINATION)
+            .unwrap_or_default();
```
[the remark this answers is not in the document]
kostko: Well the thing is that `get_balance` can actually never fail. Maybe we should just remove the result type.

runtime-sdk/modules/evm/src/evm_backend.rs
```diff
@@ -91,66 +84,55 @@ impl<'c, C: oasis_runtime_sdk::Context> EVMBackend for Backend<'c, C> {
...
-
-        let store = storage::PrefixStore::new(state, &crate::MODULE_NAME);
-        let accounts = storage::TypedStore::new(storage::PrefixStore::new(store, &state::ACCOUNTS));
+        let mut state = ctx.runtime_state();
 
-        let a: Account = accounts.get(&addr).unwrap_or_default();
+        // Derive SDK account address from the Ethereum address.
+        let address = Cfg::map_address(address);
+        // Fetch balance and nonce from SDK accounts.
+        let balance = Cfg::Accounts::get_balance(&mut state, address, Cfg::TOKEN_DENOMINATION)
```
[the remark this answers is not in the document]
kostko: Yes was trying to do that as well, but it turns out this is a rats nest so I would rather not do it in this PR.

runtime-sdk/modules/evm/src/evm_backend.rs
kostko: This is correct, since everything is now backed by accounts that use `u128` this can never actually fail unless the EVM can create tokens out of thin air (and in this case the whole module is seriously broken).

pro-wh: that the sdk total supply is u128, yeah

runtime-sdk/modules/evm/src/evm_backend.rs
kostko: It is not removed, just refactored so that it is not repeated over and over. See the `state::storage` function.

runtime-sdk/modules/evm/src/evm_backend.rs
kostko: The dumb selfdestruct opcode (see my comment above).

runtime-sdk/modules/evm/src/evm_backend.rs
pro-wh: found it, thanks

runtime-sdk/modules/evm/src/evm_backend.rs
pro-wh: oh ok

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -38,17 +39,37 @@ const MODULE_NAME: &str = "evm";
...
     pub const BLOCK_HASH_WINDOW_SIZE: u64 = 256;
+
+    /// Get a typed store for the given address' storage.
+    pub fn storage<'a, S: storage::Store + 'a>(
+        state: S,
+        address: &'a H160,
+    ) -> storage::TypedStore<impl storage::Store + 'a> {
+        let store = storage::PrefixStore::new(state, &crate::MODULE_NAME);
+        let storages = storage::PrefixStore::new(store, &STORAGES);
+        storage::TypedStore::new(storage::HashedStore::<_, blake3::Hasher>::new(
```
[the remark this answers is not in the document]
kostko: Yep :-)

runtime-sdk/modules/evm/src/lib.rs
pro-wh: I see that they "apply" in batches. maybe we could check that all the changes balance out. unless the evm can burn. this would take some additional get_balance calls. dunno if it's worth it

runtime-sdk/modules/evm/src/lib.rs
kostko: Oh I see, maybe we need to fix this, hm.

kostko: Yeah so while this works it would result in the address mapping into a different SDK account. You could still control the funds and everything would work, except that SDK account could never be used to sign any transactions, it would just be used to store the EVM balance. We could forbid these to make it less confusing for users.

kostko: Yeah I think forbidding it would be best, it can always be allowed later if we decide this would be fine after all.

runtime-sdk/modules/evm/src/lib.rs
kostko: No those stores shouldn't have a lot of overhead, but there is an extra storage read. But they should be cached anyway as the EVM executor would always first access the account. So these should be cheap?

pro-wh (approved):
gonna leave this on approved. we have some nice ideas for enhancement

- explain intentional divergences from selfdestruct behavior
- document certain infallible-but-returns-result methods
- tighten up derive_caller now that we're only supporting secp256k1eth. maybe
- increase sanity checks on balance manipulation

I'll need to retroactively review the client updates

edit: no changes to the ts-web directory hehe

runtime-sdk/modules/evm/src/lib.rs
kostko: Added a paranoid sanity check.

runtime-sdk/modules/evm/src/derive_caller.rs
```diff
@@ -29,11 +29,11 @@ pub fn from_sigspec(spec: &SignatureAddressSpec) -> H160 {
...
 
-pub fn from_tx_auth_info(ai: &AuthInfo) -> H160 {
+pub fn from_tx_auth_info(ai: &AuthInfo) -> Result<H160, Error> {
     match &ai.signer_info[0].address_spec {
         AddressSpec::Signature(SignatureAddressSpec::Secp256k1Eth(pk)) => {
-            from_secp256k1_public_key(pk)
+            Ok(from_secp256k1_public_key(pk))
         }
-        address_spec => from_non_secp256k1_address(&address_spec.address()),
+        _ => Err(Error::InvalidSignerType),
```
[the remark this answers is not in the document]
kostko: Done.

runtime-sdk/modules/evm/src/lib.rs
kostko: Added.

pro-wh:
> Yeah, it seems we don't have any EVM helpers for the ts-web client. Could you maybe add them? :-)

yes

PR #543, opened 2021-10-15
modules/consensus_accounts: query_balance succeed on zero 

### LINKED ISSUE #538: consensus.Balance query: return zero if the balance is zero
there's one piece of logic that makes it currently error out with InvalidArgument if the entry for the consensus denomination is absent. however, that seems to be the proper way of representing zero balance. we should return zero in this case instead of an unexplained InvalidArgument error

pro-wh:
accounts that have had no activity and thus no consensus denomination in their account implicitly have zero balance. this change makes the query_balance query return that implicit zero as the balance instead of erroring out with illegal argument. it's intended that the runtime configuration of the consensus denomination is trusted, in that the denomination exists

fixes #538

### DIFF
--- a/client-sdk/ts-web/rt/playground/src/consensus.js
+++ b/client-sdk/ts-web/rt/playground/src/consensus.js
@@ -94,6 +94,18 @@ export const playground = (async function () {
         await twDeposit.sign([csAlice], consensusChainContext);
         await twDeposit.submit(nic);
 
+        console.log('alice balance');
+        const balanceResult = await consensusWrapper
+            .queryBalance()
+            .setArgs({
+                address: await oasis.staking.addressFromPublicKey(alice.public()),
+            })
+            .query(nic);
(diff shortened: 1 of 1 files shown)


client-sdk/ts-web/rt/playground/src/consensus.js
```diff
@@ -94,6 +94,16 @@ export const playground = (async function () {
...
 
+        console.log('alice balance');
+        const balanceResult = await consensusWrapper
+            .queryBalance()
+            .setArgs({
+                address: await oasis.staking.addressFromPublicKey(alice.public()),
+            })
+            .query(nic);
+        console.log('balance', oasis.quantity.toBigInt(balanceResult.balance));
+        console.log('we just deposited, but it\'s okay if this is zero before the roothash callback runs');
```
pro-wh, line 105: can someone confirm that this is right?

kostko (approved):

PR #544, opened 2021-10-16
deposit/withdraw across accounts

### LINKED ISSUE #542: Add destination to consensusaccounts.Deposit/Withdraw
Currently deposits to and withdrawals from the runtime always move tokens to the address derived from the signer. However there is no reason for that and we could optionally allow the caller to specify a different `destination` address (either in the runtime - for deposits - or in the consensus layer - for withdrawals) instead.

pro-wh:
this adds a `to` field to the consensus accounts Deposit and Withdraw methods, which lets you transfer to different accounts across the consensus--runtime boundary. enter your own address for the previous behavior of only allowing transfer to your own corresponding account. this is similar to depositing and then transferring or transferring and then withdrawing (supposing you have the keys to both accounts)

fixes #542

### DIFF
--- a/tests/e2e/simple_consensus.go
+++ b/tests/e2e/simple_consensus.go
@@ -39,6 +39,21 @@ func ensureStakingEvent(log *logging.Logger, ch <-chan *staking.Event, check fun
 	}
 }
 
+func makeTransferCheck(from, to staking.Address, amount *quantity.Quantity) func(e *staking.Event) bool {
+	return func(e *staking.Event) bool {
+		if e.Transfer == nil {
+			return false
+		}
+		if e.Transfer.From != from {
+			return false
+		}
+		if e.Transfer.To != to {
+			return false
+		}
+		return e.Transfer.Amount.Cmp(amount) == 0
+	}
+}
+
 func ensureRuntimeEvent(log *logging.Logger, ch <-chan *client.BlockEvents, check func(event client.DecodedEvent) bool) error {
 	log.Info("waiting for expected runtime event...")
 	for {
@@ -54,6 +69,50 @@ func ensureRuntimeEvent(log *logging.Logger, ch <-chan *client.BlockEvents, chec
 		case <-time.After(timeout):
 			return fmt.Errorf("timeout waiting for event")
 		}
+	}
+}
+
+func makeAccountsMintCheck(owner types.Address, amount types.BaseUnits) func(e client.DecodedEvent) bool {
+	return func(e client.DecodedEvent) bool {
+		ae, ok := e.(*accounts.Event)
+		if !ok {
+			return false
+		}
+		if ae.Mint == nil {
+			return false
+		}
+		if !ae.Mint.Owner.Equal(owner) {
+			return false
+		}
+		if ae.Mint.Amount.Amount.Cmp(&amount.Amount) != 0 {
+			return false
+		}
+		if ae.Mint.Amount.Denomination != amount.Denomination {
+			return false
+		}
+		return true
+	}
+}
+
+func makeBurnCheck(owner types.Address, amount types.BaseUnits) func(e client.DecodedEvent) bool {
+	return func(e client.DecodedEvent) bool {
+		ae, ok := e.(*accounts.Event)
+		if !ok {
+			return false
+		}
+		if ae.Burn == nil {
+			return false
+		}
+		if !ae.Burn.Owner.Equal(owner) {
+			return false
+		}
+		if ae.Burn.Amount.Amount.Cmp(&amount.Amount) != 0 {
+			return false
+		}
+		if ae.Burn.Amount.Denomination != amount.Denomination {
+			return false
+		}
+		return true
 	}
 }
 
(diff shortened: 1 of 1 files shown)


pro-wh:
ok it didn't just work

runtime-sdk/src/modules/consensus_accounts/types.rs
```diff
@@ -2,14 +2,21 @@
 use crate::types::{address::Address, token};
 
 /// Deposit into runtime call.
+/// Transfer from consensus staking to an account in this runtime.
+/// The transaction signer has a consensus layer allowance benefiting this runtime's staking
+/// address. The `to` address runtime account gets the tokens.
 #[derive(Clone, Debug, cbor::Encode, cbor::Decode)]
 pub struct Deposit {
+    pub to: Option<Address>,
```
kostko, line 10: Add `#[cbor(optional)]` here to allow the field to be omitted.

runtime-sdk/src/modules/consensus_accounts/types.rs
```diff
@@ -2,14 +2,21 @@
...
     pub amount: token::BaseUnits,
 }
 
 /// Withdraw from runtime call.
+/// Transfer from an account in this runtime to consensus staking.
+/// The `to` address consensus staking account gets the tokens.
 #[derive(Clone, Debug, cbor::Encode, cbor::Decode)]
 pub struct Withdraw {
+    pub to: Option<Address>,
```
kostko, line 19: Add `#[cbor(optional)]` here to allow the field to be omitted.

client-sdk/go/modules/consensusaccounts/consensus_accounts.go
```diff
@@ -22,10 +22,10 @@ const (
 // V1 is the v1 consensus accounts module interface.
 type V1 interface {
 	// Deposit generates a consensus.Deposit transaction.
-	Deposit(amount types.BaseUnits) *client.TransactionBuilder
+	Deposit(to types.Address, amount types.BaseUnits) *client.TransactionBuilder
```
kostko, line 25: This makes it hard to omit `to`? Could accept `*types.Address` instead?

client-sdk/go/modules/consensusaccounts/types.go
```diff
@@ -4,11 +4,13 @@ import "github.com/oasisprotocol/oasis-sdk/client-sdk/go/types"
 
 // Deposit are the arguments for consensus.Deposit method.
 type Deposit struct {
+	To     types.Address   `json:"to,omitempty"`
```
kostko, line 7: This probably needs to be `*types.Address`.

pro-wh:
even without the optional `to` complexity, this is still experiencing

```
after deposit, expected bob balance 50, got 0
```

runtime-sdk/src/modules/consensus_accounts/types.rs
pro-wh: ok

client-sdk/go/modules/consensusaccounts/types.go
pro-wh: oh yeah probably

client-sdk/go/modules/consensusaccounts/consensus_accounts.go
pro-wh: I'll make it * to match the struct field type. otherwise we wouldn't be able to test unset-to easily

pro-wh: ironically, making this *types.Address might make it a little less convenient

tests/e2e/simple_consensus.go
```diff
@@ -151,33 +178,36 @@ func SimpleConsensusTest(sc *RuntimeScenario, log *logging.Logger, conn *grpc.Cl
 		return fmt.Errorf("alice withdrawing with invalid nonce should fail")
 	}
 
+	log.Info("wait for changes to propagate to client node")
+	time.Sleep(propagationTime)
```
pro-wh, line 182: is there a better way to do this?

and is it weird that this change pushed us over some cliff that we need this now?

tests/e2e/simple_consensus.go
```diff
@@ -104,9 +118,22 @@ func SimpleConsensusTest(sc *RuntimeScenario, log *logging.Logger, conn *grpc.Cl
 		return fmt.Errorf("ensuring bob deposit consensus event: %w", err)
 	}
 
+	log.Info("wait for changes to propagate to client node")
```
kostko, line 121: After the consensus layer event is emitted the runtime also needs to process the deposit. Let's not use `time.Sleep`, wait for the next round (or specific accounts transfer/mint/burn runtime event) instead.

tests/e2e/simple_consensus.go
kostko: Yes for withdrawals, this should not be needed as you are already waiting on the consensus layer event. The problem is that you are trying to query consensus layer balances through the runtime which does not use the latest hegiht AFAIK. Why don't you query the consensus layer directly?

pro-wh: > The problem is that you are trying to query consensus layer balances through the runtime which does not use the latest height AFAIK. Why don't you query the consensus layer directly?

oh correction on this: the consensus accounts balance query is not for obtaining information from the consensus _layer_. it gives the balance of the consensus _denomination_, still in the runtime layer

kostko: Only deposits should require waiting for the next round, are you doing withdrawals and deposits at the same time? Of course withdrawals can in theory be refunded if they fail at the consensus layer, but that shouldn't happen.

PR #548, opened 2021-10-18
contract-sdk: use C abi instead of wasm

ptrus:
The WASM abi seems quite broken, there are various problems with it:
- problems with u128 https://github.com/rust-lang/rust/issues/88207
- problems with compiling multivalue functions https://github.com/ptrus/wasm-bug

Switch to using C abi. This requires modifying the exported functions to not use multivalue returns (which was only used for returning `HostRegion`s currently).

### DIFF
--- a/contract-sdk/src/abi/dispatch.rs
+++ b/contract-sdk/src/abi/dispatch.rs
@@ -102,7 +102,7 @@ where
     (ctx, request)
 }
 
-fn handle_result<R, E>(ctx: Context, result: Result<Option<R>, E>) -> HostRegion
+fn handle_result<R, E>(ctx: Context, result: Result<Option<R>, E>) -> *const HostRegion
 where
     R: cbor::Encode,
     E: error::Error,
@@ -116,7 +116,7 @@ where
         Err(err) => err.to_execution_result(),
     };
 
-    HostRegion::from_vec(cbor::to_vec(result))
+    Box::into_raw(Box::new(HostRegion::from_vec(cbor::to_vec(result))))
 }
 
 /// Internal helper for calling the contract's `instantiate` function.
@@ -126,7 +126,7 @@ pub fn instantiate<C: Contract>(
     ctx_len: u32,
     request_ptr: u32,
     request_len: u32,
-) -> HostRegion {
+) -> *const HostRegion {
     let (mut ctx, request) = load_request_context(ctx_ptr, ctx_len, request_ptr, request_len);
     let result = C::instantiate(&mut ctx, request).map(Option::Some);
     handle_result(ctx, result)
@@ -139,7 +139,7 @@ pub fn call<C: Contract>(
     ctx_len: u32,
     request_ptr: u32,
     request_len: u32,
-) -> HostRegion {
+) -> *const HostRegion {
     let (mut ctx, request) = load_request_context(ctx_ptr, ctx_len, request_ptr, request_len);
     let result = C::call(&mut ctx, request).map(Option::Some);
     handle_result(ctx, result)
@@ -152,7 +152,7 @@ pub fn handle_reply<C: Contract>(
     ctx_len: u32,
     request_ptr: u32,
     request_len: u32,
-) -> HostRegion {
+) -> *const HostRegion {
     let (mut ctx, reply) = load_request_context(ctx_ptr, ctx_len, request_ptr, request_len);
     let result = C::handle_reply(&mut ctx, reply);
     handle_result(ctx, result)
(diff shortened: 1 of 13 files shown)


contract-sdk/src/abi/dispatch.rs
```diff
@@ -102,7 +102,7 @@ where
     (ctx, request)
 }
 
-fn handle_result<R, E>(ctx: Context, result: Result<Option<R>, E>) -> HostRegion
+fn handle_result<R, E>(ctx: Context, result: Result<Option<R>, E>) -> *mut HostRegion
```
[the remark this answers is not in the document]
ptrus: yeah no reason really, switched to `*const`.

runtime-sdk/modules/contracts/src/abi/oasis/memory.rs
```diff
@@ -46,6 +46,19 @@ impl Region {
         }
     }
 
+    /// Dereferences a pointer to the region.
+    pub fn deref(memory: &wasm3::Memory<'_>, arg: u32) -> Self {
+        let arg = arg as usize;
+        let dst = memory.as_slice();
```
kostko, line 52: This should validate that arg is within the memory region as otherwise the `dst` derefs below will panic. So this needs to return a `Result`.

ptrus: Done. 

Also I changed the existing `BadPointer` checks in this module from: `self.offset + self.length > memory.size()` to `self.offset + self.length >= memory.size()` as I think this is the correct check, as memory addresses are zero-based index?

runtime-sdk/modules/contracts/src/abi/oasis/memory.rs
```diff
@@ -46,6 +46,26 @@ impl Region {
         }
     }
 
+    /// Dereferences a pointer to the region.
+    pub fn deref(memory: &wasm3::Memory<'_>, arg: u32) -> Result<Self, RegionError> {
+        let arg = arg as usize;
+
+        // Make sure the pointer is within WASM memory.
+        if arg as usize >= memory.size() {
```
kostko, line 54: This should be `arg + 8` (e.g. `arg + sizeof(HostRegion)`).

runtime-sdk/modules/contracts/src/abi/oasis/memory.rs
```diff
@@ -152,4 +172,67 @@ impl<Cfg: Config> OasisV1<Cfg> {
...
+            offset: outer.offset + 8,
+            length: outer.length - 8,
+        };
+
+        instance
+            .runtime()
+            .try_with_memory(|mut memory| -> Result<(), RegionError> {
+                inner.copy_from_slice(&mut memory, &data)?;
+
+                let dst = &mut memory.as_slice_mut()[outer.offset..outer.offset + 8];
```
kostko, line 206: This may panic due to a TODO in `allocate` about early validation if the guest returns an invalid offset. Let's add that validation to `allocate` (currently the check is only performed in region's `copy_from_slice`).

runtime-sdk/modules/contracts/src/abi/oasis/memory.rs
ptrus: yeah ofcourse

runtime-sdk/modules/contracts/src/abi/oasis/memory.rs
ptrus: added

runtime-sdk/modules/contracts/src/abi/oasis/memory.rs
```diff
@@ -152,4 +186,67 @@ impl<Cfg: Config> OasisV1<Cfg> {
...
+                dst[4..].copy_from_slice(&(inner.length as u32).to_le_bytes());
+
+                Ok(())
+            })
+            .unwrap()?;
+
+        Ok(outer.offset as u32)
+    }
+
+    /// Allocates a region in the memory and returns a pointer to it.
```
kostko, line 231:
```suggestion
    /// Allocates a region in WASM memory and returns a pointer to it.
```

runtime-sdk/modules/contracts/src/abi/oasis/memory.rs
```diff
@@ -152,4 +186,67 @@ impl<Cfg: Config> OasisV1<Cfg> {
...
+    pub fn allocate_region<C: Context>(
+        instance: &wasm3::Instance<'_, '_, ExecutionContext<'_, C>>,
+        region: Region,
+    ) -> Result<u32, RegionError> {
+        let mut data = [0u8; 8];
+        data[..4].copy_from_slice(&(region.offset as u32).to_le_bytes());
+        data[4..].copy_from_slice(&(region.length as u32).to_le_bytes());
```
kostko, line 238: These conversions and copies seem quite common, should we add a helper on the region type that returns a `[u8; 8]`?

kostko (approved):

PR #595, opened 2021-10-29
cli: Initial Oasis CLI

kostko:


pro-wh:
going to rebase and merge!

matevz (commented):
First pass, some minor questions.

cli/cmd/contracts.go
```diff
@@ -0,0 +1,370 @@
...
+
+var (
+	contractsInstantiatePolicy string
+	contractsUpgradesPolicy    string
+	contractsData              string
+	contractsTokens            []string
+
+	contractsCmd = &cobra.Command{
+		Use:   "contracts",
+		Short: "WebAssembly smart contracts operations",
```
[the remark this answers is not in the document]
kostko: Currently yes, but in theory for any ParaTime that includes the `contracts` module.

cli/config/network.go
```diff
@@ -0,0 +1,131 @@
...
+	return nil
+}
+
+// Remove removes an existing network.
+func (n *Networks) Remove(name string) error {
+	if _, exists := n.All[name]; !exists {
+		return fmt.Errorf("network '%s' does not exist", name)
+	}
+
+	delete(n.All, name)
```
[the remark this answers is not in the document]
kostko: Nice catch, yes we definitely should!

client-sdk/go/crypto/signature/ed25519/signer.go
```diff
@@ -26,6 +26,10 @@ func (w wrappedSigner) Reset() {
 	w.signer.Reset()
 }
 
+func (w wrappedSigner) Unwrap() coreSignature.Signer {
+	return w.signer
```
pro-wh, line 30: could alternatively make the signer field public

could alternatively store the consensus signer and wrap on demand

pro-wh (commented):
15/36

cli/cmd/wallet.go
```diff
@@ -0,0 +1,254 @@
...
+				Config: map[string]interface{}{
+					"algorithm": answers.Algorithm,
+				},
+			}
+			src := &wallet.ImportSource{
+				Kind: wallet.ImportKind(answers.Kind),
+				Data: answers.Data,
+			}
+
+			err = cfg.Wallets.Import(name, string(answers.Passphrase), walletCfg, src)
```
pro-wh, line 203: passphrase should already be a string here

cli/cmd/wallet.go
```diff
@@ -0,0 +1,254 @@
...
+					Name: "algorithm",
+					Prompt: &survey.Select{
+						Message: "Algorithm:",
+						Options: []string{"ed25519"},
+						Default: "ed25519",
+					},
+				},
+				{
+					Name:   "data",
+					Prompt: &survey.Multiline{Message: "Data to import:"},
```
pro-wh, line 175: here's a part that we could make friendlier imo. this question depends on what what you chose for 'kind,' but we cram it into a universal 'data' question. if we already knew, we could give more detailed instructions in the prompt, e.g. "mnemonic" for mnemonic imports, we could additionally ask for account number, or we could say that the private key must be base64

maybe we could split this off into a separate `Ask`

cli/cmd/wallet.go
```diff
@@ -0,0 +1,254 @@
...
+				},
+				{
+					Name:   "passphrase",
+					Prompt: common.PromptPassphrase,
```
pro-wh, line 179: :information_source: the passphrase is not something used to protect the derivation of the account key. this is a passphrase used to seal the account key for file-based wallets

cli/wallet/file/ed25519.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"encoding/base64"
+
+	"github.com/oasisprotocol/curve25519-voi/primitives/ed25519"
+	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
+)
+
+// ed25519rawSigner is an in-memory signer that allows deserialization of raw ed25519 keys for use
+// in imported wallets that don't use ADR 0008.
```
pro-wh, line 11: what do we use in the relatively raw signers in oasis-core, if not raw ed25519 keys?

cli/wallet/file/file.go
```diff
@@ -0,0 +1,348 @@
...
+
+func (s *secretState) Seal(passphrase string) (*secretStateEnvelope, error) {
+	var salt [stateSaltSize]byte
+	_, err := rand.Read(salt[:])
+	if err != nil {
+		return nil, err
+	}
+
+	envelope := &secretStateEnvelope{
+		Algorithm:  "pbkdf2",
```
pro-wh, line 63: if anyone wants to bikeshed, tell us your favorite alternative

pro-wh (commented):
33/36

cli/wallet/file/file.go
```diff
@@ -0,0 +1,348 @@
...
+
+func (wf *fileWalletFactory) Create(name string, passphrase string, rawCfg map[string]interface{}) (wallet.Wallet, error) {
+	cfg, err := wf.unmarshalConfig(rawCfg)
+	if err != nil {
+		return nil, err
+	}
+
+	// Generate entropy.
+	entropy, _ := bip39.NewEntropy(256)
+	mnemonic, _ := bip39.NewMnemonic(entropy)
```
pro-wh, line 176: could we panic on these, if you'd rather not propagate the errors

the implementation of NewEntropy looks like it returns unrandomized bytes if rand.Read fails. we handle rand.Read failures elsewhere in this program.

cli/wallet/file/file.go
```diff
@@ -0,0 +1,348 @@
...
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/types"
+)
+
+const (
+	walletKind = "file"
+
+	cfgAlgorithm = "algorithm"
+	cfgNumber    = "number"
+
+	algorithmEd25519    = "ed25519"
```
pro-wh, line 34: seeing as the Algorithm field covers the derivation algorithm and the resulting key type, maybe this should be ed25519-adr0008?

pro-wh: or no, this is used in both walletConfig and secretState, which kind of conflates the meaning.

when walletConfig.Algorithm is algorithmEd25519, it means "loading this wallet will give you an ed25519 key." but it's stored next to a Number field, which is rather something that controls a derivation not hinted at by this field

when secretState.Algorithm is algorithmEd25519, it means "use ADR 0008 to derive an ed25519 key from the mnemonic." it's stored next to that mnemonic.

some weirdness from all this:

- the choice between ADR 0008 derivation and raw private key is within the sealed data, but the account number for the ADR 0008 case is out in the open
- there's this piece of code that switches around algorithmEd25519Raw based on a field in ImportSource

cli/wallet/file/ed25519.go
pro-wh: and if not 'raw ed25519 keys,'

1. why make the CLI use raw ed25519 keys?
2. shall we instead add this functionality directly to oasis-core?

cli/wallet/file/file.go
```diff
@@ -0,0 +1,348 @@
...
+	switch src.Kind {
+	case wallet.ImportKindMnemonic:
+		// Import a mnemonic.
+		state.Algorithm = cfg.Algorithm
+		state.Data = src.Data
+	case wallet.ImportKindPrivateKey:
+		// Import a private key. This requires us to change the algorithm to the raw variant.
+		switch cfg.Algorithm {
+		case algorithmEd25519:
+			state.Algorithm = algorithmEd25519Raw
```
pro-wh, line 245: we use a special value here instead of, say storing the ImportSource.kind alongside. what does this get us? maybe in the future we can change the default derivation scheme used when the user only specifies "mnemonic"?

cli/wallet/file/file.go
```diff
@@ -0,0 +1,348 @@
...
+	algorithmEd25519    = "ed25519"
+	algorithmEd25519Raw = "ed25519-raw"
+
+	stateKeySize  = 32
+	stateSaltSize = 32
+)
+
+type walletConfig struct {
+	Algorithm string `mapstructure:"algorithm"`
+	Number    uint32 `mapstructure:"number,omitempty"`
```
pro-wh, line 43: I think this is never assigned. how do you envision this being populated in the future? fed in through a new field in ImportSource maybe?

pro-wh (changes requested):
requesting changes: error handling when obtaining entropy

other notes: I want to learn about the higher level architecture of the wallet data model, I've added questions inline

cli/wallet/file/ed25519.go
kostko: > what do we use in the relatively raw signers in oasis-core (file, memory), if not raw ed25519 keys?

We do, but they cannot be used here as they don't support loading the raw private key value (you can only generate a new random key or generate one from a RFC 8032 seed).

cli/wallet/file/file.go
kostko: Yeah these should definitely propagate errors, not sure how this got missed.

cli/wallet/file/file.go
Yawning: Argon2

cli/config/config.go
```diff
@@ -0,0 +1,168 @@
...
+	Wallets  Wallets  `mapstructure:"wallets"`
+}
+
+// Load loads the configuration structure from viper.
+func (cfg *Config) Load(v *viper.Viper) error {
+	cfg.viper = v
+	return v.Unmarshal(cfg)
+}
+
+// encode is needed because mapstructure cannot encode structs into maps recursively.
```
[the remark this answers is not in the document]
kostko: Yeah there's a bunch of issue reports upstream (https://github.com/mitchellh/mapstructure/issues/267, https://github.com/mitchellh/mapstructure/issues/166). It is not related to serialization formats as this happens without any serialization (transforming a typed struct into `map[string]interface{}`).

cli/wallet/file/file.go
kostko: Yeah that's ugly, I've switched this over so that there's now `ed25519-adr8` and `ed25519-raw` in the open. The number field is indeed something that is specific to ADR8 derivation though.

cli/cmd/wallet.go
kostko: The prompt has been changed to be more clear (also added repeat passphrase prompt).

cli/wallet/file/ed25519.go
pro-wh: would you be opposed to adding that import functionality to the memory signer?

cli/config/config.go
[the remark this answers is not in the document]
pro-wh: ah thanks

cli/cmd/wallet.go
pro-wh: nice, thanks

pro-wh (approved):

pro-wh:
thanks for the changes!

cli/wallet/file/ed25519.go
kostko: I wouldn't be, but I would rather avoid the need for this PR to depend on an unreleased version of Oasis Core. Can you raise an issue in Oasis Core (maybe also for exposing the chain context separator) and then once it is part of a stable release we can change it here.

cli/cmd/common/transaction.go
```diff
@@ -0,0 +1,290 @@
...
+	resultCh := make(chan sdkClient.DecodedEvent)
+	go func() {
+		defer close(resultCh)
+
+		for {
+			select {
+			case bev := <-ch:
+				for _, ev := range bev.Events {
+					if result := mapFn(ev); result != nil {
+						resultCh <- result
```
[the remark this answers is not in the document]
kostko: Good point, the whole `WaitForEvent` should return `interface{}` instead.

cli/cmd/common/transaction.go
```diff
@@ -0,0 +1,290 @@
...
+// WaitForEvent waits for a specific ParaTime event.
+func WaitForEvent(
+	ctx context.Context,
+	pt *config.ParaTime,
+	conn client.Connection,
+	decoder sdkClient.EventDecoder,
+	mapFn func(sdkClient.DecodedEvent) interface{},
+) <-chan sdkClient.DecodedEvent {
+	// Start watching events.
+	ch, err := conn.Runtime(pt).WatchEvents(ctx, []sdkClient.EventDecoder{decoder}, false)
```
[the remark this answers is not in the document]
kostko: They get closed after the context is cancelled.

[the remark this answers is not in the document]
kostko: I'll add a subcontext that will get cancelled as soon as an event is matched.

cli/cmd/common/transaction.go
```diff
@@ -0,0 +1,290 @@
...
+		if result != nil {
+			err = cbor.Unmarshal(meta.Result, result)
+			cobra.CheckErr(err)
+		}
+	default:
+		panic(fmt.Errorf("unsupported transaction kind: %T", tx))
+	}
+}
+
+// WaitForEvent waits for a specific ParaTime event.
```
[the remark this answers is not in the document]
kostko: Updated the comment to mention this.

cli/cmd/common/transaction.go
```diff
@@ -0,0 +1,290 @@
...
+	}()
+
+	return resultCh
+}
+
+func init() {
+	TransactionFlags = flag.NewFlagSet("", flag.ContinueOnError)
+	TransactionFlags.BoolVar(&txOffline, "offline", false, "do not perform any operations requiring network access")
+	TransactionFlags.Uint64Var(&txNonce, "nonce", 0, "override nonce to use")
+	TransactionFlags.Uint64Var(&txGasLimit, "gas-limit", 0, "override gas limit to use (disable estimation)")
```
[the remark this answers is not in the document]
kostko: Yeah, fixed.

client-sdk/go/modules/contracts/contracts.go
```diff
@@ -273,3 +273,31 @@ func (a *v1) DecodeEvent(event *types.Event) ([]client.DecodedEvent, error) {
...
+func NewUploadTx(fee *types.Fee, body *Upload) *types.Transaction {
+	// Compress code before upload.
+	var compressedCode bytes.Buffer
+	encoder := snappy.NewBufferedWriter(&compressedCode)
+	_, err := encoder.Write(body.Code)
+	if err != nil {
+		panic(err)
+	}
+	encoder.Close()
+	body.Code = compressedCode.Bytes()
```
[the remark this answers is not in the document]
kostko: True, refactored this.

cli/main.go
```diff
@@ -0,0 +1,7 @@
+package main
```
tjanez, line 1: Could you add an initial `README.md` to the `cli` directory and explain how people can build and start using the Oasis CLI?

cli/cmd/root.go
```diff
@@ -0,0 +1,79 @@
+package cmd
```
tjanez, line 1: Could you add a `version` sub-command or a `--version` top-level flag?

Yawning:
Bikeshedding this to hell and back is great and all, but it has utility functions that I need for other work.

cli/cmd/root.go
kostko: Added.

cli/main.go
kostko: Added something basic.

cli/go.mod
```diff
@@ -0,0 +1,83 @@
...
+	github.com/AlecAivazis/survey/v2 v2.3.2
+	github.com/adrg/xdg v0.4.0
+	github.com/mitchellh/mapstructure v1.4.2
+	github.com/oasisprotocol/curve25519-voi v0.0.0-20210716083614-f38f8e8b0b84
+	github.com/oasisprotocol/deoxysii v0.0.0-20200527154044-851aec403956
+	github.com/oasisprotocol/oasis-core/go v0.2103.6
+	github.com/oasisprotocol/oasis-sdk/client-sdk/go v0.1.0
+	github.com/olekukonko/tablewriter v0.0.5
+	github.com/shopspring/decimal v1.3.1
+	github.com/spf13/cobra v1.2.1
```
tjanez, line 17: Could you update cobra to 1.3.0? It has some completion sub-command improvements.

cli/main.go
tjanez: It looks very good, thanks!

cli/cmd/root.go
tjanez: Thanks!

.gitignore
```diff
@@ -24,3 +24,4 @@ coverage-e2e-*.txt
 examples/user-witness-flow/main
 tests/e2e/e2e
 tests/benchmark/benchmark
+cli/cli
```
tjanez, line 27: Also add `cli/oasis`

cli/cmd/network.go
```diff
@@ -0,0 +1,171 @@
+package cmd
```
tjanez, line 1: Could we add a (sub)command to `network` command or a new command that would show you the current default network and/or the current default ParaTime?

Or perhaps the default network could be shown via a `(default)` suffix (or similar) when running `oasis network list`?

cli/cmd/network.go
```diff
@@ -0,0 +1,171 @@
...
+			}
+			answers := struct {
+				Description string
+				Symbol      string
+				Decimals    uint8
+			}{}
+			err := survey.Ask(questions, &answers)
+			cobra.CheckErr(err)
+
+			err = cfg.Networks.Add(name, &config.Network{
```
tjanez, line 89: Could we validate user's input before asking survey questions?
I.e. running `net.Validate()` with just the given chain context and RPC so that users aren't asked those questions if we already know that the command will fail.

To prevent things like:
```
$ oasis network add dryrun 53852332637bacb61b91b6411ab4095168ba02a50be4c3f82448438826f dryrun.grpc.oasis.dev:443
? Description: bla
? Denomination symbol: DRY
? Denomination decimal places: 9
Error: malformed chain context: encoding/hex: odd length hex string
```

cli/cmd/network.go
```diff
@@ -0,0 +1,171 @@
...
+		Args:  cobra.ExactArgs(3),
+		Run: func(cmd *cobra.Command, args []string) {
+			cfg := config.Global()
+			name, chainContext, rpc := args[0], args[1], args[2]
+
+			// Ask user for some additional parameters.
+			questions := []*survey.Question{
+				{
+					Name:   "description",
+					Prompt: &survey.Input{Message: "Description:"},
```
tjanez, line 66: There is a bit of an inconsistency. In the `oasis network list` we use `NAME` and here we ask for `Description`.
I'm slightly more in favor of using `Name` here and perhaps `network_name` for the variables to avoid confusion with all other `Name` things in this part of the code :slightly_smiling_face:

cli/cmd/network.go
```diff
@@ -0,0 +1,171 @@
...
+			})
+			cobra.CheckErr(err)
+
+			err = cfg.Save()
+			cobra.CheckErr(err)
+		},
+	}
+
+	networkRmCmd = &cobra.Command{
+		Use:     "rm <name>",
```
tjanez, line 106: This doesn't need to be resolved in this PR, but could the shell completion be configured to list all the networks' names when one presses <kbd>Tab</kbd> after typing:
```
oasis network rm
```

.gitignore
kostko: Done.

cli/go.mod
kostko: Done.

cli/cmd/network.go
kostko: Yeah this should be done in a separate PR, feel free to file an issue once this is merged.

cli/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+To show the wallet's balance on the default network/paratime, run:
+
+```bash
+oasis accounts show
+```
+
+## Configuration
+
+All configuration is stored in the `$HOME/.config/oasis` directory.
```
tjanez, line 69:
```suggestion
All configuration is stored in the `$XDG_CONFIG_HOME/oasis` directory (defaults to `$HOME/.config/oasis`).
```

cli/cmd/root.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func initConfig() {
+	v := viper.New()
+
+	if cfgFile != "" {
+		// Use config file from the flag.
+		v.SetConfigFile(cfgFile)
+	} else {
+		const configFilename = "config.toml"
```
tjanez, line 39: I would perhaps go with `cli.toml` if we ever add more configuration files under `$XDG_CONFIG_HOME/oasis`?

cli/cmd/root.go
```diff
@@ -0,0 +1,79 @@
...
+		configDir := config.Directory()
+		configPath := filepath.Join(configDir, configFilename)
+
+		v.AddConfigPath(configDir)
+		v.SetConfigType("toml")
```
tjanez, line 44: Do you think TOML will be better for the users :slightly_smiling_face: ?
I'm a bit more in favor or YAML :slightly_smiling_face:

Yawning:
Since I'm blocked on this, I might as well request changes too:
 * Can you move cli/client to somewhere inside the actual client SDK so that people writing tooling aren't stuck either reimplementing it or pulling it in from the CLI.
 * Likewise, can you move cli/common/demonimation to somewhere inside the actual client SDK?

tjanez:
> Since I'm blocked on this, I might as well request changes too:
> 
>     * Can you move cli/client to somewhere inside the actual client SDK so that people writing tooling aren't stuck either reimplementing it or pulling it in from the CLI.
> 
>     * Likewise, can you move cli/common/demonimation to somewhere inside the actual client SDK?

Yeah and perhaps extract this in a separate commit and PR so it can be reviewed and merged earlier?

cli/cmd/network.go
kostko: So there are two things, "name" and "description". The name is an identifier used in various places to reference networks, paratimes etc. while "description" is something human readable that is also shown when signing a transaction.

cli/cmd/network.go
kostko: Done.

cli/cmd/root.go
kostko: Done.

kostko:
Moved some of the config, connection and helpers stuff to the Go Client SDK.

kostko:
I'll merge the PR and we can improve the CLI in subsequent PRs.

client-sdk/go/config/default.go
```diff
@@ -0,0 +1,80 @@
...
+
+// DefaultNetworks is the default config containing known networks.
+var DefaultNetworks = Networks{
+	Default: "mainnet",
+	All: map[string]*Network{
+		// Mainnet network parameters.
+		// See https://docs.oasis.dev/general/oasis-network/network-parameters.
+		"mainnet": {
+			ChainContext: "53852332637bacb61b91b6411ab4095168ba02a50be4c3f82448438826f23898",
+			RPC:          "grpc.oasis.dev:443",
```
pro-wh, line 11: resposting as this was lost in a file move:

        does the grpc-web proxy lets us dial it from the normal grpc client? or is there an additional thing on that server that handles normal grpc?

PR #600, opened 2021-11-02
runtime-sdk/contracts: CheckTx transactions should succeed

ptrus:


runtime-sdk/modules/contracts/src/lib.rs
```diff
@@ -475,6 +475,12 @@ impl<Cfg: Config> Module<Cfg> {
             params: &params,
         };
         let result = wasm::instantiate::<Cfg, C>(&mut exec_ctx, &contract, &body);
+
+        // Always return success in CheckTx, as we might not have up-to-date state.
```
kostko, line 479: This will skip estimating gas for processing the result?

runtime-sdk/modules/contracts/src/lib.rs
```diff
@@ -516,6 +522,12 @@ impl<Cfg: Config> Module<Cfg> {
             params: &params,
         };
         let result = wasm::call::<Cfg, C>(&mut exec_ctx, &contract, &body);
+
+        // Always return success in CheckTx, as we might not have up-to-date state.
```
kostko, line 526: Same as above re gas estimation.

runtime-sdk/modules/contracts/src/lib.rs
```diff
@@ -562,6 +574,12 @@ impl<Cfg: Config> Module<Cfg> {
         };
         // Pre-upgrade invocation must succeed for the upgrade to proceed.
         let result = wasm::pre_upgrade::<Cfg, C>(&mut exec_ctx, &contract, &body);
+
+        // Always return success in CheckTx, as we might not have up-to-date state.
```
kostko, line 578: This will skip estimating gas for processing the result and calling the post upgrade handler and result processing.

runtime-sdk/modules/contracts/src/lib.rs
ptrus: Hm so this should be something like:

```rust
        let result = match results::process_execution_result(ctx, result) {
            Ok(ok) => ok,
            Err(e) => {
                // Always return success in CheckTx, as we might not have up-to-date state.
                if ctx.is_check_only() {
                    return Ok(types::InstantiateResult::default());
                } else {
                    return Err(e);
                }
            }
        };
        if let Err(e) =
            results::process_execution_success::<Cfg, C>(ctx, &params, &contract, result)
        {
            // Always return success in CheckTx, as we might not have up-to-date state.
            if !ctx.is_check_only() {
                return Err(e);
            }
        }
```

right?

kostko: Yeah this looks quite ugly, could maybe make it nicer it with a closure that returns a result?

kostko: Yeah this looks quite ugly, could maybe make it nicer it with a closure that returns a result?

ptrus: Yeah agreed, pushed a version with a closure.

kostko: Nice!

kostko (approved):

PR #601, opened 2021-11-02
runtime-sdk: Support consensus denomination scaling

kostko:
TODO
* [x] More unit tests.
* [x] E2E tests.
* [x] Queries for retrieving the consensus module parameters (e.g. the scaling factor).
* [x] Support specifying `DenominationInfo` in accounts module.
* [x] Queries for retrieving the `DenominationInfo` in accounts module.

### DIFF
--- a/runtime-sdk/src/modules/consensus/mod.rs
+++ b/runtime-sdk/src/modules/consensus/mod.rs
@@ -1,7 +1,7 @@
 //! Consensus module.
 //!
 //! Low level consensus module for communicating with the consensus layer.
-use std::str::FromStr;
+use std::{convert::TryInto, str::FromStr};
 
 use thiserror::Error;
 
@@ -17,8 +17,8 @@ use oasis_core_runtime::{
 
 use crate::{
     context::{Context, TxContext},
-    module,
-    module::Module as _,
+    error, module,
+    module::{Module as _, Parameters as _},
     modules,
     modules::core::{Module as Core, API as _},
     types::{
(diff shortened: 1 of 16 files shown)


runtime-sdk/src/modules/consensus/mod.rs
```diff
@@ -237,6 +271,7 @@ impl API for Module {
         hook: MessageEventHookInvocation,
     ) -> Result<(), Error> {
         Self::ensure_consensus_denomination(ctx, amount.denomination())?;
+        let amount = Self::amount_to_consensus(ctx, amount.amount())?;
```
ptrus, line 274: So reclaim escrow actually accepts the number of shares to reclaim, I assume this shouldn't actually be scaled?

Maybe the argument should be renamed from `amount` to `shares` to make this a bit clearer. Hm it also shouldn't be of the denomination type probably?

Note: this method isn't used by anything yet, just exposes a wapper to emit the reclaim escrow message.

kostko: Oh, yeah that's right, the number of shares should probably just be a plain `u128`? Let me fix that.

kostko: Updated.

ptrus: Looks good now, thanks :+1:

runtime-sdk/src/modules/consensus_accounts/mod.rs
```diff
@@ -128,8 +128,9 @@ impl<Accounts: modules::accounts::API, Consensus: modules::consensus::API> API
     ) -> Result<(), Error> {
         if ctx.is_check_only() {
             // In case this is not check only this weight will be emitted from Cosnensus::withdraw
```
abukosek, line 130:
```suggestion
            // In case this is not check only this weight will be emitted from Consensus::withdraw
```

runtime-sdk/src/modules/consensus_accounts/mod.rs
```diff
@@ -163,8 +164,9 @@ impl<Accounts: modules::accounts::API, Consensus: modules::consensus::API> API
     ) -> Result<(), Error> {
         if ctx.is_check_only() {
             // In case this is not check only this weight will be emitted from Cosnensus::transfer
```
abukosek, line 166:
```suggestion
            // In case this is not check only this weight will be emitted from Consensus::transfer
```

runtime-sdk/src/modules/consensus/mod.rs
```diff
@@ -36,21 +36,38 @@ mod test;
...
+    #[error("consensus scaling factor set to zero")]
+    ZeroScalingFactor,
+}
+
 impl module::Parameters for Parameters {
-    type Error = ();
+    type Error = ParameterValidationError;
+
+    fn validate_basic(&self) -> Result<(), Self::Error> {
+        if self.consensus_scaling_factor == 0 {
```
ptrus, line 65: Should there also be a check for ensuring that the scaling factor is a power of 10? I guess there's technically nothing wrong with a non power of 10 scaling factor, but not sure if there's ever a reason for the weirdness.

Could also consider negative scaling factors, but this could also be added later if there's ever a need for it.

pro-wh (approved):
consensus module changes lg on my side

other changes pass our tests

pro-wh:
we had a discussion on a call about having a type to represent consensus token units. that might be nice, and we could have a single conversion function that does the denomination check and the scaling. as we noted in our call, we couldn't use TryFrom, because the conversion needs to look at the parameters. but I still like the idea

I brought up that maybe the consensus module should have parameters in the consensus units. counter argument was that we'd probably never call it without converting from runtime units anyway. eh fine

PR #613, opened 2021-11-04
Bump Oasis Core to 21.3.5 and Rust toolchain to nightly-2021-11-04

kostko:


ptrus (approved):
Seems like clippy has some new checks in the updated nightly that need to be fixed.

PR #626, opened 2021-11-09
ts-web/rt: playground make deposit/withdraw cross-account

pro-wh:
along with some fixes elsewhere to make this work

and a release of client-rt

### DIFF
--- a/runtime-sdk/src/modules/consensus_accounts/mod.rs
+++ b/runtime-sdk/src/modules/consensus_accounts/mod.rs
@@ -215,7 +215,12 @@ impl<Accounts: modules::accounts::API, Consensus: modules::consensus::API>
 
         // Signer.
         let signer = &ctx.tx_auth_info().signer_info[0];
-        Consensus::ensure_compatible_tx_signer(ctx)?;
+        if body.to.is_none() {
+            // If no `to` field is specified, i.e. withdrawing to the transaction sender's account,
+            // only allow the consensus-compatible single-Ed25519-key signer type. Otherwise, the
+            // tokens would get stuck in an account that you can't sign for on the consensus layer.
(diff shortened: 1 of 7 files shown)


runtime-sdk/src/modules/consensus_accounts/mod.rs
```diff
@@ -215,7 +215,12 @@ impl<Accounts: modules::accounts::API, Consensus: modules::consensus::API>
 
         // Signer.
         let signer = &ctx.tx_auth_info().signer_info[0];
-        Consensus::ensure_compatible_tx_signer(ctx)?;
+        if body.to.is_none() {
+            // If no `to` field is specified, i.e. withdrawing to the transaction sender's account,
+            // only allow the consensus-compatible single-Ed25519-key signer type. Otherwise, the
+            // tokens would get stuck in an account that you can't sign for on the consensus layer.
+            Consensus::ensure_compatible_tx_signer(ctx)?;
```
kostko, line 222: Would be nice to have a unit test that covers this case.

kostko (approved):

ptrus (approved):

kostko:
Added tests and fixed ts-web lint.

pro-wh:
thanks for the fixes!

runtime-sdk/src/modules/consensus_accounts/mod.rs
pro-wh: yeah. let me check if/how we currently cover this check

kostko: I already added the tests ;-)

pro-wh: oh great, thanks!

PR #627, opened 2021-11-09
modules/consensus_accounts: Add Deposit/Withdraw events

### LINKED ISSUE #624: Add events for consensus accounts operations' success/fail
we had previously watched for the mint/burn event, which isn't good for detecting errors---the event simply wouldn't come. let's add an error event to convey if something went wrong and what/how. for symmetry, let's also add a success event.

I think this will also allow us to associate the event with the transaction more explicitly. currently we had been matching on addresses and amount, which is no as rigorous

kostko:
Fixes #624

### DIFF
--- a/runtime-sdk/src/modules/consensus_accounts/types.rs
+++ b/runtime-sdk/src/modules/consensus_accounts/types.rs
@@ -1,5 +1,5 @@
 //! Consensus module types.
-use crate::types::{address::Address, token};
+use crate::types::{address::Address, message::MessageEvent, token};
 
 /// Deposit into runtime call.
 /// Transfer from consensus staking to an account in this runtime.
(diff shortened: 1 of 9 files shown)


runtime-sdk/src/modules/consensus_accounts/types.rs
```diff
@@ -1,4 +1,6 @@
 //! Consensus module types.
+use oasis_core_runtime::consensus::roothash::MessageEvent;
```
ptrus, line 2: Maybe nicer to import the re-exported type in `crate::types::message::MessageEvent`

ptrus (approved):

runtime-sdk/src/modules/consensus_accounts/types.rs
kostko: Indeed, fixed!

PR #629, opened 2021-11-10
modules/consensus_accounts: substitute nonce for id in events

pro-wh:
with some additional changes to use these events in the consensus playground

### DIFF
--- a/client-sdk/ts-web/rt/playground/src/consensus.js
+++ b/client-sdk/ts-web/rt/playground/src/consensus.js
@@ -82,8 +82,9 @@ export const playground = (async function () {
         const davePrivExpected = await oasis.hash.hash(
             oasis.misc.fromString('oasis-runtime-sdk/test-keys: dave'),
         );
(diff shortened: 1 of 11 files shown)


client-sdk/ts-web/rt/playground/src/consensus.js
```diff
@@ -25,6 +25,51 @@ const nic = new oasis.client.NodeInternal('http://localhost:42280');
...
+                ) {
+                    // Block doesn't exist yet. Wait and fetch again.
+                    continue;
+                }
+                throw e;
+            }
+            console.log('polled block', nextRound);
+            if (events) {
+                for (const e of events) {
+                    if (!cb(e)) break poll_blocks;
```
pro-wh, line 63: this works for a test, but there's no mechanism to cancel this if the tx submission fails. we probably need more weird cancellation flags and stuff

kostko (approved):

PR #635, opened 2021-11-11
ts-web/core: playground show how to avoid eval

### LINKED ISSUE #608: protobufjs uses eval
This adds evil `eval(` call https://github.com/protobufjs/protobuf.js/blob/d13d5d5/lib/inquire/index.js#L12 

we block it with CSP so this might result in bugs

_Originally posted by @lukaw3d in https://github.com/oasisprotocol/oasis-wallet-web/pull/348#discussion_r742424658_

pro-wh:
protobufjs's 'inquire' library will always fail

not sure if we should really merge this:

- it's really downstream users that will need to do this
- this particular 'playground' page runs in dev mode with the 'eval devtool' anyway

cc #608


pro-wh:
let's merge this so that the "playground" areas represent the best configuration we know of.

kostko (approved):

PR #639, opened 2021-11-12
runtime-sdk: Add confidential store

### LINKED ISSUE #333: Confidential store
We should add a confidential store (see [`runtime-sdk/src/storage`](https://github.com/oasisprotocol/oasis-sdk/tree/main/runtime-sdk/src/storage)) that (probably) uses Deoxys-II to transparently encrypt/decrypt keys and values. Things to consider:
* What transformation to use for encrypting keys?
* How to make it easier to derive nonces when encrypting values? Possibly using some information from the transaction context like round number and transaction index within the block?

jberci:
Closes #333


runtime-sdk/src/storage/confidential.rs
```diff
@@ -0,0 +1,252 @@
...
+        ret.extend_from_slice(nonce);
+        ret.extend_from_slice(enc_key);
+        ret
+    }
+
+    fn make_key(&self, plain_key: &[u8]) -> (Nonce, Vec<u8>) {
+        let mut nonce = [0u8; NONCE_SIZE];
+        let plain_hash = Hash::digest_bytes(plain_key);
+        nonce.copy_from_slice(plain_hash.truncated(NONCE_SIZE));
+        let enc_key = self.deoxys.seal(&nonce, plain_key.to_vec(), nonce.to_vec());
```
kostko, line 65: Why is nonce being used as AD here (and elsewhere)? You should probably use `vec![]` for additional data for now.

runtime-sdk/src/storage/confidential.rs
```diff
@@ -0,0 +1,252 @@
...
+
+    fn make_value(&self, value: &[u8], nonce: &Nonce) -> Vec<u8> {
+        self.deoxys.seal(nonce, value.to_vec(), nonce.to_vec())
+    }
+}
+
+impl<S: Store> Store for ConfidentialStore<S> {
+    fn get(&self, key: &[u8]) -> Option<Vec<u8>> {
+        let (nonce, inner_key) = self.make_key(key);
+        match self.inner.get(&inner_key) {
```
kostko, line 104: Could use a `.map` instead of this match.

runtime-sdk/modules/contracts/src/store.rs
```diff
@@ -12,7 +13,18 @@ pub fn for_instance<'a, C: Context>(
     ctx: &'a mut C,
     instance_info: &types::Instance,
     store_kind: StoreKind,
-) -> Result<impl Store + 'a, Error> {
+) -> Result<Box<dyn Store + 'a>, Error> {
+    let key_pair: Option<KeyPair> = if let StoreKind::Confidential = store_kind {
+        let kmgr_client = ctx.key_manager().ok_or(Error::Unsupported)?;
+        let kid = keymanager::get_key_pair_id(&[&instance_info.id.to_storage_key()]);
```
kostko, line 19: This shouldn't directly use the instance ID, it should be namespaced (use some constant base context as the first element).

kostko (commented):
Would also be great to have some benchmarks (using `cargo bench`) to show overhead compared to plain store.

kostko (commented):
It would be nicer if the key manager was a trait and the mocked version would be completely separate.

runtime-sdk/modules/contracts/src/store.rs
```diff
@@ -2,17 +2,36 @@
...
+) -> Result<Box<dyn Store + 'a>, Error> {
+    let confidential_key: Option<StateKey> = if let StoreKind::Confidential = store_kind {
+        let kmgr_client = ctx.key_manager().ok_or(Error::Unsupported)?;
+        let kid = keymanager::get_key_pair_id(&[
+            CONFIDENTIAL_STORE_KEY_PAIR_ID_CONTEXT_BASE,
+            &instance_info.id.to_storage_key(),
+        ]);
+        let kp = kmgr_client
+            .get_or_create_keys(kid)
+            .map_err(|err| Error::ExecutionFailed(err.into()))?;
```
kostko, line 29: This should result in a batch abort (e.g. define a new error variant that has the `sdk_error(abort)` attribute and use that).

runtime-sdk/src/storage/confidential.rs
```diff
@@ -0,0 +1,400 @@
...
+    fn pack_key(&self, enc_key: &[u8], nonce: &Nonce) -> Vec<u8> {
+        let mut ret = Vec::with_capacity(nonce.len() + enc_key.len());
+        ret.extend_from_slice(nonce);
+        ret.extend_from_slice(enc_key);
+        ret
+    }
+
+    fn make_key(&self, plain_key: &[u8]) -> (Nonce, Vec<u8>) {
+        let mut nonce = [0u8; NONCE_SIZE];
+        let plain_hash = Hash::digest_bytes(plain_key);
```
kostko, line 61: This could probably leak some information about `plain_key`. Some better options:
* Use a value derived from something like the current `round || tx index || internal counter`, where the internal counter is incremented every time you need a new nonce within the same instance. You can get the first two from context and pass them to the constructor.
* Derive a "nonce key" from `key` and a KDF when initializing the instance in `new_with_key`. Use `H(nonce_key || plain_key)` here. The downside is that this will result in the same nonce for the same `plain_key`.

@Yawning any thoughts?

kostko: Ooops, the first scheme wouldn't work for keys, only for values.

Yawning: By virtue of it being the key rather than the value, leaking "keys are identical" is somewhat hard, but the keypair used to initialize the store is presumably providing domain separation (at the contract level?), so I don't think that's a big deal...

runtime-sdk/modules/contracts/src/store.rs
```diff
@@ -13,12 +13,26 @@ use crate::{state, types, Error, MODULE_NAME};
 pub const CONFIDENTIAL_STORE_KEY_PAIR_ID_CONTEXT_BASE: &[u8] =
     b"oasis-runtime-sdk/contracts: state";
 
+const CONTEXT_CONFIDENTIAL_STORE_TX_INSTANCE_COUNT: &str = "oasis-runtime-sdk/Context store instance count";
```
kostko, line 16: If we have this do we even need the tx index given that this will already be per-block and will increment on each access?

runtime-sdk/modules/contracts/src/store.rs
```diff
@@ -13,12 +13,26 @@ use crate::{state, types, Error, MODULE_NAME};
...
+
 /// Create a contract instance store.
+///
+/// Confidential stores are only available in transaction contexts.
 pub fn for_instance<'a, C: Context>(
     ctx: &'a mut C,
     instance_info: &types::Instance,
     store_kind: StoreKind,
 ) -> Result<Box<dyn Store + 'a>, Error> {
+    let instance_count: usize = *ctx.value(CONTEXT_CONFIDENTIAL_STORE_TX_INSTANCE_COUNT).or_default();
```
kostko, line 26: Should be ignored for non-confidential setups (e.g. move it down to where the confidential store is instantiated).

runtime-sdk/src/storage/confidential.rs
```diff
@@ -0,0 +1,422 @@
...
+    }
+
+    fn make_value(&mut self, plain_value: &[u8]) -> (Nonce, Vec<u8>) {
+        let mut nonce = [0u8; NONCE_SIZE];
+
+        self.nonce_counter += 1;
+        let hash = Hash::digest_bytes_list(&[
+            self.base_value_prefix.as_slice(),
+            self.nonce_counter.to_le_bytes().as_slice(),
+            plain_value,
```
kostko, line 103: No need to put value here given that the above should already be fairly unique. If you put the value here this would again require the "nonce key treatment".

runtime-sdk/src/storage/confidential.rs
```diff
@@ -0,0 +1,421 @@
...
+        },
+        storage::mkvs,
+    },
+    storage::Store,
+};
+
+type Nonce = [u8; NONCE_SIZE];
+type Kdf = Hmac<Sha512Trunc256>;
+
+pub struct EncryptionKey([u8; KEY_SIZE]);
```
kostko, line 25: This seems unused, remove?

kostko (commented):
Looks good, some minor comments.

Can you also add unit tests where the underlying storage is corrupted to make sure that the right thing happens. E.g. store something using the confidential store and then corrupt it and try to perform a get.

Also could you add a unit test for the iterator?

runtime-sdk/modules/contracts/src/store.rs
```diff
@@ -2,17 +2,61 @@
...
+
+/// Confidential store key pair ID domain separation context base.
+pub const CONFIDENTIAL_STORE_KEY_PAIR_ID_CONTEXT_BASE: &[u8] =
+    b"oasis-runtime-sdk/contracts: state";
+
+const CONTEXT_KEY_CONFIDENTIAL_STORE_INSTANCE_COUNT: &str = "contracts.ConfidentialStoreCounter";
 
 /// Create a contract instance store.
+///
+/// Confidential stores are only available in transaction contexts.
```
kostko, line 20: What happens outside a transaction context?

kostko (approved):

kostko (approved):

runtime-sdk/src/storage/confidential.rs
```diff
@@ -0,0 +1,617 @@
...
+    nonce_key: Zeroizing<Vec<u8>>,
+}
+
+impl<S: Store> ConfidentialStore<S> {
+    /// Create a new confidential store with the given keypair.
+    pub fn new_with_key(inner: S, key: [u8; KEY_SIZE], value_context: &[&[u8]]) -> Self {
+        let actual_key = Zeroizing::new(key);
+
+        // Derive a nonce key for nonces used to encrypt storage keys in the store:
+        // nonce_key = KDF(key)
```
Yawning, line 64: Hm, HMAC as a KDF isn't amazing, but it's probably ok?  This is only used for nonce derivation right?

runtime-sdk/src/storage/confidential.rs
```diff
@@ -0,0 +1,617 @@
...
+
+        let hash = Hash::digest_bytes(&nonce_src);
+        nonce.copy_from_slice(hash.truncated(NONCE_SIZE));
+
+        let enc_key = self.deoxys.seal(&nonce, plain_key.to_vec(), vec![]);
+        let key = self.pack_nonce_slice(&nonce, &enc_key);
+        (nonce, key)
+    }
+
+    fn make_value(&mut self, plain_value: &[u8]) -> (Nonce, Vec<u8>) {
```
Yawning, line 107: This seems kind of overkill to mask "the code accessed the same key", but I can't think of a nice way to handle this off the top of my head.

PR #673, opened 2021-12-02
ts-web/rt: add contracts module

pro-wh:
writing up the typescript bindings. need to port a lot of types :dizzy_face:

### DIFF
--- a/client-sdk/ts-web/rt/src/types.ts
+++ b/client-sdk/ts-web/rt/src/types.ts
@@ -674,3 +674,17 @@ export interface ContractsCustomQuery {
      */
     data: Uint8Array;
 }
+
+/**
+ * An event emitted from a contract, wrapped to include additional metadata.
+ */
+export interface ContractsContractEvent {
+    /**
+     * Identifier of the instance that emitted the event.
+     */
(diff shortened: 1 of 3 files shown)


pro-wh:
ok I did a bunch of hacky regex replaces from rust source code to port over the types

client-sdk/ts-web/rt/src/types.ts
```diff
@@ -433,3 +433,244 @@ export interface ResultEnvelopeX25519DeoxysII {
...
+    id: oasis.types.longnum;
+    /**
+     * Storage key.
+     */
+    key: Uint8Array;
+}
+
+export interface ContractsInstanceStorageQueryResult {
+    /**
+     * Storage value or `None` if key doesn't exist.
```
pro-wh, line 627: note: we transfer comments verbatim, so mentioning Rust's `` `None` `` is intended

pro-wh:
hm the contracts module has its own way of doing events so that contract codes can get their own tag keys. for now all I'm doing is adding a ContractEvent struct

kostko (approved):

PR #675, opened 2021-12-02
Fix emitting multiple of events of the same type

ptrus:
Currently, the way sdk-events are mapped into tags causes the events of the same type to override and only the last emitted event is actually emitted as a transaction tag. 

For example, currently only the last emitted EVM log is emitted as a transaction tag and others are lost  https://github.com/oasisprotocol/oasis-sdk/blob/f5b7a5429342770dbf0ea464854cd9be49d03bd1/runtime-sdk/modules/evm/src/backend.rs#L250-L257

This changes the way events are mapped into transaction tags. The tag value is now an array of events.

Clients:
- [x] `go` client library is updated to internally handle the change, there should be no changes for library users of `WatchEvents/GetEvents` once the sdk is updated
- [x] `typescript` client library is updated internally to handle the change, there should be no changes for library users once updating the sdk
- [x] add a test

### DIFF
--- a/runtime-sdk/modules/contracts/src/results.rs
+++ b/runtime-sdk/modules/contracts/src/results.rs
@@ -1,5 +1,5 @@
 //! Processing of execution results.
-use std::convert::TryInto;
+use std::{collections::BTreeMap, convert::TryInto};
 
 use oasis_contract_sdk_types::{
     event::Event,
@@ -9,7 +9,7 @@ use oasis_contract_sdk_types::{
 use oasis_runtime_sdk::{
     context::{BatchContext, Context, TxContext},
     dispatcher,
-    event::tag_for_event,
+    event::etag_for_event,
     modules::core::{self, API as _},
     types::{token, transaction, transaction::CallerAddress},
 };
@@ -57,7 +57,7 @@ fn process_events<C: TxContext>(
 ) -> Result<(), Error> {
     // Transform contract events into tags using the SDK scheme.
     for event in events {
-        ctx.emit_tag(tag_for_event(
+        ctx.emit_etag(etag_for_event(
             &if event.module.is_empty() {
                 format!("{}.{}", MODULE_NAME, contract.code_info.id.as_u64())
             } else {
@@ -69,7 +69,7 @@ fn process_events<C: TxContext>(
                 )
             },
             event.code,
-            cbor::to_vec(ContractEvent {
+            cbor::to_value(ContractEvent {
                 id: contract.instance_info.id,
                 data: event.data,
             }),
@@ -144,7 +144,7 @@ fn process_subcalls<Cfg: Config, C: TxContext>(
                 let remaining_messages = ctx.remaining_messages();
 
                 // Execute a transaction in a child context.
-                let (result, gas, tags, messages) = ctx.with_child(ctx.mode(), |mut ctx| {
+                let (result, gas, etags, messages) = ctx.with_child(ctx.mode(), |mut ctx| {
                     // Generate an internal transaction.
                     let tx = transaction::Transaction {
                         version: transaction::LATEST_TRANSACTION_VERSION,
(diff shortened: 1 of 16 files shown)


client-sdk/ts-web/rt/playground/src/consensus.js
```diff
@@ -213,21 +213,23 @@ export const playground = (async function () {
         let depositEvent = null;
         const depositEventVisitor = new oasisRT.event.Visitor([
             oasisRT.consensusAccounts.moduleEventHandler({
-                [oasisRT.consensusAccounts.EVENT_DEPOSIT_CODE]: (e, depositEv) => {
-                    console.log('polled deposit event', depositEv);
-                    const eventFromBech32 = oasis.staking.addressToBech32(depositEv.from);
-                    if (eventFromBech32 !== addrAliceBech32) {
-                        console.log('address mismatch');
-                        return;
+                [oasisRT.consensusAccounts.EVENT_DEPOSIT_CODE]: (e, depositEvs) => {
```
kostko, line 216: Could this be changed such that the changes wouldn't be needed (e.g. that the visitor/handler would handle this and just call the function multiple times instead of requiring the developer to always iterate over events)?

runtime-sdk/modules/contracts/src/results.rs
```diff
@@ -204,7 +205,7 @@ fn process_subcalls<Cfg: Config, C: TxContext>(
 
                 // Forward any emitted tags.
                 for tag in tags {
-                    ctx.emit_tag(tag);
+                    ctx.emit_tag(tag.key, tag.value.into());
```
kostko, line 208: This doesn't seem like it will do the right thing (e.g. forward the value unchanged) because there is an interim conversion into a `Tag`. We can make `commit` return something else (e.g. a `Vec<(Vec<u8>, cbor::Value)>`) as well to handle this case?

client-sdk/ts-web/rt/playground/src/consensus.js
ptrus: Done, was actually easier than the previous change

runtime-sdk/modules/contracts/src/results.rs
ptrus: Yeah i think it makes sense, will also need to move the converting into tags into the dispatcher

kostko: Given that this tuple now appears in multiple places we could also add a type like `EventTag` that would have the key and CBOR value pair.

runtime-sdk/src/context.rs
```diff
@@ -147,14 +144,22 @@ pub trait Context {
     fn emit_event<E: Event>(&mut self, event: E);
 
     /// Emits a tag.
-    fn emit_tag(&mut self, tag: Tag);
+    fn emit_etag(&mut self, etag: EventTag);
+
+    /// Emits evnet tags.
```
kostko, line 149:
```suggestion
    /// Emits event tags.
```

runtime-sdk/src/context.rs
```diff
@@ -289,7 +296,7 @@ impl<'a, R: runtime::Runtime, S: NestedStore> RuntimeBatchContext<'a, R, S> {
             key_manager,
             logger: get_logger("runtime-sdk")
                 .new(o!("ctx" => "dispatch", "mode" => Into::<&'static str>::into(&mode))),
-            block_tags: Tags::new(),
+            block_etags: BTreeMap::new(),
```
kostko, line 299: Should this be `EventTags::new()` even thought the type is currently the same. Also below.

runtime-sdk/src/event.rs
```diff
@@ -1,5 +1,8 @@
 //! Event types for runtimes.
-use oasis_core_runtime::transaction::tags::Tag;
+
```
kostko, line 2: Remove empty line.

kostko (commented):
Looks good! Some minor nits.

kostko (approved):

PR #677, opened 2021-12-03
runtime-sdk: Restrict queries in secure builds

kostko:


runtime-sdk/src/keymanager.rs
```diff
@@ -146,15 +170,6 @@ impl<'a> KeyManagerClientWithContext<'a> {
     }
 }
 
-impl<'a> Clone for KeyManagerClientWithContext<'a> {
```
pro-wh, line 149: :information_source: this is now derived

runtime-sdk/src/dispatcher.rs
```diff
@@ -365,8 +366,19 @@ impl<R: Runtime> Dispatcher<R> {
...
-                _ => R::Modules::dispatch_query(ctx, method, args)
-                    .ok_or_else(|| modules::core::Error::InvalidMethod(method.into()))?,
+                _ => {
+                    // Even though queries are not allowed to perform any private key ops, we
+                    // reduce the attack surface by completely disallowing them in secure builds.
+                    //
+                    // If one needs to perform runtime queries, she should use a regular build of
+                    // the runtime for that.
+                    if BUILD_INFO.is_secure {
+                        return Err(modules::core::Error::InvalidMethod(method.into()).into());
```
pro-wh, line 376: might be nice to have a separate error message for this condition. but I guess it's not critical for this

pro-wh (commented):
was there any discussion around this? I see that this PR branch says it's a fix.

kostko:
> was there any discussion around this? I see that this PR branch says it's a fix.

The discussion can happen in the PR. This is reducing the attack surface of secure builds that need to protect confidentiality as otherwise careless runtime modules could potentially leak private stuff via queries. Thoughts?

runtime-sdk/src/dispatcher.rs
kostko: Agreed, let's add a separate variant.

runtime-sdk/src/keymanager.rs
kostko: Yeah not sure why it wasn't before.

pro-wh:
I imagine an alternative to allowing queries in secure builds is that the runtime could make an unencrypted area of storage computed from private values during actual transactions. Then service queries from there. There's no quota or authentication for queries, so putting out the raw data would be pretty close.

I wonder if there are use cases where you'd want to enforce certain queries being run on confidential data. Oasis Labs is doing some work on differential privacy, but that's not an exact fit for queries, because there's also a privacy budget update involved. Maybe in these cases it'll be possible to precompute answers to some queries and store those instead of raw data.

pro-wh (approved):
alright

we can think about reenabling it if a desire comes up later

PR #738, opened 2022-01-15
(experimental) tests: check priority order in txgen

pro-wh:
> Error: root: failed to run scenario: root: panic caught running scenario: priority correct 3: goroutine 1 [running]:

which is to say in the txgen test, stuff worked out with the transactions in order


pro-wh:
let's go with #748

PR #748, opened 2022-01-18
tests/e2e: Add test for transaction ordering by priority

kostko:
Inspired by #738.

Also bumps Oasis Core to 21.3.8 as otherwise the test wouldn't pass due to the query returning incorrect results.


pro-wh (approved):
ya

ptrus (approved):

PR #758, opened 2022-01-24
Introspection endpoint for Core module

### LINKED ISSUE #652: Add introspection query
The core module should support some form of an introspection query that would return:
* Version of the runtime.
* Version of the SDK.
* A list of all the supported modules and their versions.
* Possibly some form of method and/or type introspection.

mitjat:
Resolves #652 .

Adds a new `core.RuntimeInfo` query to the `core` module which provides basic information about the runtime and the modules it contains. See the bottom of `core/test.rs` for sample output.

### DIFF
--- a/client-sdk/go/modules/core/types.go
+++ b/client-sdk/go/modules/core/types.go
@@ -1,8 +1,9 @@
 package core
 
 import (
+	"github.com/oasisprotocol/oasis-core/go/common/cbor"
 	"github.com/oasisprotocol/oasis-core/go/common/quantity"
-
+	"github.com/oasisprotocol/oasis-core/go/common/version"
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/types"
 )
 
@@ -49,3 +50,40 @@ type GasUsedEvent struct {
 type Event struct {
 	GasUsed *GasUsedEvent
 }
+
+// RuntimeInfoResponse is the response of the core.RuntimeInfo query
+// and provides basic introspection information about the runtime.
+type RuntimeInfoResponse struct {
+	RuntimeVersion *version.Version `json:"runtime_version"`
+	// StateVersion is the version of the schema used by the runtime for keeping its state.
+	StateVersion uint32 `json:"state_version"`
+	// Modules are the SDK modules that comprise this runtime.
+	Modules map[string]ModuleInfo `json:"modules"`
+}
+
+// ModuleInfo is the information about a single module within the runtime.
+type ModuleInfo struct {
+	// Version is the version of the module.
+	Version uint32 `json:"version"`
+	// Params are the initial parameters of the module.
+	Params cbor.RawMessage `json:"params"`
+	// Methods are the RPC methods exposed by the module.
+	Methods []MethodHandlerInfo `json:"methods"`
+}
+
+// MethodHandlerInfo describes a single RPC.
+type MethodHandlerInfo struct {
+	// Name is the name of the RPC.
+	Name string `json:"name"`
+	// Kind is the kind of the RPC.
+	Kind methodHandlerKind `json:"kind"`
+}
+
+type methodHandlerKind string
+
+// These constants represent the kinds of methods that handlers handle.
+const (
+	MethodHandlerKindCall          methodHandlerKind = "call"
+	MethodHandlerKindQuery         methodHandlerKind = "query"
+	MethodHandlerKindMessageResult methodHandlerKind = "message_result"
+)
--- a/client-sdk/ts-web/rt/playground/src/index.js
+++ b/client-sdk/ts-web/rt/playground/src/index.js
@@ -285,6 +285,47 @@ export const playground = (async function () {
         console.log('ok');
     }
 
+    // Try introspection
+    {
+        const info = await coreWrapper.queryRuntimeInfo().query(nic);
+        console.log('runtime info', info);
+
+        // Check module presence
+        const accts = info.modules['accounts'];
+        if (accts === undefined) {
+            throw new Error('accounts module was expected in RuntimeInfo response');
+        }
+
+        // Check deserialization of lists and enums
+        if (
+            !accts.methods.some(
+                (m) =>
+                    m.name === 'accounts.Transfer' &&
+                    m.kind === oasisRT.core.METHODHANDLERKIND_CALL,
+            )
+        ) {
+            throw new Error(
+                'accounts.Transfer call was expected in accounts module; instead, got the following methods: ' +
+                    JSON.stringify(accts.methods),
+            );
+        }
+
+        // Check deserialization of arbitrary objects (= module params)
+        try {
+            // If we can log it, we can access it; success
+            console.log(
+                `tx transfer gas cost in accounts module is ${accts.params.gas_costs.tx_transfer}`,
+            );
+        } catch (e) {
+            throw new Error(
+                `Couldn't find gas_costs.tx_transfer in params of the accounts module. Params were ${JSON.stringify(
+                    accts.params,
+                )}`,
+            );
+        }
+        console.log('ok');
+    }
+
     // Try the rewards parameters.
     {
         console.log('query rewards parameters');
--- a/client-sdk/ts-web/rt/src/types.ts
+++ b/client-sdk/ts-web/rt/src/types.ts
@@ -26,6 +26,30 @@ export interface CoreGasUsedEvent {
 }
 
 /**
+ * Response to the RuntimeInfo query.
+ */
+export interface CoreRuntimeInfoQueryResponse {
+    runtime_version: oasis.types.Version;
+    state_version: number;
+    modules: {[key: string]: CoreModuleInfo};
+}
+
+/**
+ * Metadata for an individual module within the runtime.
+ */
+export interface CoreModuleInfo {
+    version: number;
+    params: any;
+    methods: CoreMethodHandlerInfo[];
+}
+
+export interface CoreMethodHandlerInfo {
+    name: string;
+    // Keep these in sync with the `METHODHANDLERKIND_*` constants.
+    kind: 'call' | 'query' | 'message_result';
+}
+
+/**
  * Caller address.
  */
 export interface CallerAddress {
@@ -196,7 +220,7 @@ export interface AuthProof {
      * A flag to use module-controlled decoding. The string is an encoding scheme name that a
      * module must handle. When using this variant, the scheme name must not be empty.
      */
-    module?: String;
+    module?: string;
 }
 
 /**
--- a/runtime-sdk-macros/src/method_handler_derive.rs
+++ b/runtime-sdk-macros/src/method_handler_derive.rs
@@ -1,6 +1,7 @@
 use crate::{emit_compile_error, generators as gen};
 use proc_macro2::TokenStream;
 use quote::quote;
+use syn::parse_quote;
 
 /// Given an `impl MyModule` block, produces an `impl MethodHandler for MyModule`.
 /// See `sdk_derive()` in lib.rs for details.
@@ -189,12 +190,37 @@ pub fn derive_method_handler(impl_block: syn::ItemImpl) -> TokenStream {
         }
     };
 
+    let supported_methods_impl = {
+        let (handler_names, handler_kinds): (Vec<syn::Expr>, Vec<syn::Path>) = handlers
+            .iter()
+            .filter_map(|h| h.handler.as_ref())
+            // `prefetch` is an implementation detail of `call` handlers, so we don't list them
+            .filter(|h| h.attrs.kind != HandlerKind::Prefetch)
+            .map(|h| (h.attrs.rpc_name.clone(), h.attrs.kind.as_sdk_ident()))
+            .unzip();
+        if handler_names.is_empty() {
+            quote! {}
+        } else {
+            quote! {
+                fn supported_methods() -> Vec<core_types::MethodHandlerInfo> {
+                    vec![ #(
+                        core_types::MethodHandlerInfo {
+                            kind: #handler_kinds,
+                            name: #handler_names.to_string(),
+                        },
+                    )* ]
+                }
+            }
+        }
+    };
+
     gen::wrap_in_const(quote! {
         use #sdk_crate::{
           self as sdk,
           cbor,
           error::Error as _,
           module::{DispatchResult, CallResult},
+          modules::core::types as core_types,
           types::message::MessageResult
         };
 
@@ -205,6 +231,7 @@ pub fn derive_method_handler(impl_block: syn::ItemImpl) -> TokenStream {
             #dispatch_call_impl
             #dispatch_query_impl
             #dispatch_message_result_impl
+            #supported_methods_impl
         }
 
         impl#module_generics #module_ty {
@@ -236,6 +263,21 @@ enum HandlerKind {
     Query,
     MessageResult,
     Prefetch,
+}
+
+impl HandlerKind {
+    fn as_sdk_ident(&self) -> syn::Path {
+        match self {
+            HandlerKind::Call => parse_quote!(core_types::MethodHandlerKind::Call),
+            HandlerKind::Query => parse_quote!(core_types::MethodHandlerKind::Query),
+            HandlerKind::MessageResult => {
+                parse_quote!(core_types::MethodHandlerKind::MessageResult)
+            }
+            HandlerKind::Prefetch => {
+                unimplemented!("prefetch cannot be expressed in core::types::MethodHandlerKind")
+            }
+        }
+    }
 }
 
 #[derive(Debug, Clone, PartialEq)]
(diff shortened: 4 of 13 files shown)


client-sdk/go/go.sum
```diff
@@ -24,77 +24,119 @@ cloud.google.com/go v0.83.0/go.mod h1:Z7MJUsANfY0pYPdw0lbnivPx4/vhy/e2FEkSkF7vAV
 cloud.google.com/go v0.84.0/go.mod h1:RazrYuxIK6Kb7YrzzhPoLmCVzl7Sup4NrbKPg8KHSUM=
 cloud.google.com/go v0.87.0/go.mod h1:TpDYlFy7vuLzZMMZ+B6iRiELaY7z/gJPaqbMx6mlWcY=
 cloud.google.com/go v0.90.0/go.mod h1:kRX0mNRHe0e2rC6oNakvwQqzyDmg57xJ+SZU1eT2aDQ=
+cloud.google.com/go v0.93.3 h1:wPBktZFzYBcCZVARvwVKqH1uEj+aLXofJEtrb4oOsio=
```
ptrus: Running `go mod tidy` should remove these unneeded checksums

runtime-sdk/src/module.rs
```diff
@@ -445,6 +445,46 @@ impl InvariantHandler for Tuple {
...
 
+/// Info handler.
+pub trait ModuleInfoHandler {
+    /// Reports info about the module(s).
+    fn module_info<C: Context>(_ctx: &mut C) -> BTreeMap<String, modules::core::types::ModuleInfo>;
+
+    /// Lists the names of all RPC methods exposed by this module. The result is informational
+    /// only; a return value of None means that the implementor does not care to list the methods,
+    /// or the implementor is a tuple of modules.
+    fn list_methods() -> Option<Vec<String>> {
```
mitjat, line 456: As a first/naive/lazy implementation, I was going to allow modules to override this method _if_ they feel like providing method introspection. But that doesn't work because (stable) Rust does not support method overrides.

Options:
- Do not provide method introspection.
- Go with current implementation, add `#![feature(specialization)]` to crate attributes to allow overrides. (I don't like this, it's kludgy anyway because it doesn't enforce the result list_methods() to be in sync with reality.)
- Implement macros that would require one to annotate call handlers like so:
  ```
  #[derive(MethodHandler)]
  pub struct Module {
    #[query_handler("core.MinGasPrice")]
    fn query_min_gas_price();

    #[call_handler("core.FooCall")]
    fn call_foo();
  }
  ```
  The `derive()` would implement `dispatch_query()` and `dispatch_call()` from `MethodHandler`, plus a new `list_methods()` fn.
- Similar but without macro magic: Add fields `QUERIES` and `CALLS` of type `BTreeMap<String, FnOnce>` to the `Module` signature.
  They describe the dispatch rules. `dispatch_query()`, `dispatch_call()`, and `list_methods()` are then implemented automatically in the default impl based on those two fields.
  
I would like to go with the last (or first) option but IDK if it's OK to change all the existing runtimes.

kostko: I would be in favor of the macro variant.

runtime-sdk/src/modules/core/test.rs
```diff
@@ -808,3 +818,64 @@ fn test_emit_events() {
...
+        types::RuntimeInfoResponse {
+            runtime_version: Version::new(0, 0, 0),
+            state_version: 0,
+            modules: btreemap! {
+                "core" =>
+                    types::ModuleInfo {
+                        version: 1,
+                        params: cbor::to_vec(core_params),
+                        methods: Some(vec![
+                            MethodHandlerInfo { kind: MethodHandlerKind::Query, name: "core.EstimateGas".to_string() },
```
mitjat, line 863: Note: `MethodHandlerInfo` structs do not include arg types because that would require every arg to implement an introspection trait. Auto-deriving that for all possible types seems more work than it's worth.

runtime-sdk/src/module.rs
mitjat: Good point re/ types; implemented with macros in #774

runtime-sdk/src/modules/core/types.rs
```diff
@@ -32,3 +32,37 @@ pub struct CallDataPublicKeyQueryResponse {
...
+    pub name: String,
+}
+
+/// Metadata for an individual module.
+#[derive(Clone, Debug, cbor::Encode, cbor::Decode)]
+#[cfg_attr(test, derive(PartialEq, Eq))]
+pub struct ModuleInfo {
+    pub version: u32,
+    pub params: Vec<u8>,
+    pub methods: Option<Vec<MethodHandlerInfo>>,
```
kostko, line 58: Why not just a vector that can be empty? Any special meaning for `None`?

kostko (commented):
Nice, I think the proposed API makes sense.

runtime-sdk/src/modules/core/types.rs
mitjat: This field reflects the return value of https://github.com/oasisprotocol/oasis-sdk/pull/758/files#diff-b038871be26b907a37aa2a44827ccaa92557482f5cffcca211fbff23de45523dR194: `A return value of None means that the implementor does not care to list the methods, or the implementor is a tuple of modules.`

The idea was to distinguish between a module that explicitly has no methods vs a module that just doesn't support reflection (e.g. because it's not using `sdk_derive()`. On second thought, a module with no methods makes little sense. Lemme simplify the type to just a `Vec`.

mitjat: Done

client-sdk/ts-web/rt/src/types.ts
```diff
@@ -18,6 +18,43 @@ export interface CoreCallDataPublicKeyQueryResponse {
     public_key: KeyManagerSignedPublicKey;
 }
 
+/**
+ * Response to the RuntimeInfo query.
+ */
+export interface CoreRuntimeInfoQueryResponse {
```
mitjat, line 24: Are these TS bindings tested anywhere? I didn't find anything. Same for go.

For go, I tested the new method manually by creating a minimal runtime and its client as per https://docs.oasis.dev/oasis-sdk/guide/minimal-runtime. For TS, I was playing around trying to make `ts-web/rt/playground` work, but no luck in a reasonable amount of time. I can definitely work on this more but would appreciate some pointers.

kostko: Yes.

For Go, see `tests/e2e`, for TS see `client-sdk/ts-web/rt/playground`.

mitjat: Thank you, I saw those. They are not extensive interface tests, they don't exhaustively cover all endpoints, and seem to focus more on core business logic.
Anyway, added my (trivial) tests to both of those folders.

client-sdk/ts-web/rt/src/types.ts
```diff
@@ -18,6 +18,43 @@ export interface CoreCallDataPublicKeyQueryResponse {
...
+    runtime_version: runtime_info.Version;
+    state_version: number;
+    modules: { [key: string]: runtime_info.ModuleInfo };
+}
+
+export namespace runtime_info {
+    export interface Version {
```
mitjat, line 31: I noticed that for the current version 0.1.0, this gets deserialized into `{minor: 1}` in TS -- the zeros are omitted. That's intentional when encoding / on the wire (`#[cbor(skip_serializing_if = "num_traits::Zero::is_zero")]`), but looks a little funny in TS; we'd need the equivalent of `#[cbor(default)]`. I guess we're OK with it for now? We don't have a deserialization framework in place, AFAICT, where we could implant default values.

client-sdk/go/modules/core/types.go
```diff
@@ -12,3 +13,29 @@ type EstimateGasQuery struct {
 	// Tx is the unsigned transaction to estimate.
 	Tx *types.Transaction `json:"tx"`
 }
+
+type RuntimeInfoResponse struct {
```
kostko, line 17: Add documentation comments for these types.

client-sdk/go/modules/core/types.go
```diff
@@ -12,3 +13,29 @@ type EstimateGasQuery struct {
...
+
+type ModuleInfo struct {
+	Version uint32              `json:"version"`
+	Params  []uint8             `json:"params"`
+	Methods []MethodHandlerInfo `json:"methods"`
+}
+
+type MethodHandlerInfo struct {
+	Name string            `json:"name"`
+	Kind methodHandlerKind `json:"kind"`
```
kostko, line 31: Public struct has a field with a private type? That's weird.

runtime-sdk-macros/src/method_handler_derive.rs
```diff
@@ -215,6 +242,20 @@ enum HandlerKind {
     MessageResult,
     Prefetch,
 }
+impl HandlerKind {
```
kostko, line 245: Add empty line above.

runtime-sdk/src/modules/core/test.rs
```diff
@@ -174,12 +175,14 @@ impl GasWasterModule {
...
 impl module::Module for GasWasterModule {
     const NAME: &'static str = "gaswaster";
+    const VERSION: u32 = 42;
     type Error = std::convert::Infallible;
     type Event = ();
     type Parameters = ();
 }
 
-impl module::MethodHandler for GasWasterModule {
+//#[sdk_derive(MethodHandler)]
```
kostko, line 184: Remove?

runtime-sdk/src/modules/core/types.rs
```diff
@@ -32,3 +32,37 @@ pub struct CallDataPublicKeyQueryResponse {
...
+    pub kind: MethodHandlerKind,
+    pub name: String,
+}
+
+/// Metadata for an individual module.
+#[derive(Clone, Debug, cbor::Encode, cbor::Decode)]
+#[cfg_attr(test, derive(PartialEq, Eq))]
+pub struct ModuleInfo {
+    pub version: u32,
+    pub params: Vec<u8>,
```
kostko, line 57: Should this be of type `cbor::Value` instead? That would make the parameters always be valid CBOR embedded into the struct instead of having this arbitrary byte array. On the deserialization/Go side you can just use `cbor.RawMessage`.

kostko (commented):
Nice, looks good, left some minor comments.

client-sdk/go/modules/core/types.go
mitjat: I saw that online as a suggestion for emulating enums. The idea is that you can never create a `methodHandlerKind` from scratch, you need to use one of the predefined constants, which serve as enum values. It seems to mesh well with our cbor enums.
I'm also open to other approaches tho.

runtime-sdk/src/modules/core/types.rs
mitjat: Good call. And on the TS side we're more flexible so we can have `any`, and it deserializes nicely. Updated the type and tests.

client-sdk/ts-web/rt/src/types.ts
```diff
@@ -18,6 +18,43 @@ export interface CoreCallDataPublicKeyQueryResponse {
...
+        params: any;
+        methods: MethodHandlerInfo[];
+    }
+
+    export interface MethodHandlerInfo {
+        name: string;
+        kind: MethodHandlerKind;
+    }
+
+    export enum MethodHandlerKind {
```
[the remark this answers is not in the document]
mitjat: Aspiring to autogenerate interfaces is an excellent reason for consistency. Not that consistency isn't quite excellent on its own :). Sorry, I did not spot the other enums. What this PR introduces is still a little fancier than `CallFormat` because I restrain the type to only the valid values, rather than any `string`. But if we ever autogenerate `string` as a type, that's going to be backwards compatible.

I believe the same is true for the Go interface: What this PR adds is a stricter with types than the `CallFormat` Go interface, but were we to autogenerate the new interface in the style of `CallFormat`, usage wouldn't change.

I now also got rid of the TS namespace.

runtime-sdk/src/modules/core/types.rs
```diff
@@ -32,3 +32,37 @@ pub struct CallDataPublicKeyQueryResponse {
...
     pub public_key: SignedPublicKey,
 }
+
+#[derive(Debug, Copy, Clone, cbor::Encode, cbor::Decode)]
+#[cfg_attr(test, derive(PartialEq, Eq))]
+pub enum MethodHandlerKind {
+    Call,
+    // `Prefetch` is omitted because it is an implementation detail of handling `Call`s.
+    Query,
+    MessageResult,
```
kostko, line 42: Maybe it would make sense to rename the serialized representation of these variants into lower case, e.g.:
* `call`
* `query`
* `message-result`

kostko (commented):
Looks good from my side now, just one minor nit about variant serialization.

client-sdk/ts-web/rt/src/types.ts
```diff
@@ -18,6 +18,42 @@ export interface CoreCallDataPublicKeyQueryResponse {
     public_key: KeyManagerSignedPublicKey;
 }
 
+/**
+ * Response to the RuntimeInfo query.
+ */
+export interface CoreRuntimeInfoQueryResponse {
+    runtime_version: Version;
```
pro-wh, line 25: for `oasis_core_runtime::common::version::Version` use `oasis.types.Version`

client-sdk/ts-web/rt/src/types.ts
```diff
@@ -18,6 +18,42 @@ export interface CoreCallDataPublicKeyQueryResponse {
...
+export interface Version {
+    major: number;
+    minor: number;
+    patch: number;
+}
+
+/**
+ * Metadata for an individual module within the runtime.
+ */
+export interface ModuleInfo {
```
pro-wh, line 39: this will need the `Core` prefix too, `CoreModuleInfo`

client-sdk/ts-web/rt/src/types.ts
```diff
@@ -18,6 +18,42 @@ export interface CoreCallDataPublicKeyQueryResponse {
...
+    version: number;
+    params: any;
+    methods: MethodHandlerInfo[];
+}
+
+export interface MethodHandlerInfo {
```
pro-wh, line 45: same here, `CoreMethodHandlerInfo`

client-sdk/ts-web/rt/src/types.ts
```diff
@@ -18,6 +18,42 @@ export interface CoreCallDataPublicKeyQueryResponse {
...
+    name: string;
+    kind:
+        | typeof METHODHANDLERKIND_CALL
+        | typeof METHODHANDLERKIND_QUERY
+        | typeof METHODHANDLERKIND_PREFETCH;
+}
+
+export const METHODHANDLERKIND_CALL = 'Call';
```
pro-wh, line 53: I've been putting the constants in a typescript module roughly matching the runtime module name--`core.ts` in this case. But then we wouldn't be able to do the `typeof METHODHANDLERKIND_CALL` etc above due to the way the dependencies flow. The sad thing is that we haven't been listing the variants in the type definition in these actually-enum-typed places.

Would it be okay to hardcode the values in the type definition above?

mitjat: I don't mind if you don't :). I hardcoded the values and left a comment to help keep the two sets of strings in sync.

client-sdk/ts-web/rt/src/types.ts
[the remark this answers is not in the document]
pro-wh: thanks!

client-sdk/ts-web/rt/src/types.ts
pro-wh: :+1:

client-sdk/ts-web/rt/src/types.ts
pro-wh: yeah, we're living with it.

client-sdk/ts-web/rt/playground/src/index.js
```diff
@@ -285,6 +285,47 @@ export const playground = (async function () {
...
+            throw new Error('accounts module was expected in RuntimeInfo response');
+        }
+
+        // Check deserialization of lists and enums
+        if (
+            accts.methods.find(
+                (m) =>
+                    m.name === 'accounts.Transfer' &&
+                    m.kind === oasisRT.core.METHODHANDLERKIND_CALL,
+            ) === undefined
```
pro-wh, line 305: accts.methods.some

:superhero:

pro-wh (commented):
:+1: on ts-web

client-sdk/go/modules/core/types.go
```diff
@@ -49,3 +50,40 @@ type GasUsedEvent struct {
...
+}
+
+type methodHandlerKind string
+
+// These constants represent the kinds of methods that handlers handle.
+const (
+	MethodHandlerKindCall          methodHandlerKind = "Call"
+	MethodHandlerKindQuery         methodHandlerKind = "Query"
+	MethodHandlerKindMessageResult methodHandlerKind = "MessageResult"
```
kostko, line 88: These also need to be updated.

mitjat: Thanks! Tests caught it also :ok_hand:, I just didn't run them locally after this latest tweak.

kostko (approved):

PR #759, opened 2022-01-25
runtime-sdk: Add support for runtime schedule control

kostko:
Requires a new version of Oasis Core.

TODO
* [x] More tests.
* [x] Fetch more transactions from queue when space remains in batch.
* [x] Use released version of Oasis Core.

### DIFF
--- a/runtime-sdk/src/dispatcher.rs
+++ b/runtime-sdk/src/dispatcher.rs
@@ -12,6 +12,7 @@ use thiserror::Error;
 
 use oasis_core_runtime::{
     self,
+    common::crypto::hash::Hash,
     protocol::HostInfo,
     storage::mkvs,
     transaction::{
@@ -34,6 +35,7 @@ use crate::{
     modules,
     modules::core::API as _,
     runtime::Runtime,
+    schedule_control::ScheduleControlHost,
     storage,
     storage::Prefix,
     types,
@@ -100,6 +102,7 @@ impl From<module::CallResult> for DispatchResult {
 pub struct Dispatcher<R: Runtime> {
     host_info: HostInfo,
     key_manager: Option<KeyManagerClient>,
+    schedule_control_host: Arc<dyn ScheduleControlHost>,
     _runtime: PhantomData<R>,
 }
 
@@ -108,10 +111,15 @@ impl<R: Runtime> Dispatcher<R> {
     ///
     /// Note that the dispatcher is fully static and the constructor is only needed so that the
     /// instance can be used directly with the dispatcher system provided by Oasis Core.
-    pub(super) fn new(host_info: HostInfo, key_manager: Option<KeyManagerClient>) -> Self {
+    pub(super) fn new(
+        host_info: HostInfo,
+        key_manager: Option<KeyManagerClient>,
+        schedule_control_host: Arc<dyn ScheduleControlHost>,
+    ) -> Self {
         Self {
             host_info,
             key_manager,
+            schedule_control_host,
             _runtime: PhantomData,
         }
     }
@@ -386,17 +394,17 @@ impl<R: Runtime> Dispatcher<R> {
         .map_err(|err| -> RuntimeError { Error::QueryAborted(format!("{:?}", err)).into() })?
         .map(cbor::to_vec)
     }
-}
 
-impl<R: Runtime + Send + Sync> transaction::dispatcher::Dispatcher for Dispatcher<R> {
-    fn execute_batch(
+    fn execute_batch_common<F>(
         &self,
         mut rt_ctx: transaction::Context<'_>,
-        batch: &TxnBatch,
-    ) -> Result<ExecuteBatchResult, RuntimeError> {
-        // If prefetch limit is set enable prefetch.
-        let prefetch_enabled = R::PREFETCH_LIMIT > 0;
-
+        f: F,
+    ) -> Result<ExecuteBatchResult, RuntimeError>
+    where
+        F: FnOnce(
+            &mut RuntimeBatchContext<'_, R, storage::MKVSStore<&mut dyn mkvs::MKVS>>,
+        ) -> Result<Vec<ExecuteTxResult>, RuntimeError>,
+    {
         // Prepare dispatch context.
         let key_manager = self
             .key_manager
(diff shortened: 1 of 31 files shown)


runtime-sdk/src/dispatcher.rs
```diff
@@ -467,8 +440,130 @@ impl<R: Runtime + Send + Sync> transaction::dispatcher::Dispatcher for Dispatche
...
+                        }
+
+                        // Determine the current transaction index.
+                        let index = new_batch.len();
+
+                        new_batch.push(raw_tx);
+                        results.push(Self::execute_tx(ctx, tx_size, tx, index)?);
+                    }
+
+                    // TODO: If there's more room in the block and we got the maximum number of
```
ptrus, line 550: Also some timeout probably? So it doesn't indefinitely wait for new transactions?

kostko: Yeah so the idea is to do it until either the batch is full (remaining batch gas is below the set minimum) or we've exhausted the queue (returned length is less than the requested length).

kostko: This now has an implementation, please take a look.

runtime-sdk/src/dispatcher.rs
```diff
@@ -467,8 +447,148 @@ impl<R: Runtime + Send + Sync> transaction::dispatcher::Dispatcher for Dispatche
...
+                        let index = new_batch.len();
+
+                        new_batch.push(raw_tx);
+                        results.push(Self::execute_tx(ctx, tx_size, tx, index)?);
+                    }
+
+                    // If there's more room in the block and we got the maximum number of
+                    // transactions, request more transactions.
+                    if last_batch_tx_hash.is_some()
+                        && last_batch_len >= requested_batch_len as usize
```
nhynes, line 565: can you early break this?

runtime-sdk/src/dispatcher.rs
```diff
@@ -467,8 +447,148 @@ impl<R: Runtime + Send + Sync> transaction::dispatcher::Dispatcher for Dispatche
...
+                for (index, (tx_size, tx)) in txs.into_iter().enumerate() {
+                    results.push(Self::execute_tx(ctx, tx_size, tx, index)?);
+                }
+
+                Ok(results)
+            },
+        )
+    }
+
+    fn schedule_and_execute_batch(
```
nhynes, line 502: Is this supposed to be called from somewhere in this PR?

runtime-sdk/src/dispatcher.rs
```diff
@@ -467,8 +447,148 @@ impl<R: Runtime + Send + Sync> transaction::dispatcher::Dispatcher for Dispatche
...
+                        // Determine the current transaction index.
+                        let index = new_batch.len();
+
+                        new_batch.push(raw_tx);
+                        results.push(Self::execute_tx(ctx, tx_size, tx, index)?);
+                    }
+
+                    // If there's more room in the block and we got the maximum number of
+                    // transactions, request more transactions.
+                    if last_batch_tx_hash.is_some()
```
nhynes, line 564: I'd personally make this function operation on an iterator of batches, but that's just because I can't think about tracking so many little bits of state :P

nhynes (approved):

runtime-sdk/src/dispatcher.rs
```diff
@@ -467,8 +447,148 @@ impl<R: Runtime + Send + Sync> transaction::dispatcher::Dispatcher for Dispatche
...
+                // available in the block as determined by gas use.
+                let mut new_batch = Vec::new();
+                let mut results = Vec::with_capacity(batch.len());
+                let mut requested_batch_len = cfg.initial_batch_size;
+                'batch: loop {
+                    // Remember length of last batch.
+                    let last_batch_len = batch.len();
+                    let last_batch_tx_hash = batch.last().map(|raw_tx| Hash::digest_bytes(raw_tx));
+
+                    for raw_tx in batch.drain(..) {
```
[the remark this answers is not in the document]
kostko: Yes these are sorted by priority (defined by the runtime SDK to be based on gas price).

runtime-sdk/src/dispatcher.rs
kostko: No, this is part of the dispatcher trait from Oasis Core, see the PR in https://github.com/oasisprotocol/oasis-core/pull/4446.

runtime-sdk/src/dispatcher.rs
kostko: What do you mean?

kostko: You mean before trying all transactions? Yeah there's this:
```rust
                        // If we don't have enough gas for processing even the cheapest transaction
                        // we are done.
                        let remaining_gas = R::Core::remaining_batch_gas(ctx);
                        if remaining_gas < cfg.min_remaining_gas {
                            break 'batch;
                        }
```

runtime-sdk/src/dispatcher.rs
```diff
@@ -467,8 +447,148 @@ impl<R: Runtime + Send + Sync> transaction::dispatcher::Dispatcher for Dispatche
...
+                                // queue and don't include it in a block.
+                                tx_reject_hashes.push(Hash::digest_bytes(&raw_tx));
+                                continue;
+                            }
+                        };
+                        let tx_size = raw_tx.len().try_into().unwrap();
+
+                        // If we don't have enough gas remaining to process this transaction, just
+                        // skip it.
+                        if tx.auth_info.fee.gas > remaining_gas {
```
pro-wh, line 547: :information_source: we always compare the declared gas limit. in the flood-of-max-gas-txs-that-end-up-spending-only-a-little case, we should be able to process one per round. this check lets us skip through the rest in this round and process smaller transactions in the queue

runtime-sdk/src/dispatcher.rs
```diff
@@ -467,8 +447,148 @@ impl<R: Runtime + Send + Sync> transaction::dispatcher::Dispatcher for Dispatche
...
+                    {
+                        if let Ok(Some(fetched_batch)) = self
+                            .schedule_control_host
+                            .fetch_tx_batch(last_batch_tx_hash, cfg.batch_size)
+                        {
+                            *batch = fetched_batch;
+                            requested_batch_len = cfg.batch_size;
+                            continue;
+                        }
+                        // Failed to fetch batch or batch empty, let's just finish.
```
pro-wh, line 575: should we log the error or anything?

runtime-sdk/src/dispatcher.rs
```diff
@@ -467,8 +447,148 @@ impl<R: Runtime + Send + Sync> transaction::dispatcher::Dispatcher for Dispatche
...
+                'batch: loop {
+                    // Remember length of last batch.
+                    let last_batch_len = batch.len();
+                    let last_batch_tx_hash = batch.last().map(|raw_tx| Hash::digest_bytes(raw_tx));
+
+                    for raw_tx in batch.drain(..) {
+                        // If we don't have enough gas for processing even the cheapest transaction
+                        // we are done.
+                        let remaining_gas = R::Core::remaining_batch_gas(ctx);
+                        if remaining_gas < cfg.min_remaining_gas {
```
pro-wh, line 529: :information_source: this prevents us from decoding a tx when there's less gas than the cheapest possible transaction remaining.

any reason we might set this to a number other than the cheapest possible transaction, btw?

pro-wh (approved):
with this we'll have the most straightforward "greedy" scheduling algorithm

runtime-sdk/src/dispatcher.rs
ptrus: Looks good

runtime-sdk/src/dispatcher.rs
kostko: Nope, this should be set to the gas cost of the cheapest possible transaction. I just didn't want to add an automated mechanism to determine that.

runtime-sdk/src/dispatcher.rs
kostko: Yeah, fixed.

PR #774, opened 2022-02-04
proc_macro for deriving the `MethodHandler` trait

mitjat:
Resolves https://github.com/oasisprotocol/oasis-sdk/issues/24

Prerequisite for https://github.com/oasisprotocol/oasis-sdk/issues/652 (as it will enable introspection)

### DIFF
--- a/contract-sdk-macros/src/test_utils.rs
+++ b/contract-sdk-macros/src/test_utils.rs
@@ -18,13 +18,17 @@ pub fn rustfmt(code: &str) -> String {
         .expect("unable to communicate with rustfmt");
 
     let output = cp.wait_with_output().unwrap();
-    if !output.status.success() {
-        panic!(
-            "unable to rustfmt\n{}",
-            String::from_utf8(output.stderr).unwrap_or_default()
-        );
-    }
-    String::from_utf8(output.stdout).unwrap()
+    assert!(
+        output.status.success(),
+        "unable to rustfmt\n{}",
+        String::from_utf8(output.stderr).unwrap_or_default()
+    );
+    let output = String::from_utf8(output.stdout).unwrap();
+    // Remove the "const _: () = ..." wrapper we added above
+    let mut output =
+        output.split('\n').collect::<Vec<_>>()[1..output.matches('\n').count() - 1].join("\n");
+    output.pop();
+    output
 }
 
 #[macro_export]
@@ -35,30 +39,11 @@ macro_rules! assert_empty_diff {
         let actual_code = crate::test_utils::rustfmt(&$actual.to_token_stream().to_string());
         let expected_code = crate::test_utils::rustfmt(&$expected.to_token_stream().to_string());
 
-        let diffs = diff::lines(&actual_code, &expected_code);
+        let diff = difference::Changeset::new(&expected_code, &actual_code, "\n");
 
-        let mut has_diff = false;
-        let mut actual_lines = actual_code.split('\n');
-        let mut expected_lines = expected_code.split('\n');
-
-        for (i, diff) in diffs.iter().enumerate() {
-            match diff {
-                diff::Result::Left(l) => {
-                    eprintln!("non-empty diff on line {}", i);
-                    eprintln!("+ {}", l);
-                    eprintln!("- {}", expected_lines.next().unwrap());
-                    has_diff = true;
-                }
-                diff::Result::Right { .. } => {
-                    actual_lines.next();
-                    has_diff = true;
-                }
-                diff::Result::Both { .. } => {
-                    actual_lines.next();
-                    expected_lines.next();
-                }
-            }
+        if diff.distance > 0 {
+            eprintln!("Diff:\n{}\n\nActual output:\n{}\n", diff, actual_code);
         }
-        assert!(!has_diff);
+        assert!(diff.distance == 0);
     }};
 }
--- a/runtime-sdk-macros/src/lib.rs
+++ b/runtime-sdk-macros/src/lib.rs
@@ -7,9 +7,27 @@ use proc_macro::TokenStream;
 mod error_derive;
 mod event_derive;
 mod generators;
+mod method_handler_derive;
 #[cfg(test)]
 mod test_utils;
 mod version_from_cargo;
+
+/// Emits a compile-time error `msg` from a macro, and uses the span of the
+/// macro invocation if possible. If span info is not available (should only
+/// happen in unit tests), panics with `msg`.
+fn emit_compile_error<S: Into<String> + Clone + std::panic::RefUnwindSafe>(msg: S) -> ! {
+    std::panic::catch_unwind(|| {
+        proc_macro2::Span::call_site()
+            .unwrap()
+            .error(msg.clone().into())
+            .emit();
+    })
+    .ok()
+    .or_else(|| {
+        panic!("{}", msg.into());
+    });
+    unreachable!(); // error().emit() already halts compilation, but type checker doesn't know that
+}
 
 /// Derives the `Event` trait on an enum.
 #[proc_macro_derive(Event, attributes(sdk_event))]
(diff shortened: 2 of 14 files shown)


contract-sdk-macros/src/test_utils.rs
```diff
@@ -35,30 +40,11 @@ macro_rules! assert_empty_diff {
         let actual_code = crate::test_utils::rustfmt(&$actual.to_token_stream().to_string());
         let expected_code = crate::test_utils::rustfmt(&$expected.to_token_stream().to_string());
 
-        let diffs = diff::lines(&actual_code, &expected_code);
+        let diff = difference::Changeset::new(&expected_code, &actual_code, "\n");
```
mitjat, line 43: Not strictly related to the PR, but displays diffs in a much cleaner way.

runtime-sdk/src/modules/consensus_accounts/mod.rs
```diff
@@ -8,18 +8,19 @@ use once_cell::sync::Lazy;
 use thiserror::Error;
 
 use oasis_core_runtime::consensus::staking::Account as ConsensusAccount;
+use oasis_runtime_sdk_macros::{handler, sdk_derive};
```
mitjat, line 11: The `consensus_accounts` module has one of the more complex MessageHandler impls. I modified it to use the new derive macro to serve as an example. Once the rest of the PR is reviews, I'll update the other modules too.

runtime-sdk-macros/src/lib.rs
```diff
@@ -26,6 +43,45 @@ pub fn error_derive(input: TokenStream) -> TokenStream {
...
+
+/// A helper attribute for `#[sdk_derive(...)]`. It doesn't do anyting on its own;
+/// it only mark functions that represent a paratime method handler.
+/// The permitted forms are:
+///  - `#[handler(call = "my_module.MyCall")]`: Marks a function that handles
+///        the "my_module.MyCall" call and can be passed to
+///        oasis_runtime_sdk::module::dispatch_call.
+///  - `#[handler(prefetch = "my_module.MyCall")]`: Marks a function that handles
+///        the request to prefetch any data ahead of the "my_module.MyCall" call.
+///        Its signature should be `Fn(
```
mitjat, line 69: This signature is clumsy to explain; compare to other `#[handler(...)]` versions. I'm sure people will turn to existing examples anyway to understand the types/usage, but: It would be nicer if there was an `oasis_runtime_sdk::module:dispatch_prefetch()` helper function, just like for the the other methods. It would parse `args` (just like the other `dispatch_*()` helper fns do) and would possibly expect a `Vec<Prefix>` from the user-implemented callback, rather than passing the unusual `add_prefix` fn.

I assumed that performance was the reason `prefetch` is an exception and doesn't have that kind of an ergonomic wrapper: Not every prefetch handler needs to parse args, and exposing the global `BTreeSet` of prefixes-to-fetch avoids some copying. LMK if that's the case or if I should update prefetch to match the other handlers more closely in style.

runtime-sdk/src/modules/consensus_accounts/mod.rs
```diff
@@ -270,13 +294,15 @@ impl<Accounts: modules::accounts::API, Consensus: modules::consensus::API>
...
 
+    #[handler(query = "consensus.Account")]
     fn query_consensus_account<C: Context>(
         ctx: &mut C,
         args: types::ConsensusAccountQuery,
     ) -> Result<ConsensusAccount, Error> {
         Consensus::account(ctx, args.address).map_err(|_| Error::InvalidArgument)
     }
 
+    #[handler(message_result = "consensus.TransferFromRuntime")]
```
kostko, line 305: Would it be possible to use a constant here (e.g. `CONSENSUS_TRANSFER_HANDLER`)? Because otherwise this could get out of sync with the emitted message in `withdraw` above. Similar for the below one.

kostko (commented):
Nice, added some comments.

runtime-sdk-macros/src/method_handler_derive.rs
```diff
@@ -0,0 +1,456 @@
...
+                ) -> module::DispatchResult<cbor::Value, Result<(), error::RuntimeError>> {
+                    let mut add_prefix = |p| {prefixes.insert(p);};
+                    match method {
+                        // "Real", user-defined prefetch handlers.
+                        #(
+                          #handler_names => module::DispatchResult::Handled(
+                            Self::#handler_idents(&mut add_prefix, body, auth_info)
+                          ),
+                        )*
+                        // No-op prefetch handlers.
```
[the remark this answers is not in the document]
mitjat: I was following the pattern from https://github.com/oasisprotocol/oasis-sdk/blob/main/runtime-sdk/src/modules/consensus_accounts/mod.rs#L374, where a no-op prefetch is explicit. It's also explicitly enforced by unit tests in that module: https://github.com/oasisprotocol/oasis-sdk/blob/main/runtime-sdk/src/modules/consensus_accounts/test.rs#L895

But I do see what you're talking about:
https://github.com/oasisprotocol/oasis-sdk/blob/main/runtime-sdk/src/dispatcher.rs#L300

Should I remove the autogenerated no-op handlers and update the test in consensus_accounts to be OK with unhandled prefetch?

runtime-sdk/src/modules/consensus_accounts/mod.rs
```diff
@@ -317,6 +343,7 @@ impl<Accounts: modules::accounts::API, Consensus: modules::consensus::API>
         });
     }
 
+    #[handler(message_result = "consensus.WithdrawIntoRuntime")]
```
ptrus, line 346: This one should also be updated to use the constant (`CONSENSUS_WITHDRAW_HANDLER`)

runtime-sdk-macros/src/method_handler_derive.rs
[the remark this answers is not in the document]
ptrus: The difference between explicitly (no-op) handled and `Unhandled` is that in case of the explicitly handled, the `Tuple` implementation will return early:

https://github.com/oasisprotocol/oasis-sdk/blob/ba49e50b216f471257436e560f551f061003bfce/runtime-sdk/src/module.rs#L200-L206

This should be irrelevant in practice.

---

The only reason that Unhandled prefetch is allowed is mostly due to this being a bit of an "experimental" feature at the moment and is not widely used (also disabled by default) so only a couple of modules were provided with the implementation.

I think there are two equally valid steps we can take here:
- keep the autogenerated no-op handlers and don't allow unhandled prefetch
- allow unhandled prefetch, remove the autogenerated no-op handlers

Unless I missed/forgot any important other details, I think I would prefer the first option (keep the autogenerated handlers).

runtime-sdk-macros/src/lib.rs
ptrus: Yes, only due to performance reasons, since the existing benchmarks (although in an unrealistic scenario) showed that the performance benefit of prefetch was minimal and the additional parsing / copying could lead to poorer overall performance.

Feel free to open an issue for this, once we have some more realistic benchmarks we could revisit this.

ptrus (commented):
Nice, this will remove a lot of existing boilerplate!

runtime-sdk-macros/src/method_handler_derive.rs
```diff
@@ -0,0 +1,456 @@
+use crate::{emit_compile_error, generators as gen};
+use darling::FromMeta;
+use proc_macro2::TokenStream;
+use quote::quote;
+
+/// See `derive_method_handler()` in lib.rs.
```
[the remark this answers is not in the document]
mitjat: link rot :|
fixed.

runtime-sdk/src/modules/consensus_accounts/mod.rs
mitjat: :see_no_evil: 
fixed

runtime-sdk-macros/src/method_handler_derive.rs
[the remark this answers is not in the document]
kostko: I see, ok agree that the first option (keeping autogenerated no-op handlers) is better. And we can forbid unhandled prefetch separately.

kostko (commented):
Looks good. Is the plan to also convert the other modules to use the new handlers?

runtime-sdk-macros/src/method_handler_derive.rs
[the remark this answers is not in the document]
mitjat: Thank you for the explanation, Peter. I'm keeping the autogenerated no-op handlers then.

mitjat:
> Looks good. Is the plan to also convert the other modules to use the new handlers?

[Yep](https://github.com/oasisprotocol/oasis-sdk/pull/774#discussion_r799073287), I just wanted to get the OK on the macro first, before I hand-edit all the modules.

Modules have now been migrated to use the new sdk_derive, and the PR is ready for a final review.

mitjat:
Hrm code coverage seems to not "see" the code annotated with `#[handler]`. :/  
Visual studio code (via rust-analyzer) also marks all those methods as "unreachable".

We could instead allow a little more duplication and use a syntax that's similar to the old one, something like

```rust
sdk_derive!(MethodHandler, {
  call  "accounts.Transfer" => tx_transfer,
  query "accounts.Nonce"    => query_nonce,
  query "accounts.Balances" => query_balances,
  // ...
})
```

I find this newer proposal a little more idiomatic (good), a little more verbose (bad), friendlier to dev tooling (good), a rewrite of what we currently have (unappealing but not a big deal). I'm OK either way, somebody else call the shots :).

kostko:
> Hrm code coverage seems to not "see" the code annotated with #[handler]. :/

Where do you see this? I'm looking at the latest coverage report and it seems like it does take those into account? At least not any differenty than before?

mitjat:
> > Hrm code coverage seems to not "see" the code annotated with #[handler]. :/
> 
> Where do you see this? 

I was looking at a report where coverage fell from ~60% to ~30% (IIRC) for individual modules. Can't find it any more ... So let's assume I was looking at the wrong output; I've no idea which one / where. Sorry for the false alarm.

kostko (approved):

ptrus (approved):

PR #775, opened 2022-02-07
runtime-sdk: Reject batches that overflow batch gas limit

kostko:


Yawning (approved):

pro-wh (approved):
I take it Error::Abort has some special handling?

kostko:
> I take it Error::Abort has some special handling?

Yes, it is tagged with `#[sdk_error(abort)]` which is handled by the dispatcher to cause a batch abort.

PR #781, opened 2022-02-16
cli/wallet: Add Ledger support for consensus transactions

kostko:


pro-wh (commented):
how does cli/wallet/ledger fit in with the other various implementations of the protocol we have in the organization's repos? is this the first Go implementation?

kostko:
This is basically lifted from oasis-core-ledger with some minor rework.

The reason is:
- I didn't want to require users to use the plugin signer that Oasis Core uses as it is cumbersome.
- The current oasis-core-ledger does not export the useful parts as a library to allow reuse.

It would probably make sense to refactor oasis-core-ledger to make it usable in different contexts, but I didn't want this to block adding Ledger support to the new CLI. Once that is refactored the CLI can be updated to use the new lib.

pro-wh:
ok, can we leave a code comment in the module so that readers know where to find the canonical version?

kostko:
> ok, can we leave a code comment in the module so that readers know where to find the canonical version?

Done.

kostko:
Note that this only supports ADR 0008 currently. Should we allow people to use the old derivation scheme as well? We probably should and it would be simple to add.

pro-wh:
I think we should. I'd go as far to say that if we don't, we should add something user-facing to communicate that this is something different from basically all oasis ledger software currently in use

kostko:
I added support for the legacy derivation scheme (it can be set when creating a new wallet using `--ledger.derivation legacy`). Can someone verify that it generates the right addresses?

matevz:
> I added support for the legacy derivation scheme (it can be set when creating a new wallet using `--ledger.derivation legacy`). Can someone verify that it generates the right addresses?

Tested the first five derived accounts and the addresses match the ones from the oasis web wallet. Good job!

pro-wh (approved):
sounds good, matevz thanks for checking

kostko:
Task linked: [CU-2gf4pgf Add support to new Oasis CLI](https://app.clickup.com/t/2gf4pgf)

PR #782, opened 2022-02-17
Add debugging niceties

nhynes:
This PR:
* adds a `debug_print` method for inspecting the state of the contract
* makes methods on `Module` public so that third parties can programmatically invoke the environment (I copied over the e2e tests from this repo to my own crate, with good results)

### DIFF
--- a/contract-sdk/src/abi/env.rs
+++ b/contract-sdk/src/abi/env.rs
@@ -20,6 +20,9 @@ extern "C" {
 
     #[link_name = "address_for_instance"]
     fn env_address_for_instance(instance_id: u64, dst_ptr: u32, dst_len: u32);
+
+    #[link_name = "debug_print"]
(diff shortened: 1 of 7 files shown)


contract-sdk/src/abi/env.rs
```diff
@@ -53,6 +56,17 @@ impl Env for HostEnv {
...
     }
+
+    #[cfg(feature = "debug-utils")]
+    fn debug_print(&self, msg: &str) {
+        debug_print(msg)
+    }
+}
+
+#[cfg(feature = "debug-utils")]
+pub(crate) fn debug_print(msg: &str) {
```
kostko, line 67: Why is this separate and `pub(crate)`? It doesn't seem to be used anywhere?

nhynes: It's useful for inserting into the sdk itself before a `Ctx` is available

kostko (approved):

PR #784, opened 2022-02-17
runtime-sdk: fix builds by using latest oasis-core

nhynes:
I'm not entirely sure if this is the most acceptable way to do it, but this PR fixes the problem with the `anomaly` crate being yanked by upgrading to the latest oasis-core, but not implementing any of the message handling stuff.

Otherwise, anyone currently building without using `--offline` (and a cached copy of `anomaly`) will get an error about `tendermint` missing the dep.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,21 +3,6 @@
 version = 3
 
 [[package]]
-name = "addr2line"
-version = "0.15.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e7a2e47a1fbe209ee101dd6d61285226744c6c8d3c21c8dc878ba6cb9f467f3a"
-dependencies = [
- "gimli",
-]
-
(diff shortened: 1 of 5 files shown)


pro-wh (commented):
I kinda do want this repo to be buildable, but is it viable to bump all the way to a version of oasis-core that's not deployed yet?

ccing @kostko--would we be willing to make a buildable version of oasis-core that's still compatible with the version on mainnet/testnet?

ptrus:
I think better (meaning non-breaking) option would be to fix the compilation in the oasis-core stable/21.3.x branch and then create a new release (or just depend on the stable/21.3.x branch instead of the tag here)

ptrus:
Builds on `main` branch should now be fixed

kostko:
I suggest we close this for now and bump once Oasis Core 22.0 is out.

PR #788, opened 2022-02-21
core: emit gas used event for every transaction

ptrus:


runtime-sdk/src/module.rs
```diff
@@ -304,6 +304,13 @@ pub trait AuthHandler {
         // Default implementation doesn't do anything.
         Ok(())
     }
+
+    /// Perform any action after call, within the transaction context.
+    /// TODO: maybe move into a separate trait, or rename AuthHandler to something like "CallHandler".
```
kostko, line 309: Yeah this should probably be call handler or transaction handler.

ptrus: Done renamed to `TransactionHandler`

tests/benchmark/run-benchmarks.sh
```diff
@@ -60,4 +60,4 @@ sleep 2
     --address "${CLIENT_SOCK}" \
     --runtime.id "8000000000000000000000000000000000000000000000000000000000000000" \
     --benchmarks accounts_transfers \
-    --benchmarks.concurrency 5000
+    --benchmarks.concurrency 4000
```
ptrus, line 63: Had to reduce concurrency a bit, to make it work on CI.

Tested locally (with previous concurrency value):
- the results went from 850/s->450/s. I guess that's kind of expected since every transaction now also emits an event and account transfer does quite simple transactions

runtime-sdk/src/module.rs
```diff
@@ -304,10 +304,16 @@ pub trait AuthHandler {
         // Default implementation doesn't do anything.
         Ok(())
     }
+
+    /// Perform any action after call, within the transaction context.
```
kostko, line 308: Document what happens if this returns an error (transaction fails and updates are rolled back).

kostko (approved):

PR #790, opened 2022-02-22
ts-web: typescript strict mode

pro-wh:
checking what complaints come up in strict mode

reviewers: ignore whitespace due to some indentation changes

### DIFF
--- a/client-sdk/ts-web/core/src/consensus.ts
+++ b/client-sdk/ts-web/core/src/consensus.ts
@@ -126,6 +126,7 @@ export class TransactionWrapper<BODY> {
             },
             method,
         };
+        this.signedTransaction = null as never;
     }
 
     setNonce(nonce: types.longnum) {
@@ -134,12 +135,12 @@ export class TransactionWrapper<BODY> {
     }
 
     setFeeAmount(amount: Uint8Array) {
-        this.transaction.fee.amount = amount;
+        this.transaction.fee!.amount = amount;
         return this;
     }
 
     setFeeGas(gas: types.longnum) {
-        this.transaction.fee.gas = gas;
+        this.transaction.fee!.gas = gas;
         return this;
     }
 
@@ -180,7 +181,7 @@ export type SignatureMessageHandlersWithChainContext = {
 };
 
 export type TransactionHandler<BODY> = (body: BODY) => void;
-export type TransactionHandlers = {[method: string]: TransactionHandler<unknown>};
+export type TransactionHandlers = {[method: string]: TransactionHandler<never>};
 
 /**
  * Calls one of the handlers based on the given transaction method.
@@ -190,8 +191,8 @@ export type TransactionHandlers = {[method: string]: TransactionHandler<unknown>
  * @returns `true` if the transaction method matched one of the handlers
  */
 export function visitTransaction(handlers: TransactionHandlers, tx: types.ConsensusTransaction) {
-    if (tx.method in handlers) {
-        handlers[tx.method](tx.body);
+    if (handlers[tx.method]) {
+        handlers[tx.method](tx.body as never);
         return true;
     }
     return false;
--- a/client-sdk/ts-web/core/src/signature.ts
+++ b/client-sdk/ts-web/core/src/signature.ts
@@ -168,10 +168,10 @@ export class NaclSigner implements Signer {
 }
 
 export type MessageHandlerBare<PARSED> = (v: PARSED) => void;
-export type MessageHandlersBare = {[context: string]: MessageHandlerBare<unknown>};
+export type MessageHandlersBare = {[context: string]: MessageHandlerBare<never>};
 export type MessageHandlerWithChainContext<PARSED> = (chainContext: string, v: PARSED) => void;
 export type MessageHandlersWithChainContext = {
-    [context: string]: MessageHandlerWithChainContext<unknown>;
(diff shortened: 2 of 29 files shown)


client-sdk/ts-web/core/src/signature.ts
```diff
@@ -193,16 +193,19 @@ export function visitMessage(handlers: MessageHandlers, context: string, message
         const parts = context.split(CHAIN_CONTEXT_SEPARATOR);
         if (parts.length === 2) {
             const [context2, chainContext] = parts;
-            if (context2 in handlers.withChainContext) {
-                handlers.withChainContext[context2](chainContext, misc.fromCBOR(message));
+            if (handlers.withChainContext && context2 in handlers.withChainContext) {
```
pro-wh, line 196: this and a similar change below is good. .withChainContext may be unset

client-sdk/ts-web/core/src/signature.ts
```diff
@@ -168,10 +168,10 @@ export class NaclSigner implements Signer {
 }
 
 export type MessageHandlerBare<PARSED> = (v: PARSED) => void;
-export type MessageHandlersBare = {[context: string]: MessageHandlerBare<unknown>};
+export type MessageHandlersBare = {[context: string]: MessageHandlerBare<never>};
```
pro-wh, line 171: using bottom type so that modules can conveniently provide contravariant handler of their specific body type. but then things get annoying when we call. lmk if you have better ideas for how to do this

client-sdk/ts-web/core/src/consensus.ts
```diff
@@ -115,7 +115,7 @@ export async function hashSignedTransaction(signed: types.SignatureSigned) {
 
 export class TransactionWrapper<BODY> {
     transaction: types.ConsensusTransaction;
-    signedTransaction: types.SignatureSigned;
+    signedTransaction?: types.SignatureSigned;
```
pro-wh, line 118: :skull: there's no way to do `wrapper.signedTransaction!` in javascript

client-sdk/ts-web/core/src/signature.ts
pro-wh: come to think of it, if there's going to be annoying casting on every dispatch, maybe we should simplify and use `{}`

client-sdk/ts-web/core/src/consensus.ts
lukaw3d: ew
I would choose wrong type then
```suggestion
    signedTransaction: types.SignatureSigned = undefined as any;
```

pro-wh: turns out you can do

```js
/** @type {oasis.types.SignatureSigned} */ (tw.signedTransaction)
```

to cast away the `| undefined` ness

pro-wh: ok mostly using `null as never` in this new draft

client-sdk/ts-web/core/src/signature.ts
pro-wh: ok I made it even weirder by casting `fromCBOR(...) as never` instead of `handler as MessageHandlerBare<unknown>` :upside_down_face:

lukaw3d: how come `never` instead of `any`?

pro-wh: the parameter type is already `never` :laughing:

pro-wh: ok other answer now that I thought about it some more:

`never` makes sense here because statically we don't know of a type that will work as a parameter here. we couldn't, because it's up to the more specialized runtime modules to know what parameter types go with what contexts. if typescript is meant to answer the question "is it guaranteed to be safe if we call the handler with an argument of type X?" then for whatever type X we are able to name in this part of the code, the answer is, correctly, "no." and the contravariance works out too, with those more specialized contexts permitted to substitute a MessageHandlerBare<Something> where a MessageHandlerBare<never> is called for.

for practical purposes, `never` forces you to cast, which is more correct if you genuinely don't know what the parameter type is supposed to be. `any` saves that cast, but we don't particularly need it.

and we used to use `unknown`, which incorrectly meant that the handler should be able to handle anything and internally check the type

pro-wh:
ok another thing with strict mode: `field?: T` allows `field` to be present and have the value `undefined`. there's this "recommended" option https://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes to get rid of the `| undefined`, but it doesn't come default. so let's go without it for now and really allow present-but-undefined for now. in order to support this, I'm changing most field checks from `'field' in struct` to `struct.field`. :roll_eyes: it'll even compress better since we often then do `struct.field` again later. as far as I know, we don't do this in any situations where the field value is primitive and can be `0` or something. and I'm changing most other `key in someDictOrWhatever` to be `someDictOrWhatever[key]` for consistency. :slightly_frowning_face:

lukaw3d (approved):

pro-wh:
cool, gonna rebase and merge

pro-wh:
nvm, still can't :laughing:

pro-wh:
@kostko or @ptrus can you review

kostko (approved):
Ooops, looks like this got lost.

pro-wh:
rebased, added changelogs and changelogs for earlier changes

PR #792, opened 2022-02-22
evm: revert on selfdestruct opcode

ptrus:


Yawning:
I get that a revert is a revert, but should this ensure that nothing screwed up is happening to the contract's storage?

kostko (commented):
Also update the README in `modules/evm/README.md` to note that SELFDESTRUCT will cause a revert.

kostko:
> I get that a revert is a revert, but should this ensure that nothing screwed up is happening to the contract's storage?

In case of an error being returned from a transaction, our handler will cause any storage changes to be rolled back. So it will be as if the transaction never happened (except the fee payments).

tests/benchmark/run-benchmarks.sh
```diff
@@ -60,4 +60,4 @@ sleep 2
     --address "${CLIENT_SOCK}" \
     --runtime.id "8000000000000000000000000000000000000000000000000000000000000000" \
     --benchmarks accounts_transfers \
-    --benchmarks.concurrency 4000
+    --benchmarks.concurrency 3000
```
ptrus, line 63: Not related to changes here, but hopefully this reduces the flakiness of the benchmarks CI job (works fine locally).

kostko (approved):

PR #793, opened 2022-02-22
contract-sdk: Expose signature verification to wasm contracts

### LINKED ISSUE #368: Native cryptography function imports
The smart contracts module should provide a set of native cryptography function imports (probably under the `crypto` WASM module) in order to avoid the cost of implementing them in WASM.

Currently this should include the signature schemes used in the SDK:
* Ed25519 (verify, batch verify)
* Sr25519 (verify, batch verify)
* Secp256k1 (verify, recover pubkey)

Appropriate gas costs and upper limits (message size, batch size) should be added for each of these operations.

jberci:
Closes #368

### DIFF
--- a/contract-sdk/src/env.rs
+++ b/contract-sdk/src/env.rs
@@ -23,4 +23,19 @@ pub trait Env {
 pub trait Crypto {
     /// ECDSA public key recovery function.
     fn ecdsa_recover(&self, input: &[u8]) -> [u8; 65];
+
+    /// Verify an ed25519 message signature.
+    fn signature_verify_ed25519(&self, key: &[u8], message: &[u8], signature: &[u8]) -> bool;
+
+    /// Verify a secp256k1 message signature.
+    fn signature_verify_secp256k1(&self, key: &[u8], message: &[u8], signature: &[u8]) -> bool;
+
+    /// Verify an sr25519 message signature.
+    fn signature_verify_sr25519(
+        &self,
+        key: &[u8],
+        context: &[u8],
+        message: &[u8],
+        signature: &[u8],
+    ) -> bool;
 }
--- a/runtime-sdk/modules/contracts/src/lib.rs
+++ b/runtime-sdk/modules/contracts/src/lib.rs
@@ -153,6 +153,9 @@ pub struct GasCosts {
 
     // Crypto operations.
     pub wasm_crypto_ecdsa_recover: u64,
+    pub wasm_crypto_signature_verify_ed25519: u64,
+    pub wasm_crypto_signature_verify_secp256k1: u64,
+    pub wasm_crypto_signature_verify_sr25519: u64,
 }
 
 impl Default for GasCosts {
@@ -175,6 +178,9 @@ impl Default for GasCosts {
             wasm_env_query_base: 10,
 
             wasm_crypto_ecdsa_recover: 20,
+            wasm_crypto_signature_verify_ed25519: 20,
+            wasm_crypto_signature_verify_secp256k1: 20,
+            wasm_crypto_signature_verify_sr25519: 25,
         }
     }
 }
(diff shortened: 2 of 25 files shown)


contract-sdk/src/env.rs
```diff
@@ -23,4 +24,28 @@ pub trait Env {
...
+    ) -> bool;
+
+    /// Verify a batch of messages and signatures using the same public key.
+    ///
+    /// If the lengths of the contexts, messages and signatures slices don't
+    /// match or if there is some other error with the inputs, an Err(()) will
+    /// be returned. Otherwise, the booleans in the returned vector are the
+    /// verification results for each triplet of inputs.
+    #[allow(clippy::result_unit_err)]
+    fn signature_verify_batch(
```
Yawning, line 44: Whowa, that's an odd definition of batch verification.  Why the fixed public key.  Also AFAIK, we do not have a consensus-safe rust side ed25519 batch verify that follows our verification semantics...

Yawning:
Per discussion: The batch verification stuff should just go away, because it's a giant pain.

While I'm commenting on this:

The Ed25519 signature verification in oasis-core/runtime/src/common/crypto exposes our flavor of Ed25519 (with our janky domain separation), and our idea how to handle certain edge cases.

Is it ok that we are forcing this on end user code?  I suppose that the circumstances that lead us to use something non-standard at the consensus side in the first place haven't really changed ("Ledger doesn't support Ed25519ctx/Ed25519ph"), but actual contract authors might want Ed25519pure (as opposed to Ed25519-prehashed-oasis-flavor-because-ledger).

If we do want this, I would recommend using one of ZIP-215/FIPS 186-5/Oasis flavored Ed25519pure, which would require additional code (or pulling in a crate), with a slight preference towards one of the latter two.

kostko:
> The Ed25519 signature verification in oasis-core/runtime/src/common/crypto exposes our flavor of Ed25519 (with our janky domain separation), and our idea how to handle certain edge cases.

Yeah we should definitely not impose our domain separation scheme and we should expose Ed25519pure here, possibly the ZIP-215 one?

Yawning:
> Yeah we should definitely not impose our domain separation scheme and we should expose Ed25519pure here, possibly the ZIP-215 one?

I wish ZIP-215 was closer to the FIPS draft, though my preferred Ed25519 definition involves rejecting small-order-A.

Yawning:
Since we're going to expose Ed25519pure, the only primitive where a domain separation context makes sense is sr25519.  Should we nix the argument for now, or leave it?

Yawning:
The new API looks ok to me.

runtime-sdk/modules/contracts/src/lib.rs
```diff
@@ -153,6 +153,7 @@ pub struct GasCosts {
 
     // Crypto operations.
     pub wasm_crypto_ecdsa_recover: u64,
+    pub wasm_crypto_signature_verify: u64,
```
kostko, line 156: Should have per-scheme costs.

runtime-sdk/src/crypto/signature/ed25519.rs
```diff
@@ -56,6 +56,19 @@ impl PublicKey {
         sig.verify(&self.0, context, message)
             .map_err(|_| Error::VerificationFailed)
     }
+
+    /// Verify signature without applying domain separation.
+    pub fn raw_verify(&self, message: &[u8], signature: &Signature) -> Result<(), Error> {
```
kostko, line 61:
```suggestion
    pub fn verify_raw(&self, message: &[u8], signature: &Signature) -> Result<(), Error> {
```
(Also below.)

runtime-sdk/src/crypto/signature/secp256k1.rs
```diff
@@ -60,6 +64,17 @@ impl PublicKey {
             .verify_digest(digest, &sig)
             .map_err(|_| Error::VerificationFailed)
     }
+
+    /// Verify signature without using any domain separation schema.
```
kostko, line 68:
```suggestion
    /// Verify signature without using any domain separation scheme.
```

kostko (commented):
This should also add parameters for controlling maximum sizes of inputs and set them to some sensible defaults.

kostko (approved):

Yawning (approved):

PR #800, opened 2022-02-23
ci: cancel concurrent in-progress workflows on same branch

ptrus:
Default behavior is that in-progress runs are not canceled (e.g. when pushing new changes to a branch).


kostko (approved):

pro-wh:
yesssss goodbye 6-copies-of-all-dependabot-jobs situation

PR #802, opened 2022-02-24
contract-sdk: add macro to support single-crate contracts

nhynes:
This PR adds a wrapper around `create_contract!` that does not create a contract if the crate is being used as a dependency. This makes it easier to develop contract crates, as now there doesn't need to be a separate one full of types to avoid runtime symbols being generated.


contract-sdk-macros/src/lib.rs
```diff
@@ -24,3 +24,11 @@ pub fn event_derive(input: TokenStream) -> TokenStream {
     let input = syn::parse_macro_input!(input as syn::DeriveInput);
     event_derive::derive_event(input).into()
 }
+
+#[proc_macro]
+pub fn create_contract(input: TokenStream) -> TokenStream {
+    let contract_ident = syn::parse_macro_input!(input as syn::Ident);
+    std::env::var_os("CARGO_PRIMARY_PACKAGE")
```
nhynes, line 31: From the [docs](https://doc.rust-lang.org/cargo/reference/environment-variables.html):
> CARGO_PRIMARY_PACKAGE — This environment variable will be set if the package being built is primary. Primary packages are the ones the user selected on the command-line, either with -p flags or the defaults based on the current directory and the default workspace members. This environment variable will not be set when building dependencies. This is only set when compiling the package (not when running binaries or tests).

I used this env var in the first contract sdk for the same purpose. It was unstable then, so it's nice to see that they're keeping it.

ptrus (approved):
Nice, this is great!

kostko (approved):

PR #806, opened 2022-02-24
runtime-sdk-macros: automatically implement parameters query handler

ptrus:
TODO:
- [x] migrate rewards module to use the `sdk_derive(MethodHandler)`
- [x] implement go/ts bindings for parameters methods 
  - [x] will open a separate issue for implementing TS bindings after merging this (https://github.com/oasisprotocol/oasis-sdk/issues/812)
- [x] add tests

### DIFF
--- a/runtime-sdk-macros/src/method_handler_derive.rs
+++ b/runtime-sdk-macros/src/method_handler_derive.rs
@@ -116,11 +116,30 @@ pub fn derive_method_handler(impl_block: syn::ItemImpl) -> TokenStream {
         }
     };
 
+    let query_parameters_impl = {
+        quote! {
+            fn query_parameters<C: Context>(ctx: &mut C, _args: ()) -> Result<<Self as module::Module>::Parameters, <Self as module::Module>::Error> {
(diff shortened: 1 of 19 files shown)


runtime-sdk-macros/src/method_handler_derive.rs
```diff
@@ -116,11 +116,30 @@ pub fn derive_method_handler(impl_block: syn::ItemImpl) -> TokenStream {
         }
     };
 
+    let query_parameters_impl = {
+        quote! {
+            fn query_parameters<C: Context>(ctx: &mut C, _args: ()) -> Result<Parameters, Error> {
```
kostko, line 121: Where does the `Parameters` type come from?

ptrus: Hm, so this currently works because every module has the `Parameters` type defined in the same scope. Using the `Parameters` method of the `Module` trait would indeed be better, I'll change it to that, so: `<Self as module::Module>::Parameters`

kostko (approved):

kostko (approved):

PR #828, opened 2022-03-04
runtime-sdk: Exclude some transactions when scheduling

kostko:


peterjgilbert (approved):

Yawning (approved):

pro-wh (approved):
discussed over slack:

- we already run txes through check_tx when enqueuing them
- this is an additional pass during schedule+execute, which can give tighter results
- initial enqueuing check_tx pass makes no state changes, which is lower fidelity. we have no ordering chosen at that time, so updating state would make no sense
- this later pass, during schedule+execute, remembers changes per transaction, which allows this later check_tx to do a better job, e.g. detecting wrong nonce and inability to pay fee

PR #845, opened 2022-03-15
docs: Add list of Runtime transactions

matevz:


kostko:
Task linked: [CU-27wpk7p Document structure of UnsignedTxn and its sub structures to start Transaction Docs MVP](https://app.clickup.com/t/27wpk7p)

kostko:
Would be great if methods were grouped by kind (call, query) first.

PR #851, opened 2022-03-18
incoming message e2e tests

pro-wh:

PR #852, opened 2022-03-18
incoming message e2e tests

pro-wh:
reopened from #851 :facepalm:


pro-wh:
hm the transfer is failing. allegedly the Alice account in the runtime has insufficient balance. but was that amount to be transferred not just minted? did I accidentally set a transaction fee in the runtime tx?

> oasis-test-runner2258334103/e2e/test-runtime-simple-consensus/network/compute-1/console.log:{"msg":"incoming message transaction done","level":"WARN","ts":"2022-03-25T00:17:01.27576613Z","sdk_module":"dispatcher","mode":"execute_tx","ctx":"dispatch","module":"runtime-sdk","result_parsed":"Failed { module: \"accounts\", code: 2, message: \"insufficient balance\" }","id":0}

pro-wh:
after many corrections and some workarounds, this now passes

PR #855, opened 2022-03-22
Add confidentiality feature to the EVM module

nhynes:
This PR adds a feature flag, `confidential`, to the EVM module such that when it's enabled, all stores will be made to confidential storage. If the key manager isn't dispensing keys, the requester will receive zeros.


runtime-sdk/modules/evm/Cargo.toml
```diff
@@ -31,3 +31,6 @@ fixed-hash = "0.7.0"
 primitive-types = { version = "0.10.1", default-features = false, features = ["rlp", "num-traits"] }
 rlp = "0.5.1"
 uint = "0.9.1"
+
+[features]
+confidential = [] # Use confidential storage by default.
```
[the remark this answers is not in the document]
nhynes: It's not required to be a compile flag. It was a config option originally, and even just yesterday I considered switching back to that, but
* if it looks too configurable, people might try to switch from non-confidential to confidential or vice versa
* I wanted to keep the Emerald codepath almost completely unmodified since Sapphire is a bolt-on addition
* it's a bit ugly getting the module configuration at the point where the evm is calling the  storage methods

What's your thinking in favor of module config? The main one for me is easier testing with the current CI, but I'm not about to have github actions tie my hands.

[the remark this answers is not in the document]
kostko: Yes I was mostly thinking about making testing (in general) easier as it is more flexible if you don't need a special build.

runtime-sdk/src/storage/empty.rs
```diff
@@ -0,0 +1,63 @@
+use oasis_core_runtime::storage::mkvs;
+
+/// A key-value store that is always empty. Useful for testing,
+/// and when there is no better store is not available.
```
kostko, line 4:
```suggestion
/// and when there is no better store.
```

kostko (commented):
Ok, let's go with the feature flag for now.

Would be great to have some E2E tests as well just to make sure that this actually works.

runtime-sdk/modules/evm/src/state.rs
```diff
@@ -0,0 +1,121 @@
...
+        Some(kmgr_client) => kmgr_client,
+        None => return empty_store(),
+    };
+    let key_id = oasis_runtime_sdk::keymanager::get_key_pair_id(&[
+        CONFIDENTIAL_STORE_KEY_PAIR_ID_CONTEXT_BASE,
+        address.as_ref(),
+    ]);
+    let keypair = match kmgr_client.get_or_create_keys(key_id) {
+        Ok(keypair) => keypair,
+        Err(_) => return empty_store(),
```
kostko, line 67: This is actually a bad idea, it should return an aborting error instead (or panic). Otherwise this could make the node cause a discrepancy in case there is an error with the key manager (instead of indicating failure).

runtime-sdk/modules/evm/src/state.rs
```diff
@@ -0,0 +1,97 @@
...
+    let kmgr_client = ctx
+        .key_manager()
+        .expect("key manager must be available to use confidential storage");
+    let key_id = oasis_runtime_sdk::keymanager::get_key_pair_id(&[
+        CONFIDENTIAL_STORE_KEY_PAIR_ID_CONTEXT_BASE,
+        address.as_ref(),
+    ]);
+    let keypair = kmgr_client
+        .get_or_create_keys(key_id)
+        .expect("unable to retrieve confidential storage keys");
```
kostko, line 44: It is even better to propagate a proper aborting error (e.g. see the `Abort` error in the `contracts` module) as a panic will trigger a runtime restart.

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -97,6 +55,9 @@ pub trait Config: 'static {
     /// Token denomination used as the native EVM token.
     const TOKEN_DENOMINATION: token::Denomination;
 
+    /// Whether to use confidential storage by default.
+    const CONFIDENTIAL: bool;
```
kostko, line 59: Could have a default of `false`.

kostko (approved):

runtime-sdk/modules/evm/src/state.rs
kostko: > Why not modify dispatch_tx to also std::panic::catch_unwind and turn it into a batch failure? That would solve the symptom.

Ah yes, that would also work, can you open an issue for this?

PR #865, opened 2022-03-31
Configurable queries

### LINKED ISSUE #858: Support limiting query methods in config
There's currently a single config setting, `allow_expensive_queries`. When `true`, it has a double effect:
1. Allows some queries to run that would otherwise return `Forbidden`; it's hardcoded which queries those are.
2. Causes `core.EstimateGas` to simulate smart contracts; otherwise, only the non-smart-contract parts of each tx are looked at for gas estimation.

Under 1, we don't want a hardcoded list. And we want to not lump 1 and 2 under the same flag.

mitjat:
Closes #858

### DIFF
--- a/runtime-sdk/src/context.rs
+++ b/runtime-sdk/src/context.rs
@@ -22,6 +22,7 @@ use oasis_core_runtime::{
 use crate::{
     event::{Event, EventTag, EventTags},
     keymanager::KeyManager,
+    module::MethodHandler as _,
     modules::core::Error,
     runtime,
     storage::{self, NestedStore, Store},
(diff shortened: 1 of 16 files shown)


runtime-sdk/src/context.rs
```diff
@@ -83,16 +92,43 @@ pub trait Context {
...
+            Mode::ExecuteTx => true,
+            Mode::SimulateTx => self
+                .local_config(LOCAL_CONFIG_ESTIMATE_GAS_BY_SIMULATING_CONTRACTS)
+                .unwrap_or_default(),
+            // When just checking a transaction, we always want to be fast and skip contracts.
+            Mode::CheckTx => false,
         }
+    }
 
-        self.local_config(LOCAL_CONFIG_ALLOW_EXPENSIVE_QUERIES)
```
kostko, line 94: On second thought it may be good if we maintained backwards compatibility and use a similar default in case the older config option is present.

kostko (commented):
I think this looks good.

One (slightly unrelated) thing, in EVM module's `query_simulate_call` we should do something similar as in contract module's `query_custom` and that is to have a local config option called `query_simulate_call_max_gas` that would specify the max value `gas_limit` can have.

mitjat:
> One (slightly unrelated) thing, in EVM module's `query_simulate_call` we should do something similar as in contract module's `query_custom` and that is to have a local config option called `query_simulate_call_max_gas` that would specify the max value `gas_limit` can have.

Created https://github.com/oasisprotocol/oasis-sdk/pull/885/files to keep the scope of this PR cleaner.

runtime-sdk/src/context.rs
mitjat: :+1: done

mitjat:
This is ready for a final look.

Besides the changes discussed in comments, I also realized I never added the new error code (`Forbidden`, in `core`) to the client SDKs. I added a constant to TS now, but it looks like we don't use those constants anywhere, and in golang we don't provide any interpretation of raw error codes whatsoever. LMK if I missed something and I can update.

kostko (approved):

pro-wh (commented):
lg on ts-web additions :+1:

PR #872, opened 2022-04-05
cli: Default to legacy derivation on ledger

matevz:
Default oasis cli derivation path on ledger is currently ADR8 and is inconsistent with Oasis web/ext wallets. This PR defaults to legacy, so we avoid issues when users don't see correct accounts and balances using defaults.


Yawning (approved):

kostko (approved):

kostko:
Task linked: [CU-2gf4pgf Add support to new Oasis CLI](https://app.clickup.com/t/2gf4pgf)

PR #876, opened 2022-04-06
runtime-sdk: Panic on malformed local config

mitjat:
Currently, if a runtime's local config has a value that is syntactically valid but of an incorrect type (e.g. `0x132` instead of `"0x123"`), the config entry will be silently ignored.

I was thinking this would be nice to have as I was walking through the code ... Then chose not to get sidetracked ... Then a day later, got bitten by hidden bug caused by a mistyped local-config value :)

I think panicking, rather than warning, is the right thing to do from a UX perspective. It's unfortunate that we panic on the first use of each config value, rather than at initialization time, but I don't think we can do better as long as `LocalConfig` remains schema-less and module initialization is static.

*Testing*: Intentionally changed the type of a local config entry used by the core module, ran e2e test, observed the `client-0` node panic.

### DIFF
--- a/runtime-sdk/src/context.rs
+++ b/runtime-sdk/src/context.rs
@@ -106,10 +106,18 @@ pub trait Context {
             return None;
         }
 
-        self.host_info()
-            .local_config
-            .get(key)
-            .and_then(|v| cbor::from_value(v.clone()).ok())
+        self.host_info().local_config.get(key).and_then(|v| {
+            cbor::from_value(v.clone()).unwrap_or_else(|e| {
+                let msg = format!(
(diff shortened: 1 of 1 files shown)


pro-wh (approved):
ok yeah

kostko:
Would be better to also log something.

kostko:
The runtime is managed by `oasis-node` so it is not of limited value as eventually (https://github.com/oasisprotocol/oasis-core/issues/1819) it will process and propagate those logs, making it easier to actually see what's going on.

kostko (approved):

PR #888, opened 2022-04-14
runtime-sdk: Permanently enable schedule control

kostko:
Given upcoming oasisprotocol/oasis-core#4665 which will mandate runtime schedule control, this makes schedule control always enabled.


client-sdk/go/modules/core/types.go
```diff
@@ -27,6 +27,7 @@ type GasCosts struct {
 // Parameters are the parameters for the consensus accounts module.
 type Parameters struct {
 	MaxBatchGas        uint64                                   `json:"max_batch_gas"`
+	MaxTxSize          uint32                                   `json:"max_tx_size,omitempty"`
```
ptrus, line 30: Maybe add a comment that this one has `omitempty` for backwards comparability, and can be removed later?

ptrus (approved):

PR #889, opened 2022-04-14
More CLI commands/improvements

Yawning:
Requires:
- [x] https://github.com/oasisprotocol/oasis-core/pull/4733


Yawning:
Probably should checkpoint this, before I add even more commands and this becomes a unreasonably large doom PR.

pro-wh (approved):
always :+1: for cli

PR #893, opened 2022-04-19
Bump Rust dependencies and minor test cleanup

kostko:


runtime-sdk/modules/contracts/src/abi/oasis/test.rs
```diff
@@ -37,6 +37,20 @@ fn test_validate_and_transform() {
             "malformed code shoud fail validation"
         );
 
+        // Malformed WASM code.
+        let code = [
+            0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00, 0x01, 0x06, 0x01, 0x60, 0x01, 0x7e,
```
pro-wh, line 42: what's this from?

pro-wh (approved):
ok, only indirect dependency bumps. as long as this works :crossed_fingers:

PR #894, opened 2022-04-20
runtime-sdk/modules/contracts: Propagate errors in WASM lib calls

kostko:


pro-wh (commented):
I haven't worked with the contracts module, probably gonna rely on someone else to review.

question about this PR though: we do a lot of error propagation through `Result`s, whereas here it's an `Option` in a context structure. Is it better this way? Maybe are there limitations with having it as part of an error type? It looks like, for one thing, we're locked into a wasm3-specific error enum.

kostko:
> question about this PR though: we do a lot of error propagation through Results, whereas here it's an Option in a context structure. Is it better this way? Maybe are there limitations with having it as part of an error type? It looks like, for one thing, we're locked into a wasm3-specific error enum.

The problem is that this is about propagating errors from inside functions that are called from within WASM (e.g. imported WASM modules provided by the Oasis ABI). You cannot directly propagate an error as you can only trigger an interpreter abort. So we need to store the error in the execution context structure that can be inspected after the overall call into WASM returns.

runtime-sdk/modules/contracts/src/abi/oasis/storage.rs
```diff
@@ -166,29 +166,42 @@ fn get_instance_store<'a, C: Context>(
     // Determine which store we should be using.
     let store_kind: StoreKind = store_kind.try_into().map_err(|_| wasm3::Trap::Abort)?;
 
-    Ok(store::for_instance(
-        ec.tx_context,
-        ec.instance_info,
-        store_kind,
-    )?)
+    let instance_store = store::for_instance(ec.tx_context, ec.instance_info, store_kind);
```
ptrus, line 169: Should the crypto functions also be updated to return nicer errors (e.g. if invalid input parameters are provided). Can also be done in a later PR.

ptrus (approved):

runtime-sdk/modules/contracts/src/abi/oasis/storage.rs
kostko: Good idea, will add those as well.

PR #903, opened 2022-04-25
ts-web: fix consensusGetSignerNonce return type (can be undefined)

lukaw3d:
Example that returns undefined
```ts
await oasisClient.consensusGetSignerNonce({
  account_address: oasis.staking.addressFromBech32('oasis1qqqnpm9l8khrj9zaqzmt537gvnx6y38ntsc8jdrh'),
  height: 0,
})
```

(And playground already handles undefined return: https://github.com/oasisprotocol/oasis-sdk/blob/037546906fa06cae3983fb6b9b6ebadeaf63ac4e/client-sdk/ts-web/core/playground/src/index.js#L53-L56
)


pro-wh:
wow why is this possible? it's typed as returning uint64 from Go

I thought the `?? 0` was a vestige from when this used to get the whole account and look at the possibly-absent nonce field

pro-wh:
cross reference https://github.com/grpc/grpc-web/pull/1230

lukaw3d:
Added references

pro-wh:
thanks!

PR #920, opened 2022-05-06
ts-web: testing set runtime version in network fixture

pro-wh:
otherwise it expects 0.0.0 and none of the nodes launch the runtime

regressed in #915


pro-wh:
nope still spinning

pro-wh:
image

:angry: why ci gotta be like that

pro-wh:
image

it even passes, but why doesn't the step complete

pro-wh:
tried running it locally, and it didn't hang there

pro-wh:
I noticed that the video capture was going at 0.05 fps so I tried disabling that. now it finishes. gonna add a commit to disable that throughout. we don't even upload the videos.

re-enable locally for local testing, although I suppose you'd sooner use `cypress open` in that case

pro-wh:
same problem with benchmark-rt. I'm going to disable that for now. if anyone wants to have it start running again, make changes similar to how sample-run-network patches the fixture. or maybe make the net runner work on its own again, that'd be neat

Yawning (approved):

ptrus (approved):

Yawning:
I'm just merging this now, because this issue is blocking other stuff.

Issue #922, opened 2022-05-09
ts-web: new core types


pro-wh:
there's quite some diffs in the code generated from Go. let's apply them and make a release

pro-wh:
we did, tsk github not closing this

PR #923, opened 2022-05-09
ts-web: reflect updates

### LINKED ISSUE #822: ts-web/core/reflect-go: make sure reflect-go is updated for oasis-core v22.0
Oasis-core was updated to v22.0 (https://github.com/oasisprotocol/oasis-sdk/pull/816). `reflect-go` was updated so that it compiles, but I didn't check in depth if all the relevant services/types/structs of the oasis-core v22.0 are traversed by the package.

pro-wh:
fixes #822 #922 

- maintaining the list of types from transaction method bodies is too much work, so let's read from the registered methods list. but unfortunately it's unexported, so it's some kind of ugly `go:linkname` thing that I found out the idea of online

up to 22.1.5

### DIFF
--- a/client-sdk/ts-web/core/reflect-go/main.go
+++ b/client-sdk/ts-web/core/reflect-go/main.go
@@ -20,6 +20,7 @@ import (
 	"time"
 	_ "unsafe"
 
+	beacon "github.com/oasisprotocol/oasis-core/go/beacon/api"
 	"github.com/oasisprotocol/oasis-core/go/common/cbor"
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
 	"github.com/oasisprotocol/oasis-core/go/common/node"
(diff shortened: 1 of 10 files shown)


pro-wh:
adding manual client updates too

pro-wh:
adding manual module updates too

client-sdk/ts-web/core/reflect-go/main.go
```diff
@@ -504,46 +500,32 @@ func write() {
 	}
 }
 
-func main() {
-	visitType(reflect.TypeOf((*beacon.EpochTime)(nil)).Elem())
+// might be nicer to add a function to list these in oasis-core
```
kostko, line 503: Yeah please open up an issue in Oasis Core.

kostko (approved):

client-sdk/ts-web/core/reflect-go/main.go
pro-wh: 👍 filed https://github.com/oasisprotocol/oasis-core/issues/4747

PR #926, opened 2022-05-11
Add call envelope encryption to confidential EVM module

nhynes:


runtime-sdk/src/context.rs
```diff
@@ -296,6 +302,9 @@ pub trait BatchContext: Context {
 
 /// Runtime SDK transaction context.
 pub trait TxContext: Context {
+    /// The index of the transaction in the block.
```
xmzheng, line 305: do you mean `in the batch` instead of `in the block` here? How can client know the exact index of a transaction in the block?

nhynes: It's the batch. The client can't know the index in the batch, but it doesn't need to. The index is used to determine the nonce of the message returned by the paratime. The nonce is always included in the envelope (both ways, in fact).

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -551,6 +566,57 @@ impl<Cfg: Config> Module<Cfg> {
...
+        ctx: &C,
+        data: Vec<u8>,
+        tx_index: usize,
+        assume_km_reachable: bool,
+    ) -> Result<Option<(Vec<u8>, callformat::Metadata)>, Error> {
+        if !Cfg::CONFIDENTIAL {
+            return Ok(Some((data, callformat::Metadata::Empty)));
+        }
+        let call = cbor::from_slice(&data)
+            .map_err(|_| CoreError::InvalidCallFormat(anyhow::anyhow!("invalid packed call")))?;
```
kostko, line 605: Hm, so this will prevent one to use the non-backwards-compatible confidential mode where the encryption happens at the outer layer (e.g. in the actual transaction) as in that case `data` will already be the raw data that EVM expects. We could solve this by storing the original call format in the transaction context (e.g. in `with_tx`).

Then the above check could be changed to:
```rust
if !Cfg::CONFIDENTIAL || ctx.tx_call_format() != CallFormat::Plain {
```
And this would support both variants (would be good to document these two variants in a comment somewhere in this module).

runtime-sdk/src/testing/keymanager.rs
```diff
@@ -6,12 +6,25 @@ pub use oasis_core_keymanager_api_common::{
 
 use crate::keymanager::KeyManager;
 
-#[derive(Clone, Default)]
-pub struct MockKeyManagerClient {}
+#[derive(Default)]
+pub struct MockKeyManagerClient {
+    keys: std::sync::Mutex<std::collections::HashMap<KeyPairId, KeyPair>>,
```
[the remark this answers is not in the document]
nhynes: But that increases the file size by two lines! ;)

runtime-sdk/modules/evm/src/lib.rs
nhynes: Good point. I hadn't considered that people would submit non-web3 compat txs to an Eth-like runtime.

runtime-sdk/modules/evm/src/test.rs
```diff
@@ -280,6 +340,53 @@ impl<C: Config> Runtime for EVMRuntime<C> {
 fn do_test_evm_runtime<C: Config>() {
     let mut mock = mock::Mock::default();
     let mut ctx = mock.create_ctx_for_runtime::<EVMRuntime<C>>(context::Mode::ExecuteTx);
+    let client_keypair =
+        oasis_runtime_sdk::core::common::crypto::mrae::deoxysii::generate_key_pair();
+
+    // This is a macro to avoid mucking with borrow scopes.
+    macro_rules! encode_data {
```
kostko, line 347: These look duplicated from above?

nhynes: It is, but the macro uses idents from the lexical scope (a closure, in some sense) to remove several arguments. Decl macros can't generate macros, so it's copy-pasted. I don't feel bothered by this kind of duplication.

Making a generic function absolutely doesn't work. That's the first thing I tried, but then rustc ends up having trouble determining what kind of `Context` I want.

kostko (approved):

nhynes:
Ooops, automerge strikes again. I'll put those tests in a new PR

PR #931, opened 2022-05-12
tests/benchmark: fix benchmarks on 21.3.5

ptrus:


Yawning (approved):

pro-wh:
oh cool, no special patching step needed. thanks for the fix

PR #933, opened 2022-05-13
ts-web/core: version 0.1.0-alpha9

pro-wh:
for compatibility with oasis-core 22.1.x

the API changes are to parts not known to be in widespread use.

there's also a nice typing change for the typescript users


kostko (approved):

pro-wh:
ok as long as required tests pass and applicable tests pass

PR #947, opened 2022-05-25
ts-web: docs updates

pro-wh:
that getting-started doc was way out of date

- client class has moved
- public node urls changed
- our sample now uses the cors way


kostko (approved):

pro-wh:
:facepalm: and I still misspelled tightened

PR #949, opened 2022-05-25
ts-web/rt: v0.2.1-alpha.1

pro-wh:
typescript changes from #790

turns out we had additions from before, which we never made a release since. using those as the spotlight change

the version numbering is updated, see #948


kostko (approved):

pro-wh:
ok as long as required tests pass and applicable tests pass

PR #962, opened 2022-06-01
Xmzheng/client sdk callformat lib

xmzheng:
add encryption transport lib to web-ts. Add same test logic for derive_symmetric_key for golang, rust and typescript.


client-sdk/ts-web/rt/src/callformat.ts
```diff
@@ -1,4 +1,105 @@
...
+    cf: types.CallFormat,
+    cfg?: EncodeConfig,
+): Promise<[types.Call, metaEncryptedX25519DeoxysII?]> {
+    switch (cf) {
+        case types.CALLFORMAT_PLAIN:
+            return [call, undefined];
+        case types.CALLFORMAT_ENCRYPTED_X25519DEOXYSII:
+            if (cfg && cfg.publicKey) {
+                const privateKey = ed.utils.randomPrivateKey();
+                const publicKey = ed.curve25519.scalarMultBase(privateKey);
```
pro-wh, line 44: could we move this code for generating a keypair into the mrae file, to even out the level of abstraction here

client-sdk/ts-web/rt/src/callformat.ts
```diff
@@ -1,4 +1,105 @@
...
+ * decodeResult performs result decoding based on the specified call format metadata.
+ */
+
+export function decodeResult(
+    result: types.CallResult,
+    meta?: metaEncryptedX25519DeoxysII,
+): types.CallResult {
+    if (meta == undefined) {
+        /**
+         * In case of plain-text data format, we simply pass on the result unchanged.
```
pro-wh, line 86: use `// ` for this kind of inline comments that aren't documentation

client-sdk/ts-web/rt/package.json
```diff
@@ -16,6 +16,9 @@
     "dependencies": {
         "@oasisprotocol/client": "^0.1.0-alpha9",
         "elliptic": "^6.5.3",
+	"@noble/ed25519": "^1.6.0",
```
pro-wh, line 19: we've been trying to use tweetnacl for ed25519 stuff. which when we last compared with noble's, tweetnacl was better audited. would it be possible to use that?

client-sdk/ts-web/rt/src/callformat.ts
```diff
@@ -1,4 +1,105 @@
+import * as oasis from '@oasisprotocol/client';
+
+import * as ed from '@noble/ed25519';
+import * as types from './types';
+import * as mrae from './mrae';
+
+var deoxysii = require('deoxysii');
```
pro-wh, line 7: stick with imports

client-sdk/ts-web/rt/src/callformat.ts
```diff
@@ -1,4 +1,105 @@
+import * as oasis from '@oasisprotocol/client';
+
+import * as ed from '@noble/ed25519';
```
pro-wh, line 3: could you rearrange these for consistency with the rest of the codebase? I looked around and I mostly see:

- other packages from this monorepo
- external packages
- blank line
- other files from this package

client-sdk/ts-web/rt/src/callformat.ts
```diff
@@ -1,4 +1,105 @@
...
+
 /**
  * Call data key pair ID domain separation context base.
  */
 export const CALL_DATA_KEY_PAIR_ID_CONTEXT_BASE = 'oasis-runtime-sdk/private: tx';
+
+/**
+ *  EncodeConfig is call encoding configuration.
+ */
+export interface EncodeConfig {
```
pro-wh, line 17: is this converted from a Go or Rust type? those should go in types.ts, so that we can automatically generate them in the future.

client-sdk/ts-web/rt/src/callformat.ts
```diff
@@ -1,4 +1,105 @@
...
+            } else {
+                throw new Error('callformat: runtime call data public key not set');
+            }
+    }
+}
+
+/**
+ * decodeResult performs result decoding based on the specified call format metadata.
+ */
+
```
pro-wh, line 79: remove empty line between doc string and the thing it's attached to

xmzheng:
@pro-wh thanks for your comments, I will update my diff. 

this diff related to other implementations: https://github.com/oasisprotocol/oasis-core/pull/4772

client-sdk/ts-web/rt/src/mrae.ts
```diff
@@ -0,0 +1,53 @@
+import * as oasis from '@oasisprotocol/client';
+import * as ed from '@noble/ed25519';
+import {sha512_256} from 'js-sha512';
+
+var deoxysii = require('deoxysii');
+
+const BoxKDFTweak = 'MRAE_Box_Deoxys-II-256-128';
```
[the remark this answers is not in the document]
nhynes: > UPPERCASE_UNDERSCORES

[SCREAMING_SNAKE_CASE](https://en.wikipedia.org/wiki/Snake_case)

client-sdk/ts-web/rt/src/callformat.ts
```diff
@@ -1,4 +1,105 @@
...
+ *  EncodeConfig is call encoding configuration.
+ */
+export interface EncodeConfig {
+    /**
+     * publicKey is an optional runtime's call data public key to use for encrypted call formats.
+     */
+    publicKey?: types.KeyManagerSignedPublicKey;
+}
+
+export interface metaEncryptedX25519DeoxysII {
```
nhynes, line 24:
```suggestion
export interface MetaEncryptedX25519DeoxysII {
```

for consistency with TS naming conventions, and link to the Go src please

client-sdk/ts-web/rt/src/callformat.ts
```diff
@@ -1,4 +1,105 @@
...
+}
+
+/**
+ * encodeCall encodes a call based on its configured call format.
+ * It returns the encoded call and any metadata needed to successfully decode the result.
+ */
+export async function encodeCall(
+    call: types.Call,
+    cf: types.CallFormat,
+    cfg?: EncodeConfig,
```
nhynes, line 36:
```suggestion
    format: types.CallFormat,
    config?: EncodeConfig,
```
otherwise it's too many `c`s and `f`s

client-sdk/ts-web/rt/src/callformat.ts
```diff
@@ -1,4 +1,105 @@
...
+): Promise<[types.Call, metaEncryptedX25519DeoxysII?]> {
+    switch (cf) {
+        case types.CALLFORMAT_PLAIN:
+            return [call, undefined];
+        case types.CALLFORMAT_ENCRYPTED_X25519DEOXYSII:
+            if (cfg && cfg.publicKey) {
+                const privateKey = ed.utils.randomPrivateKey();
```
nhynes, line 43: Generating a fresh keypair every time is unnecessarily expensive. Could you take the keypair as a parameter? As long as the nonce is high quality, the security properties are similar.

The (Go, Rust, and now TS) `boxSeal` impls are also unnecessarily expensive since they always re-evaluate the shared secret, but the way it's done is deeply entrenched by this point (although a cache could help).

client-sdk/ts-web/rt/src/callformat.ts
```diff
@@ -1,4 +1,105 @@
...
+                crypto.getRandomValues(nonce);
+                const rawCall = oasis.misc.toCBOR(call);
+                const sealedCall = mrae.boxSeal(
+                    nonce,
+                    rawCall,
+                    null,
+                    cfg.publicKey.key,
+                    privateKey,
+                );
+                const envolope: types.CallEnvelopeX25519DeoxysII = {
```
nhynes, line 55:
```suggestion
                const envelope: types.CallEnvelopeX25519DeoxysII = {
```

client-sdk/ts-web/rt/src/callformat.ts
```diff
@@ -1,4 +1,105 @@
...
+export async function encodeCall(
+    call: types.Call,
+    cf: types.CallFormat,
+    cfg?: EncodeConfig,
+): Promise<[types.Call, metaEncryptedX25519DeoxysII?]> {
+    switch (cf) {
+        case types.CALLFORMAT_PLAIN:
+            return [call, undefined];
+        case types.CALLFORMAT_ENCRYPTED_X25519DEOXYSII:
+            if (cfg && cfg.publicKey) {
```
nhynes, line 42: please replace this with
```typescript
if (cfg?.publicKey === undefined) {
    throw new Error('callformat: runtime call data public key not set');
}
```
and dedent the interior for readability

client-sdk/ts-web/rt/src/callformat.ts
```diff
@@ -1,4 +1,105 @@
...
+export function decodeResult(
+    result: types.CallResult,
+    meta?: metaEncryptedX25519DeoxysII,
+): types.CallResult {
+    if (meta == undefined) {
+        /**
+         * In case of plain-text data format, we simply pass on the result unchanged.
+         */
+        return result;
+    } else {
```
nhynes, line 89: unnecessary `else`

client-sdk/ts-web/rt/src/callformat.ts
```diff
@@ -1,4 +1,105 @@
...
+        if (result.unknown) {
+            const envelop = oasis.misc.fromCBOR(
+                result.unknown,
+            ) as types.ResultEnvelopeX25519DeoxysII;
+            const pt = mrae.boxOpen(envelop.nonce, envelop.data, null, meta.pk, meta.sk);
+            const output = oasis.misc.fromCBOR(pt) as types.CallResult;
+            return output;
```
nhynes, line 96:
```suggestion
            return oasis.misc.fromCBOR(pt) as types.CallResult;
```
saves thinking about a variable

client-sdk/ts-web/rt/src/mrae.ts
```diff
@@ -0,0 +1,53 @@
+import * as oasis from '@oasisprotocol/client';
+import * as ed from '@noble/ed25519';
+import {sha512_256} from 'js-sha512';
+
+var deoxysii = require('deoxysii');
```
nhynes, line 5: `import`

client-sdk/ts-web/rt/src/mrae.ts
```diff
@@ -0,0 +1,53 @@
...
+ * boxSeal boxes ("seals") the provided additional data and plaintext via
+ * Deoxys-II-256-128 using a symmetric key derived from the provided
+ * X25519 public and private keys.
+ */
+export function boxSeal(
+    nonce: Uint8Array,
+    plainText: Uint8Array,
+    associateData: Uint8Array,
+    publicKey: Uint8Array,
+    privateKey: Uint8Array,
```
nhynes, line 31: Can you use something like
```ts
type Nonce = Opaque<Uint8Array>;
type PublicKey = Opaque<Uint8Array>;
type SecretKey = Opaque<Uint8Array>;
```
or take named args via an object so that it's harder to mix up these five positional args of the same type?

[docs for `Opaque`](https://github.com/sindresorhus/type-fest/blob/HEAD/source/opaque.d.ts)

client-sdk/ts-web/rt/src/mrae.ts
```diff
@@ -0,0 +1,53 @@
...
+ * boxOpen unboxes ("opens") the provided additional data and plaintext via
+ * Deoxys-II-256-128 using a symmetric key derived from the provided
+ * X25519 public and private keys.
+ */
+export function boxOpen(
+    nonce: Uint8Array,
+    ciperText: Uint8Array,
+    associateData: Uint8Array,
+    publicKey: Uint8Array,
+    privateKey: Uint8Array,
```
nhynes, line 48: similarly

client-sdk/ts-web/rt/src/mrae.ts
```diff
@@ -0,0 +1,53 @@
...
+ */
+export function boxSeal(
+    nonce: Uint8Array,
+    plainText: Uint8Array,
+    associateData: Uint8Array,
+    publicKey: Uint8Array,
+    privateKey: Uint8Array,
+): Uint8Array {
+    const sharedKey = deriveSymmetricKey(publicKey, privateKey);
+    var aead = new deoxysii.AEAD(sharedKey);
```
nhynes, line 34: This also has a non-trivial setup cost such that the cipher could (should) be cached. It's worse because the JS DeoxysII is not hw accelerated. For consistency with the other impls, though, I won't make a fuss about this. This is also not the lowest hanging fruit–just annoying.

client-sdk/ts-web/rt/test/mrae.test.ts
```diff
@@ -0,0 +1,16 @@
+import * as oasisRT from './../src';
+import * as ed from '@noble/ed25519';
+import * as oasis from '@oasisprotocol/client';
+
+describe('mrae', () => {
+  describe('symmetricKey', () => {
+    it('Should drive symmetric key correctly', async () => {
+      const privateKeyHex = "c07b151fbc1e7a11dff926111188f8d872f62eba0396da97c0a24adb75161750";
```
nhynes, line 8: Can you link to where you found this test vector?

client-sdk/ts-web/rt/test/mrae.test.ts
```diff
@@ -0,0 +1,16 @@
+import * as oasisRT from './../src';
+import * as ed from '@noble/ed25519';
```
nhynes, line 2: sort and space

client-sdk/ts-web/rt/test/mrae.test.ts
xmzheng: I just use some random same string at: https://github.com/oasisprotocol/oasis-core/pull/4772
```
 oasis-core/runtime/src/common/crypto/mrae/deoxysii.rs
 oasis-core/go/common/crypto/mrae/deoxysii/asymmetric_test.go
```

client-sdk/ts-web/rt/src/callformat.ts
pro-wh: aha

client-sdk/ts-web/rt/src/mrae.ts
pro-wh: this whole sdk is built to operate directly on types that show up from grpc. it's all Uint8Array. keys, hashes, encoded structures, quantities, IDs, addresses, aaaaaaaaaaaa

https://github.com/oasisprotocol/oasis-sdk/tree/a0da10b705b91aaf902dfa3aa0562af25a32f38c/client-sdk/ts-web/core#design-notes design note number 2

pro-wh (commented):
thanks for the changes so far

client-sdk/ts-web/rt/package.json
```diff
@@ -16,7 +16,9 @@
     "dependencies": {
         "@oasisprotocol/client": "^0.1.0-alpha9",
         "elliptic": "^6.5.3",
-        "sha3": "^2.1.4"
+	"js-sha512": "^0.8.0",
+        "sha3": "^2.1.4",
+	"tweetnacl": "^1.0.3"
```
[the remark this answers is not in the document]
xmzheng: How did you notice the difference? any plugin?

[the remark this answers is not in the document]
pro-wh: github uses 4-column spaces, so it's visible

client-sdk/ts-web/rt/src/callformat.ts
```diff
@@ -1,4 +1,92 @@
...
+ * It returns the encoded call and any metadata needed to successfully decode the result.
+ */
+export async function encodeCall(
+    nonce: Uint8Array,
+    sk: Uint8Array,
+    pk: Uint8Array,
+    call: types.Call,
+    format: types.CallFormat,
+    config?: EncodeConfig,
+): Promise<[types.Call, MetaEncryptedX25519DeoxysII?]> {
```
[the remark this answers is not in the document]
nhynes: Strictly speaking, you could use something like [this](https://www.typescriptlang.org/play?#code/FAOwhgtgpgzgDmAxlABAWSgFzAEzNlAb2BVJSgA84B7AJ0xUwE85UAFAGzAEsQUBeFCACuHDgG4SZSjXqMWqABoAmAKyqAjAE4AIlGoUmMAJLGBRKWTIBrKEwBcQ4RABGUWpKsBfSV+Ay6BmZWdCxcfDBzDGw8bAA6Th4+AB9QmIi4lXVtPQMjU0lQSFgEZBQAYTAxADE6CHwLKwC5YPYuXiiw2LAE9pBLcipA+RCszV19QxMzQWIrK2h07EdCIWLHAHIKNXGAWhxJo25uDZQfAasQahBkRwBVXkwADgBBWlowJk950lZ3e8er3en2+pD8TSGLQUKAA6gALfBQABu7nMcx+KEW4WWKA23Bw1g2oPm+OsjhgmFovAA5sS-H5mkFoZUanUGoIWRxarR6phMjscodpihUpzuby4vDESjaMBgAAzYQ3TDca7kG7UA4AHgAKoNMFAQDgYEQsd1HNFsWAvAA+AAUmEcOoAlE6ANoAIjNEQ9AF1GihaFhhLQ+HzvdhgH5-BqDnb0aQI2AVihbA4UAAWM4AGgGVxuUEcICgAHcUA8QM83h8mHa3RpsyhlI2AMy+525qx-WhF0vlwHVz51jON1SNgBs7dzXmdQA) to implement generic associated types. Go doesn't have that ability, which is why it's `interface{}`.

With the aim of not suggesting sweeping changes to the codebase, `unknown` is fine since at least it won't be a breaking change when other metadatas are added. It'll just be inconvenient to inspect the metadata.

client-sdk/ts-web/rt/src/mrae.ts
```diff
@@ -1,26 +1,26 @@
...
  * box primitives from the provided X25519 public and private keys.
  */
 
 export function deriveSymmetricKey(publicKey: Uint8Array, privateKey: Uint8Array): Uint8Array {
-    const pmk = ed.curve25519.scalarMult(oasis.misc.toHex(privateKey), oasis.misc.toHex(publicKey));
-    var kdf = sha512_256.hmac.create(BoxKDFTweak);
+    const pmk = nacl.scalarMult(privateKey, publicKey);
+    var kdf = sha512_256.hmac.create(BOX_KDF_TWEAK);
     kdf.update(pmk);
     return oasis.misc.fromHex(kdf.hex());
```
nhynes, line 18: why not save a parsing step and use [`arrayBuffer`](https://github.com/emn178/js-sha512/blob/b6c443143f4a32cbd62b18bfc8f984c4b54c97f5/index.d.ts#L24)

```js
return new Uint8Array(kdf.arrayBuffer());
```

client-sdk/ts-web/rt/src/callformat.ts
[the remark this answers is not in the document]
pro-wh: image

I guess I'll check on that sample later today :speak_no_evil:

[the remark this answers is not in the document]
pro-wh: ugh we used to do this thing in Go where there would be an overarching struct with a named field for each variant.

ok I loaded up the sample, but it's not clear how to go between associated Metadata.* and CallFormat.* type. (seems incidental and slightly abusive that CallFormat.Plain = Metadata.Plain.)

another answer would be to have multiple typescript declaration overloads of such an encode function. as nhynes says though, feel free not to and use `unknown` instead

I think it's fine that these types come out of encode relatively opaque. users will only pass them verbatim to submit and decode, respectively, neither of which internally could take advantage more type info

client-sdk/ts-web/rt/src/callformat.ts
```diff
@@ -1,4 +1,93 @@
...
+): Promise<[types.Call, unknown]> {
+    switch (format) {
+        case transaction.CALLFORMAT_PLAIN:
+            return [call, undefined];
+        case transaction.CALLFORMAT_ENCRYPTED_X25519DEOXYSII:
+            if (config?.publicKey === undefined) {
+                throw new Error('callformat: runtime call data public key not set');
+            }
+            const rawCall = oasis.misc.toCBOR(call);
+            const sealedCall = mrae.boxSeal(nonce, rawCall, null, config.publicKey.key, sk);
```
pro-wh, line 49: philosophical question about the layers of abstraction here:

- this callformat asks for x25519 deoxysii
- the primitive we've currently selected to occupy the mrae ts module is x25519 deoxysii
- we use the mrae module here

but the mrae module is named agnostically. does that mean we're not sure it'll always be x25519 deoxysii? if we change out the selected crypto primitive, how would we do it? change it directly and fix this callsite? add the new primitive in a new module instead? redefine the callformat as well?

or is the binding of x25519 deoxysii to the oasis projects' notion of mrae a permanent thing? 

on the go side https://github.com/oasisprotocol/oasis-sdk/blob/a0da10b705b91aaf902dfa3aa0562af25a32f38c/client-sdk/go/callformat/callformat.go#L11, the module is named deoxysii :shrug:

client-sdk/ts-web/rt/src/callformat.ts
```diff
@@ -1,4 +1,93 @@
...
+            throw new Error(`callformat: unsupported call format: ${format}`);
+    }
+}
+
+/**
+ * decodeResult performs result decoding based on the specified call format metadata.
+ */
+export function decodeResult(
+    result: types.CallResult,
+    meta?: MetaEncryptedX25519DeoxysII,
```
pro-wh, line 75: did I leave this comment earlier? sorry if this is a duplicate.

so following on from the discussion of supporting other encodings, it may not be possible to tell what encoding was used just from this kind of metadata. for example, if there's a different encryption algorithm that also uses a public and private key in the metadata, you wouldn't be able to tell which algorithm to use.

the Go side uses different structs (instead of interfaces) and the runtime type information, which we don't have in typescript.

maybe we could have decodeResult also take a `format: types.CallFormat` parameter. would that be appropriate for your use cases? I'm imagining users will call `encodeCall`, send the tx, then call `decodeResult` in pretty close proximity, and they'll have the desired CallFormat handy in both places

client-sdk/ts-web/rt/src/callformat.ts
```diff
@@ -1,4 +1,93 @@
...
+}
+
+/**
+ * encodeCall encodes a call based on its configured call format.
+ * It returns the encoded call and any metadata needed to successfully decode the result.
+ */
+export async function encodeCall(
+    nonce: Uint8Array,
+    sk: Uint8Array,
+    pk: Uint8Array,
```
pro-wh, line 36: these are pretty specialized for the deoxysii encryption. I see you've added these in response to nhynes's request to make a keypair reusable. I would like for us still to have a variant that internally generates the nonce and keys though, so that we have parity with the go client sdk's functionality.

at this point, the function signature so specialized that maybe it would be better to have separate encoding+decoding functions for each supported callformat. do you have any use cases in mind where the callsite doesn't know which callformat is chosen and would prefer to use a general encodeCall+decodeResult function?

client-sdk/ts-web/rt/src/mrae.ts
pro-wh: how much setup cost is there in this setup? I see from the syntax that this creates an object, which must have some kind of state. is the setup separable from the lifetime of this state at all?

if it is, maybe we could expose something like nacl's 'box before' and 'box/open after' system https://www.npmjs.com/package/tweetnacl/v/1.0.3#naclboxbeforetheirpublickey-mysecretkey

pro-wh (commented):
some philosophical stuff

client-sdk/ts-web/rt/src/callformat.ts
```diff
@@ -1,4 +1,93 @@
...
+        const envelop = oasis.misc.fromCBOR(result.unknown) as types.ResultEnvelopeX25519DeoxysII;
+        const pt = mrae.boxOpen(envelop.nonce, envelop.data, null, meta.pk, meta.sk);
+        return oasis.misc.fromCBOR(pt) as types.CallResult;
+    } else if (result.fail) {
+        throw new Error(
+            `callformat: failed call: module :${result.fail.module} code: ${result.fail.code} message: ${result.fail.message}`,
+        );
+    }
+
+    throw new Error(`callformat: unexpected result: ${result.ok}`);
```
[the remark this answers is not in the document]
nhynes: what I usually do is attach the result/cause to what I throw. something like [this](https://github.com/oasislabs/parcel-nfts/blob/main/integration/src/utils.ts#L5-L8) can be used to keep TS happy.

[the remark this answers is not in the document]
pro-wh: source :+1:

nhynes (approved):

client-sdk/ts-web/rt/src/callformat.ts
pro-wh: format param added, thanks

client-sdk/ts-web/rt/test/callformat.test.ts
```diff
@@ -0,0 +1,34 @@
...
+import * as nacl from 'tweetnacl';
+
+describe('callformat', () => {
+  describe('encodeCall/DecodeResult', () => {
+    it('Should encode and decode the message correctly', () => {
+      const message = "I will find some random message here"
+      const pairs = nacl.box.keyPair()
+      const publicKey = pairs.publicKey
+      const rawCall: oasisRT.types.Call = {
+	format: oasisRT.transaction.CALLFORMAT_ENCRYPTED_X25519DEOXYSII,
```
pro-wh, line 11: some tabs here, could you convert to spaces

client-sdk/ts-web/rt/src/callformat.ts
```diff
@@ -1,4 +1,146 @@
...
+export interface MetaEncryptedX25519DeoxysII {
+    sk: Uint8Array;
+    pk: Uint8Array;
+}
+
+/**
+ * encodeCallWithNonceAndKeys encodes a call based on its configured call format.
+ * It returns the encoded call and any metadata needed to successfully decode the result.
+ */
+export function encodeCallWithNonceAndKeys(
```
pro-wh, line 38: in some cases, we're making cryptography-related operations unnecessarily async, e.g. hashing a public key into an address. this is so that we could more easily change implementations in the future, if a more preferred library comes out exposing an async interface. or if the primitives we use later become available in the async web crypto interface and we could use that.

on the other hand, in other places, such as converting between binary and bech32 addresses, it's a synchronous interface. I think these places are the ones where it's a more niche kind of operation, and there's a single well established library for doing it, and that library does it synchronously.

I think this particular case is closer to the first scenario. but this is a late coming comment, and I understand we're interested in merging this right away to support some ongoing work elsewhere, so eh

client-sdk/ts-web/rt/src/callformat.ts
pro-wh: added, thanks

client-sdk/ts-web/rt/src/callformat.ts
xmzheng: I am going to change the signature.

client-sdk/ts-web/rt/src/callformat.ts
```diff
@@ -1,4 +1,146 @@
...
+ * It returns the encoded call and any metadata needed to successfully decode the result.
+ */
+export async function encodeCall(
+    call: types.Call,
+    format: types.CallFormat,
+    config?: EncodeConfig,
+): Promise<[types.Call, unknown]> {
+    const nonce = randomBytes(deoxysii.NonceSize);
+    const keyPair = nacl.box.keyPair();
+    return await encodeCallWithNonceAndKeys(
```
xmzheng, line 93: do I need to use `await` here? I think I can directly return the Promise, right?

PR #976, opened 2022-06-07
docs: Move ADRs to separate repo

matevz:


pro-wh (approved):
new link works, thanks

PR #980, opened 2022-06-08
evm: add untested impl of signed queries

nhynes:


runtime-sdk/modules/evm/src/signed_query.rs
```diff
@@ -0,0 +1,248 @@
...
+    ])
+}
+
+fn hash_domain<Cfg: Config>() -> &'static [u8; 32] {
+    static DOMAIN_SEPARATOR: OnceCell<[u8; 32]> = OnceCell::new(); // Not `Lazy` because of generic.
+    DOMAIN_SEPARATOR.get_or_init(|| {
+        const DOMAIN_TYPE_STR: &str = "EIP712Domain(string name,string version,uint256 chainId)";
+        hash_encoded(&[
+            encode_bytes(DOMAIN_TYPE_STR),
+            encode_bytes("Sapphire ParaTime"),
```
kostko, line 135: This should be generic and possibly follow our domain separation context scheme? E.g.
```
oasis-runtime-sdk/evm: signed query
```

runtime-sdk/modules/evm/src/types.rs
```diff
@@ -43,6 +43,26 @@ pub struct SimulateCallQuery {
     pub address: H160,
     pub value: U256,
     pub data: Vec<u8>,
+    #[cbor(default)]
```
kostko, line 46:
```suggestion
    #[cbor(optional, default)]
```

runtime-sdk/modules/evm/src/signed_query.rs
```diff
@@ -0,0 +1,248 @@
...
+    context::Context, core::common::crypto::hash::Hash, modules::accounts::API as _,
+};
+
+use crate::{
+    state,
+    types::{Leash, SimulateCallQuery},
+    Config, Error,
+};
+
+pub(super) fn verify<C: Context, Cfg: Config>(
```
kostko, line 18: Add documentation comments.

Also explain in the relevant comments why the eth ABI is used for some of these structures (I assume it is for compatibility with the eth personal signing scheme?).

runtime-sdk/modules/evm/src/signed_query.rs
```diff
@@ -0,0 +1,248 @@
...
+    let invalid_signature = || Error::InvalidSignedQuery("invalid signature");
+    signature[64] = signature[64].saturating_sub(27); // Some wallets generate high `s`.
+    let sig =
+        recoverable::Signature::try_from(signature.as_slice()).map_err(|_| invalid_signature())?;
+    let signed_message = hash_call_toplevel::<Cfg>(&query);
+    let signer_key = sig
+        .recover_verify_key_from_digest_bytes(&signed_message.into())
+        .map_err(|_| invalid_signature())?;
+    let signer_addr_digest = Keccak256::digest(&signer_key.to_encoded_point(false).as_bytes()[1..]);
+    if &signer_addr_digest[12..] != query.caller.as_ref() {
```
kostko, line 38: Unify this with similar address recovery in `raw_tx`.

runtime-sdk/modules/evm/src/signed_query.rs
```diff
@@ -0,0 +1,248 @@
...
+    );
+    hash_encoded(&[
+        encode_bytes(CALL_TYPE_STR),
+        Token::Address(query.caller.0.into()),
+        Token::Address(query.address.0.into()),
+        Token::Uint(ethabi::ethereum_types::U256(query.value.0)),
+        Token::Uint(ethabi::ethereum_types::U256(query.gas_price.0)),
+        Token::Uint(query.gas_limit.into()),
+        encode_bytes(&query.data),
+        Token::Uint(hash_leash(query.leash.as_ref().unwrap() /* checked above */).into()),
```
kostko, line 115:
```suggestion
        Token::Uint(hash_leash(query.leash.as_ref().unwrap() /* checked in verify */).into()),
```

runtime-sdk/modules/evm/src/signed_query.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    /// This was generated using the `@oasislabs/sapphire-paratime` JS lib.
+    const SIGNED_QUERY: &str =
+"a2657175657279a764646174614401020304656c65617368a4656e6f6e63651903e76a626c6f636b5f686173685820c92b675c7013e33aa88feaae520eb0ede155e7cacb3c4587e0923cba9953f8bb6b626c6f636b5f72616e6765036c626c6f636b5f6e756d626572182a6576616c75655820000000000000000000000000000000000000000000000000000000000000002a6663616c6c65725411e244400cf165ade687077984f09c3a037b868f676164647265737354b5ed90452aac09f294a0be877cbf2dc4d55e096f696761735f6c696d69740a696761735f70726963655820000000000000000000000000000000000000000000000000000000000000007b697369676e617475726558412496b1b40cfa931d6e62f76148bdb37e0a775213d757d715d379e60dfadfd18602296c8480d117fce51a52e2e420dbc6017d1e42accb0e5393a78050fcca73d11c";
+
+    fn setup_context<C: Context>(ctx: &mut C, query: &SimulateCallQuery, omit_block: bool) {
+        let leash = query.leash.as_ref().unwrap();
+        let mut state = ctx.runtime_state();
+
+        // Set the nonce to what the leash requires.
```
kostko, line 166: Add a test with bad nonce.

kostko (commented):
Note that queries are currently explicitly forbidden from accessing any private state and are actually completely forbidden in secure builds.

The way this should probably be handled is to introduce an explicit method whitelist in the `Runtime` trait where the runtime could specify which query methods are still allowed.

runtime-sdk/modules/evm/src/signed_query.rs
nhynes: Okay, but the function is going to return a `[u8;32]` to avoid a copy.

nhynes: Screen Shot 2022-06-10 at 04 01 18

runtime-sdk/modules/evm/src/signed_query.rs
nhynes: Thanks. I submitted this patch at 3 am. I could barely keep my eyes open, and I had the feeling I was missing something :)

Issue #631, opened 2021-11-10
EVM module: broken gas estimation for some transactions (EIP-150)


ptrus:
The upstream EVM module run in [estimation mode](https://github.com/rust-blockchain/evm/blob/0fbde9fa7797308290f89111c6abe5cee55a5eac/runtime/src/lib.rs#L256-L257) overestimates the gas needed for transactions using CREATE calls that are affected by [EIP-150](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-150.md).
The costs are overestimated by ~1/64 of the estimate call `gas_limit` (which depends on the runtime configured `batch_gas_limit`). 

Additionally, when the estimation mode is disabled, the reported `used_gas` is less than the amount of gas that is actually required for the transaction to not run out of gas. Therefore, disabling the "estimation mode" doesn't fix the issue, as it underestimates the required gas costs.

Upstream tracking issue: https://github.com/rust-blockchain/evm/issues/8

---
Additional info:
- Seems like the upstream estimation code works correctly if the `gas_limit` during estimation is the exactly required gas limit of the transaction
  - Parity's frontier uses this by doing a [binary search for the correct gas limit](https://github.com/paritytech/frontier/blob/7be5145b286fb806204181ab16de03b83736e584/client/rpc/src/eth.rs#L1133-L1169) which seems bizarre.
- This is the suspicious part of the upstream code: https://github.com/rust-blockchain/evm/blob/43a595ac05fd133b0bd24b4b5923ab76ac12d2a2/src/executor/stack/executor.rs#L647-L663

PR #981, opened 2022-06-08
runtime-sdk: gas estimate binary search

ptrus:
Adds configurable gas estimation via binary search support. 

Enables fixing some EVM edge cases (https://github.com/oasisprotocol/oasis-sdk/issues/631) where gas estimation depends on actual gas limit used during estimation.

There are some EVM "gotcha's", hopefully the comments explain it well enough.

### DIFF
--- a/runtime-sdk/src/modules/core/mod.rs
+++ b/runtime-sdk/src/modules/core/mod.rs
@@ -144,6 +144,13 @@ pub struct TxSimulationFailure {
     code: u32,
 }
 
+impl TxSimulationFailure {
+    /// Returns true if the failure is "core::Error::OutOfGas".
+    pub fn is_error_core_out_of_gas(&self) -> bool {
+        self.module_name == MODULE_NAME && self.code == 12
+    }
+}
+
 impl Display for TxSimulationFailure {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         write!(f, "{}", self.message)
@@ -244,6 +251,10 @@ pub trait API {
 
     /// Takes and returns the stored transaction priority.
     fn take_priority<C: Context>(ctx: &mut C) -> u64;
+
+    /// Returns the configured max iterations in the binary search for the estimate
+    /// gas.
+    fn estimate_gas_search_max_iters<C: Context>(ctx: &C) -> u64;
 }
 
 /// Genesis state for the accounts module.
@@ -265,6 +276,14 @@ pub struct LocalConfig {
     /// batch will be used.
     #[cbor(optional, default)]
     pub max_estimated_gas: u64,
+
+    /// The maximum number of iterations of the binary search to be done when simulating contracts for
+    /// gas estimation in `core.EstimateGas`.
+    /// The special value of 0 means that binary search won't be performed, and the transaction will be
+    /// simulated using maximum possible gas, which might return an overestimation in some special cases.
+    /// This setting should likely be kept at 0, unless the runtime is using the EVM module.
+    #[cbor(optional, default)]
+    pub estimate_gas_search_max_iters: u64,
 }
 
 /// State schema constants.
@@ -283,6 +302,10 @@ pub trait Config: 'static {
     const DEFAULT_LOCAL_MIN_GAS_PRICE: once_cell::unsync::Lazy<
         BTreeMap<token::Denomination, u128>,
     > = once_cell::unsync::Lazy::new(BTreeMap::new);
+
+    /// Default local estimate gas max search iterations configuration that is used in case no overrides
+    /// are set in the local per-node configuration.
+    const DEFAULT_LOCAL_ESTIMATE_GAS_SEARCH_MAX_ITERS: u64 = 0;
 
     /// Methods which are exempt from minimum gas price requirements.
     const MIN_GAS_PRICE_EXEMPT_METHODS: once_cell::unsync::Lazy<BTreeSet<&'static str>> =
@@ -407,6 +430,13 @@ impl<Cfg: Config> API for Module<Cfg> {
             .take()
             .unwrap_or_default()
     }
+
+    fn estimate_gas_search_max_iters<C: Context>(ctx: &C) -> u64 {
+        ctx.local_config(MODULE_NAME)
+            .as_ref()
+            .map(|cfg: &LocalConfig| cfg.estimate_gas_search_max_iters)
+            .unwrap_or(Cfg::DEFAULT_LOCAL_ESTIMATE_GAS_SEARCH_MAX_ITERS)
+    }
 }
 
 #[sdk_derive(MethodHandler)]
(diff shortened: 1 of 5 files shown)


runtime-sdk/src/modules/core/mod.rs
```diff
@@ -133,6 +133,19 @@ pub struct TxSimulationFailure {
     code: u32,
 }
 
+impl TxSimulationFailure {
+    /// Returns true if the failure is "core::Error::OutOfGas" or "evm::OutOfGas".
+    pub fn is_error_out_of_gas(&self) -> bool {
+        (self.module_name == MODULE_NAME && self.code == 12)
+            || (self.module_name == "evm" && self.code == 2 && self.message.contains("out of gas"))
```
kostko, line 140: Hm this doesn't seem ok as this now relies on implementation details of the `evm` module. Can we do this differently?

We could stop using the special "evm out of gas" error and use `core::Error::OutOfGas` instead (similar to what we do in the `contracts` module). Should be simple to do in the `From<evm::ExitError>` implementation.

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -133,6 +133,19 @@ pub struct TxSimulationFailure {
...
+    }
+
+    /// Returns true if the failure is "evm::Reverted".
+    pub fn is_error_evm_reverted(&self) -> bool {
+        self.module_name == "evm" && self.code == 8
```
kostko, line 145: Hm this doesn't seem ok as this now relies on implementation details of the `evm` module. Can we do this differently?

Is it possible to detect (in the `evm` module) that the revert was due to an out of gas condition? If yes, then we should remap to `core::Error::OutOfGas`, at least during simulation.

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -453,22 +468,89 @@ impl<Cfg: Config> Module<Cfg> {
...
+                    // Don't report success or failure. If the call fails, we still report
+                    // how much gas it uses while it fails.
+                    let gas_used = *tx_ctx.value::<u64>(CONTEXT_KEY_GAS_USED).or_default();
+                    Ok(gas_used)
+                })
             })
-        })
+        };
+
+        // Do a binary search for exact gas limit.
```
kostko, line 490: It would be great if we could increase unit test coverage of the below if not too hard.

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -453,22 +468,89 @@ impl<Cfg: Config> Module<Cfg> {
...
+        // If binary search is disabled the loop exits before doing the first iteration.
+        let (mut lo, mut hi) = (10_u128, args.tx.auth_info.fee.gas as u128); // Use u128 to avoid overflows when computing the mid point.
+        let (cap, mut iters, mut first_success) = (hi, 0, false);
+        while (lo + 1 < hi) && iters < bs_max_iters {
+            iters += 1;
+
+            let mid = (hi + lo) / 2;
+            args.tx.auth_info.fee.gas = mid as u64; // hi,lo are capped at Max::u64.
```
kostko, line 498:
```suggestion
            args.tx.auth_info.fee.gas = mid as u64; // hi,lo are capped at u64::MAX.
```

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -453,22 +468,89 @@ impl<Cfg: Config> Module<Cfg> {
...
+                        // Check if simulate with the returned estimate succeeds, so we can further clamp the high value.
+                        args.tx.auth_info.fee.gas = r;
+                        match simulate(args.tx.clone(), true) {
+                            Ok(_) => hi = r as u128,
+                            _ => continue,
+                        }
+                        // Check if the returned estimate is also the lower bound. In that case we can exit early.
+                        args.tx.auth_info.fee.gas = (hi - 1) as u64;
+                        match simulate(args.tx.clone(), true) {
+                            Err(_) => lo = hi - 1,
```
kostko, line 518: The early exit is due to `lo + 1 < hi` not holding as `lo + 1 = hi` after this. To make it more clear could either add a comment here or add an explicit loop break.

runtime-sdk/src/modules/core/mod.rs
ptrus: Yeah I think it shouldn't be too hard.

runtime-sdk/src/modules/core/mod.rs
ptrus: Makes sense, ll do. (I was actually surprised this was not done this way already).

runtime-sdk/src/modules/core/mod.rs
ptrus: I think not, it looks like the engine fails transactions that don't have enough of gas for EIP-150 with a revert instead of an "out-of-gas" (after increasing the gas used, the transaction passes). This might be a bug in the upstream evm crate code, will take a look.

ptrus: Fixed, by doing the same as frontier (doing one simulation at max gas and that way knowing if the transaction is failing due to gas limit or something else)

runtime-sdk/src/context.rs
```diff
@@ -126,6 +133,12 @@ pub trait Context {
         }
     }
 
+    /// Returns the maximum number of iterations when doing binary search for estimate gas.
+    fn estimate_gas_estimate_gas_search_max_iters(&self) -> u64 {
```
kostko, line 137: Should this be just `estimate_gas_search_max_iters`? Actually should we move this to the `core` module's `LocalConfig` (and add a method to `core::API` for other modules to use)?

ptrus: Could move it to core yes. I actually had this there initially, don't completely recall why I moved it. Probably due to needing it in the EVM module. But adding it to `core::API` does solve that.

ptrus: Fixed!

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -265,6 +276,14 @@ pub struct LocalConfig {
...
     #[cbor(optional, default)]
     pub max_estimated_gas: u64,
+
+    /// The maximum number of iterations of the binary search to be done when simulating contracts for
+    /// gas estimation in `core.EstimateGas`.
+    /// The special value of 0 means that binary search won't be performed, and the transaction will be
+    /// simulated using maximum possible gas, which might return an overestimation in some special cases.
+    /// This setting should likely be kept at 0, unless the runtime is using the EVM module.
+    #[cbor(optional, default)]
+    pub estimate_gas_search_max_iters: u64,
```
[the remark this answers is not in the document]
ptrus: Makes sense yes,

kostko (approved):

PR #984, opened 2022-06-09
runtime-sdk: Add support for signed queries

kostko:


runtime-sdk-macros/src/method_handler_derive.rs
```diff
@@ -232,6 +233,42 @@ pub fn derive_method_handler(impl_block: syn::ItemImpl) -> TokenStream {
...
+            .filter_map(|h| h.handler.as_ref())
+            .filter(|h| h.attrs.kind == HandlerKind::Call && h.attrs.allow_interactive)
+            .map(|h| h.attrs.rpc_name.clone())
+            .collect();
+        if handler_names.is_empty() {
+            quote! {}
+        } else {
+            quote! {
+                fn allowed_interactive_calls() -> Vec<&'static str> {
+                    vec![ #( #handler_names, )* ]
```
nhynes, line 266: similarly

runtime-sdk-macros/src/method_handler_derive.rs
```diff
@@ -232,6 +233,42 @@ pub fn derive_method_handler(impl_block: syn::ItemImpl) -> TokenStream {
...
+            .filter(|h| h.attrs.kind == HandlerKind::Query && h.attrs.allow_confidential)
+            .map(|h| h.attrs.rpc_name.clone())
+            .collect();
+        if handler_names.is_empty() {
+            quote! {}
+        } else {
+            quote! {
+                fn allowed_confidential_queries() -> Vec<&'static str> {
+                    vec![ #( #handler_names, )* ]
+                }
```
nhynes, line 249:
```suggestion
                fn allowed_confidential_queries() -> &'static [&'static str] {
                    &[ #( #handler_names, )* ]
                }
```
to avoid an allocation. You could also do
```
                fn allowed_confidential_queries() -> [&'static str; #handler_count] {
                    [ #( #handler_names, )* ]
                }
```
if you want to operate over them and don't like the double-deref required by `Iterator` methods.

Or you could do neither. I just hate using the system allocator. It's so slow.

runtime-sdk/src/dispatcher.rs
```diff
@@ -705,11 +764,20 @@ impl<R: Runtime + Send + Sync> transaction::dispatcher::Dispatcher for Dispatche
         method: &str,
         args: Vec<u8>,
     ) -> Result<Vec<u8>, RuntimeError> {
+        // Determine whether the method is allowed to access confidential state and provide an
+        // appropriately scoped instance of the key manager client.
+        let is_confidential_allowed = R::Modules::allowed_confidential_queries()
+            .iter()
+            .any(|&name| name == method);
```
nhynes, line 771: Right, so the reason I made this `Runtime::query_is_secure(method)` is because runtimes with a large number of queries called with about the same frequency could use a more efficient test like a `phf` lookup table, dynamic Huffman tree, or whatever. I originally had `Runtime::SECURE_QUERIES`, but it didn't seem incredibly easier for paratime devs one way or the other.

Also, I like this `R::Modules` approach. I didn't like how it was on `Runtime` for PT devs to configure, but I tried to rationalize your suggestion and ended up thinking it's a good thing that modules can't dictate what's secure. Maybe someone wants to run a c10l EVM that allows fetching encrypted storage (to "prove" that it's encrypted?) or disallowing calls because they don't like the KM implications. Perhaps an opt-out like `R::confidential_query_is_allowed` with a default implementation of `true` that gets ANDed with `is_confidential_allowed` suggested by the module would be useful.

runtime-sdk/src/modules/accounts/mod.rs
```diff
@@ -857,6 +857,22 @@ impl module::TransactionHandler for Module {
         ctx: &mut C,
         tx: &Transaction,
     ) -> Result<(), modules::core::Error> {
+        // Check whether the transaction is currently valid.
+        let round = ctx.runtime_header().round;
```
nhynes, line 861: Do we have a fork ID or something? This is useful, but not complete without domain separation.

runtime-sdk/src/modules/accounts/mod.rs
```diff
@@ -857,6 +857,22 @@ impl module::TransactionHandler for Module {
...
+        if let Some(not_before) = &tx.auth_info.not_before {
+            if round < *not_before {
+                // Too early.
+                return Err(modules::core::Error::ExpiredTransaction);
+            }
+        }
```
nhynes, line 867:
```suggestion
        if tx.auth_info.not_before.map(|nbf| round < nbf).unwrap_or_default() {
                return Err(modules::core::Error::ExpiredTransaction);
        }
```
I'm being pedantic here, but I do think that the `&` isn't necessary on like 862.

runtime-sdk/src/modules/accounts/mod.rs
kostko: All transactions are domain-separated by consensus layer chain context and runtime ID.

runtime-sdk-macros/src/method_handler_derive.rs
kostko: Good point. But currently the problem is the tuple implementations need to merge the lists.

I guess we don't really need to get a list of everything, we just need an iterator over all or even just a check that returns true if the given method is in the set.

runtime-sdk/src/dispatcher.rs
kostko: That's a good idea. Instead of placing it on the `Runtime` I would maybe add it to the core module's `Config` trait (currently I added a flag called `ALLOW_INTERACTIVE_READ_ONLY_TRANSACTIONS` that disables the whole thing -- defaults to disabled).

kostko: Also about `phf`, we actually used this in the past but it was somewhat clunky. It was for method dispatch so maybe this needs a revisit now.

runtime-sdk/src/dispatcher.rs
```diff
@@ -196,25 +219,33 @@ impl<R: Runtime> Dispatcher<R> {
             return (e.into_call_result(), call_format_metadata);
         }
 
+        // Make sure that a read-only call did not result in any modifications.
+        if call.read_only && ctx.runtime_state().has_pending_updates() {
```
nhynes, line 223: If you do this, then it's harder to reuse the normal `Call` method for queries. In my ideal paratime, there are no queries and people can submit read-only txs to any method they can, and get whatever results it produces–just with the caveat that it's not persisted.

From another perspective, is there really any security risk to call an explicitly allowlisted interactive call that _may_ modify state, but then revert it? I guess the runtime itself could revert its own state, but that's a little annoying.

nhynes (approved):

runtime-sdk-macros/src/method_handler_derive.rs
```diff
@@ -320,27 +364,45 @@ impl syn::parse::Parse for MethodHandlerAttr {
...
             if tag == "expensive" {
                 if kind != HandlerKind::Query {
                     return Err(syn::Error::new(
                         tag.span(),
                         "`expensive` tag is only allowed on `query` handlers",
                     ));
                 }
-                true
+                is_expensive = true;
+            } else if tag == "allow_confidential" {
```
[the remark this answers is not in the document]
kostko: Good idea, changed it to `allow_private_km`.

PR #878, opened 2022-04-07
Add runtime tx vectors for Ledger

matevz:
This PR adds test vectors for runtime transactions primarily used by Ledger developers including secp256k1 and sr25519 signers.

Also fixes #30.

PR #991, opened 2022-06-14
evm: add c10l EVM e2e test

nhynes:


runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -654,17 +654,17 @@ impl<Cfg: Config> Module<Cfg> {
         Self::get_storage(ctx, body.address, body.index)
     }
 
-    #[handler(query = "evm.Code")]
+    #[handler(query = "evm.Code", allow_confidential)]
```
kostko, line 657: Code isn't confidential though? So `allow_confidential` only enables access to confidential state via the key manager. Since this query does not access the key manager, this shouldn't be needed.

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -654,17 +654,17 @@ impl<Cfg: Config> Module<Cfg> {
...
     fn query_code<C: Context>(ctx: &mut C, body: types::CodeQuery) -> Result<Vec<u8>, Error> {
         Self::get_code(ctx, body.address)
     }
 
-    #[handler(query = "evm.Balance")]
+    #[handler(query = "evm.Balance", allow_confidential)]
```
kostko, line 662: Same as above.

runtime-sdk/modules/evm/src/lib.rs
nhynes: Thanks. May I then suggest renaming `allow_confidential` to `permit_km` or something?

client-sdk/go/modules/evm/signed_calls.go
```diff
@@ -0,0 +1,146 @@
+package evm
```
nhynes, line 1: this can be moved into the sapphire-paratime repo now that it's public

client-sdk/go/testing/testing.go
```diff
@@ -15,9 +15,10 @@ import (
 
 // TestKey is a key used for testing.
 type TestKey struct {
-	Signer  signature.Signer
-	Address types.Address
-	SigSpec types.SignatureAddressSpec
+	SecretKey [32]byte
```
kostko, line 18: Are all secret keys the same length? A slice would probably make more sense. Also you don't seem to set this for ed25519 test keys. You can get it by using something like:
```golang
ms := memorySigner.NewTestSigner(seed)
sk := ms.(signature.UnsafeSigner).UnsafeBytes()
```

runtime-sdk/modules/evm/src/signed_query.rs
```diff
@@ -29,24 +29,26 @@ pub(crate) fn verify<C: Context, C10lCfg: Config>(
...
 
     // Next, verify the leash.
     let current_block = ctx.runtime_header().round;
     let mut state = ctx.runtime_state();
-    let sdk_address = C10lCfg::map_address(query.caller.into());
-    let nonce = C10lCfg::Accounts::get_nonce(&mut state, sdk_address).unwrap();
-    if nonce < leash.nonce {
+    let sdk_address = Cfg::map_address(query.caller.into());
+    let nonce = Cfg::Accounts::get_nonce(&mut state, sdk_address).unwrap();
+    if nonce > leash.nonce {
```
kostko, line 51: So the leashed nonce is an upper bound? Why not just require equality?

client-sdk/go/testing/testing.go
kostko: Also @matevz added a similar thing in #878.

nhynes: I'll use Matevz's one

kostko (approved):

PR #993, opened 2022-06-15
ci: allow passing setup code to hash-rust

nhynes:
For me, this is to allow `rustup target add x86_64-fortanix-unknown-sgx` before building, as I cannot have a `rust-toolchain.toml` auto-install it because [actions-rs does not support it](https://github.com/actions-rs/toolchain/pull/209).

### DIFF
--- a/.github/actions/hash-rust/action.yml
+++ b/.github/actions/hash-rust/action.yml
@@ -24,6 +24,10 @@ inputs:
     description: Control whether to produce sgxs binaries
     default: no
     required: false
+  setup:
+    description: Code to run before building
+    default: ""
+    required: false
 outputs:
   hashes:
     description: Comma-separated list of binary hashes
@@ -39,6 +43,7 @@ runs:
       run: |
         docker run --rm -i -v ${{ inputs.dir }}:/src ${{ inputs.image }} /bin/bash <<-'EOF'
           set -e
+          ${{ inputs.setup }}
           cd /src
           CARGO_TARGET_ROOT="/src/target"
           TARGET=""
@@ -51,7 +56,7 @@ runs:
 
             for pkg in $PKG_DIRS; do
               pushd ${pkg}
-              cargo build --release --target $TARGET
+              cargo build --release --locked --target $TARGET
               cargo elf2sgxs --release
               popd
             done
@@ -60,7 +65,7 @@ runs:
 
             for pkg in $PKG_DIRS; do
               pushd ${pkg}
-              cargo build --release
+              cargo build --release --locked
               popd
             done
           fi


.github/actions/hash-rust/action.yml
```diff
@@ -39,19 +43,20 @@ runs:
...
+          ${{ inputs.setup }}
           cd /src
           CARGO_TARGET_ROOT="/src/target"
           TARGET=""
           PKG_DIRS=$(tr ',' ' ' <<< "${{ inputs.pkg-dirs }}")
           if [ "${{ inputs.sgx }}" == "yes" ]; then
             TARGET="x86_64-fortanix-unknown-sgx"
             export CARGO_TARGET_DIR="$CARGO_TARGET_ROOT/sgx"
-            export CFLAGS_x86_64_fortanix_unknown_sgx="-isystem/usr/include/x86_64-linux-gnu -mlvi-hardening -mllvm -x86-experimental-lvi-inline-asm-hardening"
+            export CFLAGS_x86_64_fortanix_unknown_sgx="-isystem/usr/include/x86_64-linux-gnu -mlvi-hardening -mllvm -x86-experimental-lvi-inline-asm-hardening -fstack-protector-all"
```
kostko, line 54: Why the `-fstack-protector-all`, this will not work because the Fortanix ABI currently sets FS and GS bases to the same value and the stack protector assumes the canary value is at FS:0x28 but with the Fortanix ABI that contains a copy of RSP. This causes stuff to crash due to canary mismatch.

nhynes: I just want to get _something_ to insert the `__stack_check_guard`. I believe that it's inserted into the code generated by `ring` (it would be `ring`), and isn't defined by LLVM during the top-level build because the global is the old way of doing things.

I put in https://github.com/oasislabs/sapphire-paratime/commit/866b1fd301df472f476768d89f95f0d9fc3f1688#diff-8383ec3e7666578793741ef1253328c1dc4bd0e3012176816493eb57e6d81773R10-R15, but I'm not sure if that works now.

nhynes: [This kind of thing is a perennial problem with fortanix](https://github.com/nhynes/rusqlite/commit/e75b82a248c5499807ccff7237b80c381ef126c6#diff-170350079cc8de1ef8188c0a93de33519dbfa8899965f1e9489e448790345a85R234230-R234245).

kostko: Regular `ring` doesn't work with SGX and the upstream author explicitly refuses to merge support for it despite other people implementing everything, for (IMO) stupid reasons. This is why we also got rid of `ring` everywhere. Why do you need `ring` anyway?

kostko:
> For me, this is to allow rustup target add x86_64-fortanix-unknown-sgx before building, as I cannot have a rust-toolchain.toml auto-install it because https://github.com/actions-rs/toolchain/pull/209.

But the build happens in a Docker container so whatever actions do shouldn't matter? The Docker image specifically uses the same Rust toolchain version as the SDK. We are planning on bumping that version btw.

nhynes:
> But the build happens in a Docker container so whatever actions do shouldn't matter?

The Rust action is in the sapphire-paratime repo and uses the `rust-toolchain` file found in the same. That `rust-toolchain` is also used within the docker container, but it doesn't install the sgx target because it's not a `rust-toolchain.toml`. Thus, I need to install the latter manually. I'm not using the same rust version as the SDK because I like new features.

kostko (approved):

PR #1021, opened 2022-06-24
runtime-sdk: Mask gas limit/wanted in case EMIT_GAS_USED_EVENTS is false

kostko:


runtime-sdk/modules/contracts/src/abi/oasis/mod.rs
```diff
@@ -158,7 +154,11 @@ impl<Cfg: Config> OasisV1<Cfg> {
                 // Compute how much gas was wanted.
                 let final_gas = gas::get_remaining_gas(instance);
                 let wanted_gas = initial_gas + exhausted_gas.saturating_sub(final_gas);
-                core::Error::OutOfGas(initial_gas, wanted_gas).into()
+                core::Error::out_of_gas::<<<C::Runtime as Runtime>::Core as core::API>::Config>(
```
pro-wh, line 157: :dizzy_face:

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -136,6 +136,19 @@ pub enum Error {
     TxSimulationFailed(#[from] TxSimulationFailure),
 }
 
+impl Error {
+    /// Generate a proper OutOfGas error, depending on whether the module is configured to emit gas
+    /// use information or not.
+    pub fn out_of_gas<Cfg: Config>(limit: u64, wanted: u64) -> Self {
+        if Cfg::EMIT_GAS_USED_EVENTS {
```
pro-wh, line 143: this already existed? what did it do before?

runtime-sdk/modules/contracts/src/abi/oasis/mod.rs
kostko: Yeah unfortunately the Rust compiler wants explicit casts here :(

runtime-sdk/src/modules/core/mod.rs
kostko: It suppresses emission of gas used events (which are by default emitted for every transaction). With this PR the same flag is also used for masking exact amounts in out of gas errors.

pro-wh: :+1:

pro-wh (approved):
ok that's a reasonable extension of the flag's meaning

PR #1037, opened 2022-07-06
runtime-sdk: add RNG

nhynes:
This PR adds a VRF-based RNG to the contract and EVM modules.


runtime-sdk/modules/contracts/src/abi/oasis/crypto.rs
```diff
@@ -91,33 +90,68 @@ impl<Cfg: Config> OasisV1<Cfg> {
...
 
-                ctx.instance
-                    .runtime()
-                    .try_with_memory(|memory| -> Result<_, wasm3::Trap> {
-                        let key = get_key(kind, key, &memory).map_err(|err| {
-                            ec.aborted = Some(Error::CryptoMalformedPublicKey);
-                            err
-                        })?;
-                        let message = Region::from_arg(message)
+                let rt = ctx.instance.runtime();
```
nhynes, line 93: nothing here changed. I just split `rt` into a var to reduce one layer of indentation for readability

runtime-sdk/src/crypto/random.rs
```diff
@@ -0,0 +1,49 @@
...
+        let round_header_hash = ctx.runtime_header().encoded_hash();
+        let ctx_pers = [
+            b"oasis-runtime-sdk/crypto: random_bytes".as_slice(),
+            nonce_bytes.as_slice(),
+            &[ctx.mode() as u8],
+            round_header_hash.as_ref(),
+        ];
+        let key_id =
+            crate::keymanager::get_key_pair_id(ctx_pers.into_iter().chain(pers.iter().copied()));
+        let keypair = km.get_or_create_keys(key_id).ok()?;
```
Yawning, line 25: I know there's domain separation and everything, but I would rather you didn't do this.

In an ideal (everyone has lots of time) world, the key manager could implement a real entropy source for these sorts of requests that's distinct from the master secret used to derive actual keys.

Alternatively, something like this is less... "abusive".
```
sekrit = keymanager.get_keypair_id(my_contract_key)
pi = ECVRF_prove(sekrit, h(pers | round_header_hash | nonce)
beta = ECVRF_proof_to_hash(pi) // <- Use this as the seed
```

runtime-sdk/src/crypto/random.rs
```diff
@@ -0,0 +1,49 @@
...
+        let seed = keypair.state_key.0; // This is an abuse of the key, surely, but entropy is entropy.
```
Yawning, line 26: At least run the entropy through a cryptographic hash function please.

runtime-sdk/src/crypto/random.rs
nhynes: Thanks for the review. I updated it to use the [`schnorkell` VRF](https://docs.rs/schnorrkel/0.10.2/schnorrkel/vrf/index.html) since the signature scheme seems already to be being used. Please let me know if this is a better use of primitives.

I also went and removed all of the user-provided personalization because it shouldn't be necessary as long as there's only one `Rng` created per block (and if I'm going to pretend to be a 3rdparty dev, I'm going to set it to `&[]` every time).

Yawning: Sure, it's a bit non-standard, but I have nothing against sr25519.

Yawning:
As a side note:

Do we want to expose a VRF construct somewhere so that it would be easy to make a chainlink VRF knockoff?

I think to do it "correctly", you would need prove, verify, and proof to hash (roughly speaking the chainlink design is a 2-step process: commit to alpha in one step, retrieve pi/beta in another step).  This is done so that properly coded applications can prevent users/executors/etc from fishing for an alpha value (because it is committed), to get a beta that produces a desired result.

runtime-sdk/modules/contracts/src/abi/oasis/crypto.rs
```diff
@@ -91,33 +90,62 @@ impl<Cfg: Config> OasisV1<Cfg> {
...
+                gas::use_gas(ctx.instance, cost)?;
+
+                let rt = ctx.instance.runtime();
+                rt.try_with_memory(|mut memory| -> Result<_, wasm3::Trap> {
+                    let output = Region::from_arg((dst_ptr, dst_len))
+                        .as_slice_mut(&mut memory)
+                        .map_err(|_| wasm3::Trap::Abort)?;
+                    Ok(ec
+                        .tx_context
+                        .rng()
```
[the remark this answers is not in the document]
nhynes: The idea is that the runtime module can decide what to do if there's no KM available. For the contract sdk, it'll return 0 bytes written, which is technically true. The EVM implementation will revert the subcall. In both cases, recovery is possible in a way that simply trapping doesn't.

runtime-sdk/src/crypto/random.rs
```diff
@@ -0,0 +1,94 @@
...
+            round_header_hash.as_ref(),
+        ]);
+        let km_kp = match km
+            .get_or_create_keys(key_id)
+            .ok()
+            .map(|ks| ks.input_keypair)
+        {
+            Some(kp) => kp,
+            None => {
+                return Self { state: None };
```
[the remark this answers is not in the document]
nhynes: why wouldn't one entity aborting not also be a discrepancy, just of a different kind?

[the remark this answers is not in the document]
kostko: Yeah this is exactly the thing.

If you just silently compute a different result such a discrepancy is subject to slashing for incorrect results.

But if the compute node knows that there was a problem while executing a batch it will instead indicate an explicit failure. Which may lead to availability slashing down the road (but this depends also whether other nodes were able to perform or not).

runtime-sdk/modules/contracts/src/abi/oasis/crypto.rs
bennetyee, line 144: i think there may be an issue here, but please confirm.

we don't implement commit-then-reveal.

this means that a contract call that does not commit -- the execution committee members can presumably speculatively execute transactions that never commit (?) -- that has visible side-effects such as events can leak information about the output of the RNG that we supply.  to ensure that this info leak is not useful, we should do key splitting, so that the VRF random value isn't used to directly seed all the RNGs in the speculative execution tree.   instead, it should be used to derive per-execution seeds (e.g., by being hashed with other values) that is dependent on both the input state to the transaction and the current transaction parameters, so that even if/when a speculative contract call can reveal RNG output from its generator, its generator's output is uncorrelated with the RNG output of any other contract call (speculative or otherwise).

even if execution committee members can't (easily) spawn speculative executions, it wouldn't hurt to be more careful here as defense-in-depth.

i see that we're using `keymanager::get_key_pair_id` using the `round_header_hash` -- and i may be (probably am) misunderstanding the code in runtime-sdk/src/crypto/random.rs -- so we might be dependent on the input state at the beginning of the transaction batch (?) but not of this particular transaction itself, and there doesn't seem to be any dependency on the current transaction's parameters.

Yawning:
> should clients provide additional pers strings? it might be more secure if used properly, but will people know (how) to use it?

I like having an integrated way to do domain-separation.  The case when nothing is provided should (still) be secure if this is done correctly.

runtime-sdk/modules/contracts/src/abi/oasis/crypto.rs
nhynes: ## Background information

It is true that the RNG is seeded once per transaction batch.

The seed depends on the execution mode (real vs simulation) and the `round_header_hash`, which itself is dependent on all of the inputs to the transaction batch, including the transactions and their ordering. The RNG is not reseeded before each transaction, so the state progression is linear. It remains advanced even if a tx reverts. The batch proposer can fish for good seeds by trying all permutations of txs or delaying to adjust the block timestamp; these items are reflected in the round header and therefore its hash.

The RNG may be replayed if a batch fails to commit during optimistic execution, at which point side effects will become visible. A batch must either commit or fail, so there's no backing out of a tx. The next batch will have a new `round_header_hash`, so the RNG seed will change. Proposers cannot 

## Discussion

**General MEV**  The compute committee can simulate transactions to observe side effects, and use this information to pick good batches. This form of MEV is general to all computations on the platform (and not just the RNG). We could eliminate MEV by either implementing commit-reveal at the platform layer or having a pre-execution phase where all committee members agree on the batch to execute before executing. Both of these low-level solutions impose high complexity and cost, especially to applications that are not vulnerable to speculation. Without significant protocol amendments, the best we can do is ask contract authors to write constant-time, pure code when dealing with secrets. A steep ask, surely.

**RNG**  The tx proposer can speculatively execute evil contracts that attempt to dump RNG state and, in the same tx, use that knowledge to extract value (this must happen in the same tx because changing the batch will change the round header hash and therefore the RNG seed). This will be successful if it's possible for a contract to infer RNG state based on a sequence of outputs, which is hopefully impossible given the construction of the CSPRNG and execution environment (but let's not rely on practical matters like inability to do number crunching within contracts).

To summarise,  it is my belief that this implementation of RNG is not unduly vulnerable to exploitation if devs are careful to write constant-side-effect code when dealing with secrets, which we already recommend for security-sensitive applications. The key mitigations to speculation are the dependence of of the RNG seed on the speculative tx and the resistance of the CSPRNG to inversion (and, of course, quality application code where required).

Please correct me if I'm mistaken.

bennetyee: Let me try to describe an attack scenario which might be feasible.

I think it's still the case that a compute commitee leader choses the batch for other committee members to execute.  Assume that this is feasible for the MEV simulate transaction approach -- I think that's what's going on in the **General MEV** above.  The leader is in cahoots with somebody who is running a lottery contract.  The lottery contract sells lottery tickets until some block height, and sometime after that (whenver a transaction from the owner triggers) uses the RNG to pick the winning numbers and emits a public event with the list of winners.  So the co-conspirators do the following:  (1) buy a large amount of tickets before the deadline.  (2) committee leader pick batches and simulate their execution, and see what events say are winners (and how much is won).  the RNG seeds change based on the transaction order, so there will be many different outcomes.  presumably there is a time limit, but many batches can be tried.  (3) pick the batch where the tickets that the co-conspirators' win the most money, and send that batch to the other (unsuspecting) committee members.  (4) if this sufficiently biases expected win, profit.  so that's the general MEV case.  arguably, this can probably be addressed by better lottery design, and auditing such applications to ensure that they use best practices, though we should probably provide this guidance somewhere.

given the description of how the seed is derived, I agree that RNG internal state cannot be extracted (assuming CSPRNG) so even if one or more co-conspirator written contract leak their outputs, those are useless for attacking the execution of other contracts having an execution in the same batch.  and since the seed depends on the "identity" of the batch (txn params, order, etc), outputs from many executions of leaking contracts will not help in attacking the execution of contracts in a different batch.

but does this also means that the seed is derivable by everyone?  is the dependency on the decrypted txn parameters or encrypted txn parameters?  if it's txn order and encrypted txn parameters, that's all public information after the batch commits, so while the txns cannot determine this info when they execute, after the batch commits everybody can determine the seed.  if that's the case, then there are many applications where this is not good enough.  for example:  poker where we need to shuffle the deck and deal down cards to multiple players.  if the seed is known after the batch commits, then every player can determine what's in the hands of every other player.

Yawning: In the attack scenario @bennetyee presents, "uses the RNG to pick the winning numbers and emits a public event with the list of winners" is the step where the contract loses protection against MEV, if it is done as part of a single Tx.  We absolutely "should probably provide this guidance somewhere", since it is likely non-obvious to most developers.

As far as the seed goes, the RNG uses the Merlin transcript RNG that is initialized with the random VRF output (beta, I assume) over a static string, with a key fetched from the key manager.

While the inputs into the key-pair-id derivation is public information (runtime-sdk/src/crypto/random.rs:25), the resulting key that is fetched from the key manager is not (runtime-sdk/src/crypto/random.rs:30) exposed anywhere.  As long as it is not possible to issue arbitrary key manager requests, this should be fine.

In the "hindsight is 20/20" department, I didn't envision the key manager being used this way, so neither of the key requests calls have added provisions for domain separation beyond the runtime ID and key pair ID.  Having an extra context here would have been nice to make the seed stuff "extra-safe", but as it stands, just audit every key manager key request in the runtime/sdk and make sure that domain separation is applied.

nhynes: Yeah, we should provide guidance on how to prevent MEV. We've so far been saying that there's no MEV, but really there is, but at least you can prevent it with careful contract design.

And as far as the seed is concerned, as Yawning mentions, the keypair used as input to the VRF is a domain-separated key derived from the KM's ~master~ main key, which is generated within the enclave. Right now, we're just manually auditing that domain separation is applied in the runtime-sdk and signing the KM policies that allow this kind of key use by cipphire.

I think the takeaway from this discussion is that we should prepare good docs on avoiding MEV (and references to to-be-written contract libraries that make it easier) alongside the release of the RNG. Maybe also make the marketing materials reflect the nuance...

runtime-sdk/Cargo.toml
```diff
@@ -19,7 +19,8 @@ hmac = "0.11.0"
 sha2 = "0.9.8"
 sha3 = { version = "0.10.1", default-features = false }
 k256 = { version = "0.10.4" }
-schnorrkel = "0.10.2"
+schnorrkel = { version = "0.10.2", features = ["asm", "avx2_backend"] }
```
nhynes, line 22: I assume these instruction sets are fast and supported (@Yawning).

runtime-sdk/modules/evm/src/backend.rs
```diff
@@ -174,6 +174,35 @@ impl<'ctx, C: Context, Cfg: Config> EVMBackend for Backend<'ctx, C, Cfg> {
...
+        (*self).random_bytes(num_words, pers)
+    }
+}
+
+impl<'ctx, C: Context, Cfg: Config> EVMBackendExt for Backend<'ctx, C, Cfg> {
+    fn random_bytes(&self, num_words: u64, pers: &[u8]) -> Vec<u8> {
+        // Refuse to generate more than 2 KiB in one go.
+        // EVM memory gas is checked only before and after calls, so we won't
+        // see the quadratic memory cost until after this call uses its time.
+        let num_words = num_words.min(64);
```
nhynes, line 192: I think that we'll see people hit this upper limit relatively frequently, as EVM gas costs disincentivize using sub-word sized values, and 52 is the size of a deck of cards. Maybe we should set this to 32 and force people to be efficient? (@kostko)

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -76,7 +76,7 @@ pub trait Config: 'static {
     /// If any of the precompile addresses returned is the same as for one of
     /// the builtin precompiles, then the returned implementation will
     /// overwrite the builtin implementation.
-    fn additional_precompiles() -> Option<precompile::PrecompileSetType> {
+    fn additional_precompiles() -> Option<()> {
```
kostko, line 79: Hm, why this change? I assume it is because the new precompile set type is now pretty hard to provide externally, but this seems like it makes the function useless as it is now impossible to specify additional precompiles.

I believe this is not currently used anywhere so I would be fine with just removing the function for now if you think this can no longer be implemented.

runtime-sdk/modules/evm/src/precompile/confidential.rs
```diff
@@ -58,53 +104,30 @@ pub(super) fn call_x25519_derive(
...
-        });
-    }
-
-    let mut words = input.array_chunks::<WORD>().take(SLOTS);
-    let key = words.next().unwrap();
-    let nonce_word = words.next().unwrap();
-    let text_len = words.next().unwrap();
-    let ad_len = words.next().unwrap();
+) -> Result<([u8; KEY_SIZE], [u8; NONCE_SIZE], Vec<u8>, Vec<u8>), PrecompileFailure> {
+    let mut call_args = ethabi::decode(
```
kostko, line 108: Does this break compatibility? Can we pretend nobody is using this yet?

runtime-sdk/modules/evm/src/backend.rs
```diff
@@ -174,6 +176,34 @@ impl<'ctx, C: Context, Cfg: Config> EVMBackend for Backend<'ctx, C, Cfg> {
...
+        let num_bytes = num_bytes.min(RNG_MAX_BYTES) as usize;
+        let mut ctx = self.ctx.borrow_mut();
+        ctx.rng(pers)
+            .ok()
+            .and_then(|mut rng| {
+                let mut rand_bytes = vec![0u8; num_bytes];
+                rand_core::RngCore::try_fill_bytes(&mut rng, &mut rand_bytes).ok()?;
+                Some(rand_bytes)
+            })
+            .unwrap_or_default()
```
kostko, line 203: So this will still ignore key manager errors? I would rather we either propagate the error (e.g. so the dispatcher sees the `Abort` error -- this will probably require storing it somewhere where it can be retrieved by the interpreter invocation and triggering an EVM abort) or if this is really hard to do, panic. Both will trigger an explicit failure instead of a silent discrepancy which is slashable.

runtime-sdk/modules/evm/src/precompile/confidential.rs
nhynes: > Does this break compatibility?

It does, but we can definitely pretend that nobody is using this yet. I'm very confident that if we checked, we'd find that nobody is using this yet. Either way, it's testnet, and this new ABI is much easier to use.

runtime-sdk/modules/evm/src/backend.rs
nhynes: There's no `Result`-returning function on the path from here to the runtime. Precompiles don't produce errors, I guess. I wasn't aware that panicking did anything special. I'll do that instead.

runtime-sdk/modules/evm/src/precompile/confidential.rs
kostko: Agreed about the ease of use, not sure why we didn't use this immediately? This will also require fixing the documentation and library, in teh sapphire repo, will you do the fixes there as well?

Mainnet deployment is imminent so let's get this to Testnet quickly and make sure that Mainnet doesn't have the old version.

runtime-sdk/modules/evm/src/backend.rs
kostko: > There's no Result-returning function on the path from here to the runtime

Right, that's why I mentioned that it would probably require storing an error somewhere and triggering an EVM abort (if possible) then the caller of the interpreter can check if an error is stored and propagate that. Not the nicest, but aside from panicking it's probably the only option.

> It's the ctx.rng() one that does.

Yes, so the `ctx.rng()` is the one that can produce an error in case there is a problem with key manager communication. Deterministic errors are fine to be `.unwrap_or_default()`, it's just the non-deterministic ones that are a problem (e.g. key manager communication failing for some reason).

runtime-sdk/modules/evm/src/lib.rs
nhynes: I didn't see that this was part of a trait. I added the option back.

runtime-sdk/modules/evm/src/precompile/confidential.rs
kostko: We should also make sure this ABI is used in #1176.

nhynes: > This will also require fixing the documentation and library, in teh sapphire repo, will you do the fixes there as well?

100% going to do that after this PR goes in and before the next upgrade.

> Agreed about the ease of use, not sure why we didn't use this immediately? 

Didn't think of it then, as I hadn't experienced first-hand how awful it was to write and maintain the Solidity assembly.

> We should also make sure this ABI is used in #1176.

I'll apply the changes to this one, too.

runtime-sdk/modules/contracts/src/abi/oasis/crypto.rs
```diff
@@ -91,33 +92,67 @@ impl<Cfg: Config> OasisV1<Cfg> {
...
+                    .and_then(|g| g.checked_add(ec.params.gas_costs.wasm_crypto_random_bytes_base))
+                    .unwrap_or(u64::max_value()); // This will certainly exhaust the gas limit.
+                gas::use_gas(ctx.instance, cost)?;
+
+                let rt = ctx.instance.runtime();
+                rt.try_with_memory(|mut memory| -> Result<_, wasm3::Trap> {
+                    let pers = Region::from_arg((pers_ptr, pers_len))
+                        .as_slice(&memory)
+                        .map_err(|_| wasm3::Trap::Abort)?;
+                    let mut rng = ec.tx_context.rng(pers).map_err(|_| wasm3::Trap::Abort)?;
```
kostko, line 148: Actually, this also needs to propagate the error, similarly to how it is done in `storage::get_instance_store`, e.g.:
https://github.com/oasisprotocol/oasis-sdk/blob/04944cbb7a3cf346acb2693b0a7dfda4ab0f9f63/runtime-sdk/modules/contracts/src/abi/oasis/storage.rs#L186-L194

nhynes: Ok, now it batch aborts any time a RNG/KM error happens.

kostko (approved):

contract-sdk/src/testing.rs
```diff
@@ -51,13 +57,23 @@ impl PublicStore for MockStore {}
...
     pub fn new() -> Self {
-        Self {}
+        Default::default()
+    }
+}
+
+impl Default for MockEnv {
+    fn default() -> Self {
+        Self {
+            rng: Arc::new(Mutex::new(XorShiftRng::seed_from_u64(0))),
```
Yawning, line 75: Should this be deterministic?

Yawning (commented):
I know you merged this, but 2 comments.

contract-sdk/src/testing.rs
nhynes: it helps if you need to hardcode values into tests or something

runtime-sdk/modules/evm/src/precompile/confidential.rs
jberci: > > We should also make sure this ABI is used in #1176.
> 
> I'll apply the changes to this one, too.

No need, I'll be doing this for the next rebase, and I've got cleanup in progress anyway.

runtime-sdk/Cargo.toml
nhynes: this is why we can't have nice things. I'll disable it :P

PR #1052, opened 2022-07-19
contracts: support changing contract upgrades policy

ptrus:
TODO:
- [x] add go/typescript client functionality

### DIFF
--- a/client-sdk/go/modules/contracts/types.go
+++ b/client-sdk/go/modules/contracts/types.go
@@ -133,6 +133,14 @@ type Upgrade struct {
 	Tokens []types.BaseUnits `json:"tokens"`
 }
 
+// ChangeUpgradePolicy is the body of the contracts.ChangeUpgradePolicy call.
+type ChangeUpgradePolicy struct {
+	// ID is the unique instance identifier.
+	ID InstanceID `json:"id"`
+	// UpgradesPolicy is the updated upgrade policy.
+	UpgradesPolicy Policy `json:"upgrades_policy"`
+}
+
 // CodeQuery is the body of the contracts.Code query.
 type CodeQuery struct {
 	// ID is the code identifier.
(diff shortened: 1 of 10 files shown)


client-sdk/go/modules/contracts/types.go
```diff
@@ -133,6 +133,16 @@ type Upgrade struct {
...
 }
 
+// ChangeUpgradePolicy is the body of the contracts.ChangeUpgradePolicy call.
+type ChangeUpgradePolicy struct {
+	// ID is the unique instance identifier.
+	ID InstanceID `json:"id"`
+	// UpgradesPolicy is the updated upgrade policy.
+	UpgradesPolicy Policy `json:"upgrades_policy"`
+	// Tokens that should be sent to the contract as part of the change upgrade policy call.
+	Tokens []types.BaseUnits `json:"tokens"`
```
kostko, line 143: Hm, why would one send tokens when updating the policy? The other calls have that when they are running the contract code to allow atomic payments as part of contract code execution.

runtime-sdk/modules/contracts/src/lib.rs
```diff
@@ -563,6 +565,31 @@ impl<Cfg: Config> Module<Cfg> {
...
+    #[handler(call = "contracts.ChangeUpgradePolicy")]
+    pub fn tx_change_upgrade_policy<C: TxContext>(
+        ctx: &mut C,
+        body: types::ChangeUpgradePolicy,
+    ) -> Result<(), Error> {
+        let params = Self::params(ctx.runtime_state());
+
+        <C::Runtime as Runtime>::Core::use_tx_gas(ctx, params.gas_costs.tx_change_upgrade_policy)?;
+
+        if !ctx.should_execute_contracts() {
```
kostko, line 577: This does not execute contracts so this check can be replaced with just a check for "check tx mode".

kostko (commented):
Could you also add support for this to the CLI?

kostko (approved):

PR #1056, opened 2022-07-22
orc tool: support for signing SGXS binaries

ptrus:
This adds support for offline singing of SGX enclaves, removing the need for  https://github.com/oasislabs/rust-sgx/tree/ravenac95/sgxs-sign-support-offline-sign

Changes:
- `orc sgx-sign-data <bundle.orc>` subcommand is added which constructs the SIGSTRUCT and outputs the hash to be signed (this replaces the `sgx-sign gendata` from the rust-sgx tool).
- `orc sgx-set-sig <bundle.orc> <signature.sig> <public_key.pub>` is updated so that it accepts the signature (of the SIGSTRUCT hash) and the signing public key. It then constructs the signed SIGSTRUCT and inserts it into the manifest (this replaces the two-step process from before, which used: `sgx-sign catsig` and `orc sgx-set-sig`)

At the moment the command flags for configuring the SIGSTRUCT fileds matches arguments in the rust-sgx tool: https://github.com/oasislabs/rust-sgx/blob/79ceb56714234f30576d8ccb61548828cfd934c4/sgxs-tools/src/bin/sgxs-sign.rs#L110-L121

NOTE: The above code says that: `"ATTRIBUTES.XFRM and inverse ATTRIBUTEMASK.XFRM fields (default: 0x3/0)"`, but due to a bug the `ATTRIBUTEMASK.XFRM` actually defaults to `0x3` (the command ignores the flag) and it uses the default defined here: https://github.com/oasislabs/rust-sgx/blob/79ceb56714234f30576d8ccb61548828cfd934c4/sgxs/src/sigstruct.rs#L80-L84

TODO:
- [x] requires release of oasis-core 22.1.9 with: https://github.com/oasisprotocol/oasis-core/pull/4860 included
- [x] test the offline signing flow (and compare results with https://github.com/fortanix/rust-sgx/compare/master...oasislabs:rust-sgx:ravenac95/sgxs-sign-support-offline-sign the `orc sgx-sign-data` should match the `sgx-sign gendata`) 
- [x] add documentation for the offline signing flow


Yawning (approved):

docs/runtime/reproducibility.md
```diff
@@ -124,3 +124,74 @@ You can omit the signature initially and add it later by using:
...
+##### Attach the singed SIGSTRUCT to the bundle
+
+With the signature in `sigstruct.sha256.sig` we can now generate a valid
+SIGSTRUCT and attach it into the bundle.
+
+```bash
+orc sgx-set-sig bundle.orc sigstruct.sha256.sig public.pem
+```
+
+If there are no errors, `bundle.orc` will now contain a valid SGX SIGSTRUCT
```
Yawning, line 196: Does the SDK orc tool have a dump/validate sub-command that will open, verify, and dump the manifest in a nice human readable format?  If not, one should exist, and if such a thing exists, should show how to confirm that there is a SIGSTRUCT in the newly altered bundle.

docs/runtime/reproducibility.md
```diff
@@ -124,3 +124,74 @@ You can omit the signature initially and add it later by using:
...
+#### Generate signing data for your enclave
+
+Generating signing data is done with the `orc sgx-gen-sign-data` subcommand,
+like so:
+
+```bash
+orc sgx-gen-sign-data [options] bundle.orc
+```
+
+_See `orc sgx-gen-sign-data --help` for details on available options_
```
kostko, line 160:
```suggestion
:::tip

See `orc sgx-gen-sign-data --help` for details on available options.

:::
```

tools/orc/sgx.go
```diff
@@ -34,3 +41,106 @@ func sgxVerifySignature(bnd *bundle.Bundle) error {
```
kostko, line 41: The `sgxVerifySignature` can probably now be removed as this is performed as part of the bundle validation?

docs/runtime/reproducibility.md
kostko: Yeah having an `orc show <bundle.orc>` would be really nice.

kostko (approved):

docs/runtime/reproducibility.md
kostko: Added the `show` subcommand and updated the docs.

PR #1064, opened 2022-07-31
evm: don't require encryption of empty string

nhynes:
This PR enables native token transfers from MetaMask (or any other unmodified client). ~It also allows constifying some gas costs/storage ops†.~

~There's another question of whether confidential EVM should pass through txs that fail to decrypt, assuming that they're not encrypted. That's not super safe for least reason of which that it won't be rejected by the client node before the entire network sees the plaintext.~

~† I don't think that we should enable these by default since they're possible for the application/client to mitigate, leaving applications that don't need/care about it lower costs. It's left as a config option just in case (and for others who may wish to use the module).~
 * ~the difference between non-zero/zero values in tx data can be fixed by~
    - ~having the call contain random bytes inserted into the call data to obfuscate the number of zero/non-zero bytes~
    - ~having the contract either calculate or caller specify the number of zeros so that the contract can waste gas to make up the differences~
* ~the difference between storage set and reset can be fixed by using a sentinel value instead of zero~

### DIFF
--- a/runtime-sdk/modules/evm/src/backend.rs
+++ b/runtime-sdk/modules/evm/src/backend.rs
@@ -277,9 +277,7 @@ impl<'c, C: Context, Cfg: Config> ApplyBackendResult for Backend<'c, C, Cfg> {
                         let val: H256 = value.into();
 
                         let ctx = self.ctx.get_mut();
-                        if value == primitive_types::H256::default()
-                            && (Cfg::CONFIDENTIAL && !Cfg::CONFIDENTIAL_CONSTIFY)
-                        {
+                        if value == primitive_types::H256::default() {
                             with_storage!(*ctx, &addr, |store| store.remove(&idx));
                         } else {
                             with_storage!(*ctx, &addr, |store| store.insert(&idx, val));
--- a/runtime-sdk/modules/evm/src/backend.rs
+++ b/runtime-sdk/modules/evm/src/backend.rs
@@ -277,7 +277,9 @@ impl<'c, C: Context, Cfg: Config> ApplyBackendResult for Backend<'c, C, Cfg> {
                         let val: H256 = value.into();
 
                         let ctx = self.ctx.get_mut();
-                        if value == primitive_types::H256::default() {
+                        if value == primitive_types::H256::default()
+                            && (Cfg::CONFIDENTIAL && !Cfg::CONFIDENTIAL_CONSTIFY)
+                        {
                             with_storage!(*ctx, &addr, |store| store.remove(&idx));
                         } else {
                             with_storage!(*ctx, &addr, |store| store.insert(&idx, val));
--- a/runtime-sdk/modules/evm/src/lib.rs
+++ b/runtime-sdk/modules/evm/src/lib.rs
@@ -14,7 +14,7 @@ use evm::{
     executor::stack::{MemoryStackState, PrecompileFn, StackExecutor, StackSubstateMetadata},
     Config as EVMConfig,
 };
-use once_cell::sync::Lazy;
+use once_cell::sync::OnceCell;
 use thiserror::Error;
 
 use oasis_runtime_sdk::{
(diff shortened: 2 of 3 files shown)


runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -625,6 +616,37 @@ impl<Cfg: Config> Module<Cfg> {
...
+    /// Returns the decrypted call data or `None` if this transaction is simulated in
+    /// a context that may not include a key manager (i.e. SimulateCall but not EstimateGas).
+    fn decode_call_data<C: Context>(
+        ctx: &C,
+        data: Vec<u8>,
+        format: transaction::CallFormat,
+        tx_index: usize,
+        assume_km_reachable: bool,
+    ) -> Result<Option<(Vec<u8>, callformat::Metadata)>, Error> {
+        if !Cfg::CONFIDENTIAL || format != transaction::CallFormat::Plain || data.is_empty() {
```
nhynes, line 628: I moved this around. The only change is `|| data.is_empty()`

xmzheng: do we also need encryption for 'create' calls?

xmzheng: did I miss anything here? Shouldn't it be `format == transaction::CallFormat::Plain`, then we return the data immediately?

nhynes: sapphire treats plain/unencrypted as encrypted, for compatibility with an unmodified eth envelope

runtime-sdk/modules/evm/src/backend.rs
```diff
@@ -277,7 +277,9 @@ impl<'c, C: Context, Cfg: Config> ApplyBackendResult for Backend<'c, C, Cfg> {
                         let val: H256 = value.into();
 
                         let ctx = self.ctx.get_mut();
-                        if value == primitive_types::H256::default() {
+                        if value == primitive_types::H256::default()
+                            && (Cfg::CONFIDENTIAL && !Cfg::CONFIDENTIAL_CONSTIFY)
```
kostko, line 281: Shouldn't this be:
```suggestion
                            && (!Cfg::CONFIDENTIAL || !Cfg::CONFIDENTIAL_CONSTIFY)
```

Also do I see this right that if `CONFIDENTIAL_CONSTIFY` is enabled, there is no way to remove stuff from storage? This doesn't seem like the best thing.

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -69,6 +71,48 @@ pub trait Config: 'static {
             address.as_ref(),
         )
     }
+
+    fn evm_config(estimation: bool) -> &'static EVMConfig {
```
kostko, line 75: Add documentation comment.

runtime-sdk/modules/evm/src/backend.rs
nhynes: This is a bit aggressive in terms of constifying, I agree. There are a few levels of constification. I think the big ones are 1) time/gas, 2) io data, 3) io patterns. `CONFIDENTIAL_CONSTIFY` does 1 and 2 currently. Maybe this should be an enum

nhynes: Actually, I'm just going to remove it. I'm not going to enable it, so it's just more code to maintain. For future readers, I'd do something like this

```rust
/// How much side-channel resistance do you want the runtime to have?
///
/// Mitigations can be applied by the application, but then the app authors
/// need to know about the issue and then write secure code (difficult).
/// Instead the runtime can apply mitigations, but this increases costs and
/// introduces incompatibilities for all apps including ones that don't need
/// side-channel resistance.
#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, PartialOrd, Ord)]
#[non_exhaustive]
enum Constancy {
    /// Operations may be data dependent. This is the default option.
    #[default]
    None,
    /// Data dependencies that are easy/cheap to mitigate are removed.
    ///
    /// Specifically: tx calldata zero/non-zero bytes cost the same, and
    /// binary opcodes do not short circuit.
    ///
    /// Caveats: nothing else is constified
    Light,
    /// Expensive mitigations are also applied.
    ///
    /// Specifically: storage set and reset gas become the same, reset gas is not
    /// refunded, and storage values are not removed when reset to zero.
    ///
    /// Caveats: Storage observers may still infer data based on access patterns.
    /// Short-circuiting operators within app code still need to be made constant-time.
    Heavy,
    /// Side channels are eliminated. Applications do not need to write constant-time code.
    ///
    /// Caveat: super expensive, even when you don't need confidentiality
    // TODO: oram, differentially private time+gas
    Full,
}
```

kostko (approved):

peterjgilbert (approved):

PR #1066, opened 2022-08-03
runtime-sdk/modules/contracts: Fix subcall dispatch

kostko:
Previously various gas and fee checks were performed even for internal
subcalls (e.g. when smart contracts called into other modules) which
was wrong and also caused failures due to minimum gas price not being
met (the fee is always set to zero for subcalls).

### DIFF
--- a/runtime-sdk/modules/contracts/src/test.rs
+++ b/runtime-sdk/modules/contracts/src/test.rs
@@ -484,6 +484,7 @@ impl Runtime for ContractRuntime {
             core::Genesis {
                 parameters: core::Parameters {
                     max_batch_gas: 10_000_000,
+                    min_gas_price: BTreeMap::from([(Denomination::NATIVE, 1)]),
                     ..Default::default()
                 },
             },
@@ -513,7 +514,9 @@ impl Runtime for ContractRuntime {
 }
 
 #[test]
-fn test_hello_contract_subcalls() {
+fn test_hello_contract_subcalls_overflow() {
+    use cbor::cbor_map;
+
     let mut mock = mock::Mock::default();
     let mut ctx = mock.create_ctx_for_runtime::<ContractRuntime>(context::Mode::ExecuteTx);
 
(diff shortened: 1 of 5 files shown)


runtime-sdk/modules/contracts/src/test.rs
```diff
@@ -513,7 +513,9 @@ impl Runtime for ContractRuntime {
 }
 
 #[test]
-fn test_hello_contract_subcalls() {
+fn test_hello_contract_subcalls_overflow() {
```
pro-wh, line 516: :information_source: apparently this was always a stack overflow test. this renames it to something more descriptive. in the same change, we're adding a new non-stack-overflow test with the old name

runtime-sdk/modules/contracts/src/test.rs
```diff
@@ -553,7 +559,68 @@ fn test_hello_contract_subcalls() {
...
+                amount: Default::default(),
+                gas: 2_000_000,
+                consensus_messages: 0,
+            },
+            ..Default::default()
+        },
+    };
+    ctx.with_tx(0, 0, tx, |mut tx_ctx, call| {
+        let result = Contracts::tx_call(&mut tx_ctx, cbor::from_value(call.body).unwrap())
+            .expect("call should succeed");
```
pro-wh, line 612: :information_source: same test as above except `once: true` and it succeeds

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -872,6 +872,15 @@ impl<Cfg: Config> module::TransactionHandler for Module<Cfg> {
             return Err(Error::GasOverflow);
         }
 
+        // Attempt to limit the maximum number of consensus messages.
+        let consensus_messages = ctx.tx_auth_info().fee.consensus_messages;
+        ctx.limit_max_messages(consensus_messages)?;
+
+        // Skip additional checks/gas payment for internally generated transactions.
+        if ctx.is_internal() && !ctx.is_simulation() {
```
pro-wh, line 880: if I were to do a simulation of a contract call that did a subcall, does this mean we'll additionally check the gas price on that subcall?

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -872,6 +872,15 @@ impl<Cfg: Config> module::TransactionHandler for Module<Cfg> {
...
+            return Ok(());
```
pro-wh, line 881: if things later get more complex, we should move subsets of the checking logic, e.g. the gas price checking and signature verification gas use, to helper functions and call them inside `if` blocks instead of early returning

pro-wh (commented):
need to ask about simulation

pro-wh (approved):
explicit internal flag

PR #1069, opened 2022-08-09
evm: expose public data in signed queries

nhynes:
Previously, the entire signed query was encoded into the `data` field and then encrypted, which makes it very difficult for tools and middleboxes to examine public data. This PR modifies the signed query format such that only the data is encrypted. Just the minimum of extensions are packed into the `data` field.

related: https://github.com/oasisprotocol/sapphire-paratime/pull/31


kostko (approved):
Looks cleaner to me.

PR #1078, opened 2022-08-22
Bump Rust toolchain to 2022-08-22

kostko:


runtime-sdk/modules/contracts/src/abi/oasis/test.rs
```diff
@@ -367,7 +367,7 @@ fn test_hello_contract_out_of_gas() {
     assert_eq!(result.code(), 12);
     assert_eq!(
         &result.to_string(),
-        "core: out of gas (limit: 1000 wanted: 1008)"
+        "core: out of gas (limit: 1000 wanted: 1013)"
```
pro-wh, line 370: what do you reckon does this? I do see an `.unwrap_or(!is_expensive)` converted from `unwrap_or_else`

kostko: But that is not in the Contract SDK so it should not have any effect. Could be some Rust std or compiler changes.

PR #1081, opened 2022-08-23
client-sdk/go: Add "cory" test key used by oasis-net-runner

matevz:
This PR makes `oasis` CLI more tailored to local (dApp) development and debugging:
- adds `cory` test account which is the default entity created by `oasis-net-runner` when using `--fixture.default.fund_entities` flag

### DIFF
--- a/client-sdk/go/testing/testing.go
+++ b/client-sdk/go/testing/testing.go
@@ -83,6 +83,8 @@ var (
 	Bob = newEd25519TestKey("oasis-runtime-sdk/test-keys: bob")
 	// Charlie is the test key C.
 	Charlie = newEd25519TestKey("oasis-runtime-sdk/test-keys: charlie")
+	// Cory is the default test key used by oasis-net-runner.
+	Cory = newEd25519TestKey("ekiden test entity key seed")
 	// Dave is the test key D.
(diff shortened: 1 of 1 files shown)


kostko (approved):

client-sdk/go/testing/testing.go
```diff
@@ -83,6 +83,8 @@ var (
 	Bob = newEd25519TestKey("oasis-runtime-sdk/test-keys: bob")
 	// Charlie is the test key C.
 	Charlie = newEd25519TestKey("oasis-runtime-sdk/test-keys: charlie")
+	// Cory is the default test key used by oasis-net-runner.
+	Cory = newEd25519TestKey("ekiden test entity key seed")
```
pro-wh, line 87: where is the name Cory from? It looks like we've mostly been using one name per first initial

kostko: It is the entity key from Oasis Core, hence Cory :-)

The other idea was Debbie, the debug entity :P

pro-wh: o lol

matevz: FYI Cory is deliberately separated from Alice, Bob, Charlie convention, because it's managed by the core tests/tooling (test-runner, net-runner). But it's still useful to have it around in Oasis CLI.

PR #1102, opened 2022-09-01
runtime-sdk/modules/evm: Add encryption precompiles

jberci:


runtime-sdk/modules/evm/src/precompile/test.rs
```diff
@@ -0,0 +1,28 @@
+use evm::Context;
+use oasis_runtime_sdk::{modules::accounts::Module, types::token::Denomination};
+pub use primitive_types::H160;
+
+use super::{get_precompiles, PrecompileResult};
+
+struct TestConfig {}
```
kostko, line 7:
```suggestion
struct TestConfig;
```

kostko (approved):

runtime-sdk/modules/evm/src/precompile/confidential.rs
```diff
@@ -0,0 +1,289 @@
...
+    Ok(result)
+}
+
+pub(super) fn call_deoxysii_open(
+    input: &[u8],
+    target_gas: Option<u64>,
+    _context: &Context,
+    _is_static: bool,
+) -> PrecompileResult {
+    let gas_cost = linear_cost(target_gas, input.len() as u64, 10_000, 1000)?;
```
nhynes, line 131: similarly, also open and seal can have the same cost (and hence be constants).

runtime-sdk/modules/evm/src/precompile/confidential.rs
```diff
@@ -0,0 +1,289 @@
...
+
+    // Input encoding: bytes32 key || bytes32 nonce || uint ciphertext_len || uint ad_len || ciphertext || ad.
+    let (key, nonce, ciphertext, ad) = decode_deoxysii_call_args(input)?;
+    let ciphertext = ciphertext.to_vec();
+
+    let deoxysii = DeoxysII::new(&key);
+    let result = match deoxysii.open(&nonce, ciphertext, ad) {
```
nhynes, line 138: you can just `return Ok(...)` here

runtime-sdk/modules/evm/src/precompile/confidential.rs
```diff
@@ -0,0 +1,289 @@
...
+        Ok(decrypted) => PrecompileOutput {
+            exit_status: ExitSucceed::Returned,
+            cost: gas_cost,
+            output: decrypted,
+            logs: Default::default(),
+        },
+        Err(_) => PrecompileOutput {
+            exit_status: ExitSucceed::Returned,
```
nhynes, line 146: This should cause a reversion. The data, in fact, could be empty, so that's not a reliable differentiator.

runtime-sdk/modules/evm/src/precompile/confidential.rs
```diff
@@ -0,0 +1,289 @@
...
+        });
+    }
+    public.copy_from_slice(&input[0..32]);
+    private.copy_from_slice(&input[32..64]);
+
+    let public = x25519_dalek::PublicKey::from(public);
+    let private = x25519_dalek::StaticSecret::from(private);
+
+    let mut kdf =
+        Hmac::<sha2::Sha512Trunc256>::new_from_slice(b"MRAE_Deoxys-II-256-128").map_err(|_| {
```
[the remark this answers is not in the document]
kostko: @jberci can we add some tests against known test vectors to ensure these are correct?

runtime-sdk/modules/evm/src/precompile/confidential.rs
```diff
@@ -0,0 +1,289 @@
...
+        .to_usize()
+        .ok_or_else(|| PrecompileFailure::Error {
+            exit_status: ExitError::Other("text length out of bounds".into()),
+        })?;
+    let text_size = text_len.saturating_add(31) & (!0x1f); // Round up to 32 bytes.
+
+    ad_len.copy_from_slice(&input[96..128]);
+    let ad_len_big = BigUint::from_bytes_be(&ad_len);
+    let ad_len = ad_len_big
+        .to_usize()
```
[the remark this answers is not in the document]
jberci: I suppose that could work, but then the code would need to manually check that the upper 24 bytes of the field are zeroes, unless we decide to ignore them. There is precedent for the BigUint approach in the bigmodexp implementation.

runtime-sdk/modules/evm/src/precompile/confidential.rs
```diff
@@ -0,0 +1,289 @@
...
+    Ok((key, nonce, text, ad))
+}
+
+pub(super) fn call_deoxysii_seal(
+    input: &[u8],
+    target_gas: Option<u64>,
+    _context: &Context,
+    _is_static: bool,
+) -> PrecompileResult {
+    let gas_cost = linear_cost(target_gas, input.len() as u64, 10_000, 1000)?;
```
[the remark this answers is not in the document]
jberci: I added some benchmarks and changed the costs a bit. They're based on #1015 (where e.g. a byte costs 6 gas for confidential storage operations), but still comparatively high to the standard precompiles.

nhynes (approved):

kostko (approved):

PR #1122, opened 2022-09-15
runtime-sdk: Add support for consensus staking

kostko:
Based on #1118

### DIFF
--- a/runtime-sdk/src/modules/accounts/mod.rs
+++ b/runtime-sdk/src/modules/accounts/mod.rs
@@ -439,7 +439,7 @@ impl API for Module {
         to: Address,
         amount: &token::BaseUnits,
     ) -> Result<(), Error> {
-        if ctx.is_check_only() {
+        if ctx.is_check_only() || amount.amount() == 0 {
             return Ok(());
         }
 
@@ -459,6 +459,10 @@ impl API for Module {
     }
 
     fn mint<C: Context>(ctx: &mut C, to: Address, amount: &token::BaseUnits) -> Result<(), Error> {
+        if ctx.is_check_only() || amount.amount() == 0 {
+            return Ok(());
+        }
+
         // Add to destination account.
         Self::add_amount(ctx.runtime_state(), to, amount)?;
 
@@ -479,6 +483,10 @@ impl API for Module {
         from: Address,
         amount: &token::BaseUnits,
     ) -> Result<(), Error> {
+        if ctx.is_check_only() || amount.amount() == 0 {
+            return Ok(());
+        }
+
         // Remove from target account.
         Self::sub_amount(ctx.runtime_state(), from, amount)?;
 
--- a/runtime-sdk/src/modules/consensus/mod.rs
+++ b/runtime-sdk/src/modules/consensus/mod.rs
@@ -1,23 +1,29 @@
 //! Consensus module.
 //!
 //! Low level consensus module for communicating with the consensus layer.
-use std::str::FromStr;
+use std::{convert::TryInto, num::NonZeroUsize, str::FromStr, sync::Mutex};
 
+use once_cell::sync::Lazy;
 use thiserror::Error;
 
 use oasis_core_runtime::{
     common::versioned::Versioned,
     consensus::{
+        beacon::EpochTime,
         roothash::{Message, StakingMessage},
         staking,
-        staking::Account as ConsensusAccount,
-        state::{staking::ImmutableState as StakingImmutableState, StateError},
+        staking::{Account as ConsensusAccount, Delegation as ConsensusDelegation},
+        state::{
+            beacon::ImmutableState as BeaconImmutableState,
+            staking::ImmutableState as StakingImmutableState, StateError,
+        },
+        HEIGHT_LATEST,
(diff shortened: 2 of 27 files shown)


runtime-sdk/src/modules/accounts/mod.rs
```diff
@@ -459,6 +459,10 @@ impl API for Module {
     }
 
     fn mint<C: Context>(ctx: &mut C, to: Address, amount: &token::BaseUnits) -> Result<(), Error> {
+        if amount.amount() == 0 {
```
ptrus, line 462: Any reason why not adding the same check to `transfer`? Also I noticed that only `transfer` has the `is_check_only` check.

runtime-sdk/src/modules/consensus/mod.rs
```diff
@@ -316,6 +330,45 @@ impl API for Module {
...
+            Lazy::new(|| Mutex::new(lru::LruCache::new(NonZeroUsize::new(128).unwrap())));
+
+        // Check the cache first to avoid more expensive traversals.
+        let mut cache = HEIGHT_CACHE.lock().unwrap();
+        if let Some(height) = cache.get(&epoch) {
+            return Ok(*height);
+        }
+
+        // Resolve height for the given epoch.
+        let mut height = HEIGHT_LATEST;
```
ptrus, line 345: Minor, probably irrelevant? improvement: could start at the height of the smallest cached epoch that is higher than `epoch`.

ptrus (commented):
self TODO: check the main part of the PR, changes in the consensus_accounts module.

runtime-sdk/src/modules/consensus_accounts/mod.rs
```diff
@@ -419,6 +724,82 @@ impl<Accounts: modules::accounts::API, Consensus: modules::consensus::API>
...
+                from: ud.from,
+                to: ud.to,
+                shares: udi.shares,
+                amount,
+            });
+        }
+    }
 }
 
 impl<Accounts: modules::accounts::API, Consensus: modules::consensus::API> module::InvariantHandler
```
ptrus, line 805: Could add an invariant that checks the number of shares the runtime has ecrowed in consensus. Should be `>=`  what is it in its internally tracked delegations, for each consensus account.

runtime-sdk/src/modules/consensus_accounts/mod.rs
```diff
@@ -419,6 +724,82 @@ impl<Accounts: modules::accounts::API, Consensus: modules::consensus::API>
...
+                    let height = Consensus::height_for_epoch(ctx, ud.epoch)
+                        .expect("failed to determine height for epoch");
+
+                    // Find the relevant reclaim escrow event.
+                    let totals = ctx
+                        .history()
+                        .consensus_events_at(height, EventKind::Staking)
+                        .expect("failed to fetch historic events")
+                        .iter()
+                        .find_map(|ev| match ev {
```
ptrus, line 760: Maybe add a comment here that there will always be exactly one matching reclaim escrow event here, because debonding delegations get merged in consensus (if there are multiple reclaims for the same accounts on the same epoch).

runtime-sdk/src/modules/consensus_accounts/state.rs
```diff
@@ -0,0 +1,278 @@
...
+    from: Address,
+    to: Address,
+    epoch: EpochTime,
+    shares: u128,
+) -> Result<(), Error> {
+    let store = storage::PrefixStore::new(ctx.runtime_state(), &MODULE_NAME);
+    let undelegations = storage::PrefixStore::new(store, &UNDELEGATIONS);
+    let account = storage::PrefixStore::new(undelegations, &to);
+    let mut entry = storage::TypedStore::new(storage::PrefixStore::new(account, &from));
+    let mut di: types::DelegationInfo = entry.get(epoch.to_be_bytes()).unwrap_or_default(); // XXX: use to_storage_key
```
ptrus, line 124: There's this note to use `to_storage_key` in few places.

runtime-sdk/src/modules/consensus_accounts/state.rs
```diff
@@ -0,0 +1,278 @@
...
+
+        Ok(Self {
+            epoch: EpochTime::from_be_bytes(value[..8].try_into()?),
+            to: Address::from_bytes(&value[8..8 + Address::SIZE])?,
+            from: Address::from_bytes(&value[8 + Address::SIZE..])?,
+        })
+    }
+}
+
+/// Retrieved all queued undelegations for epochs earlier than or equal to the passed epoch.
```
ptrus, line 192:
```suggestion
/// Retrieve all queued undelegations for epochs earlier than or equal to the passed epoch.
```

ptrus (approved):

runtime-sdk/src/modules/consensus_accounts/mod.rs
kostko: Added the invariant check.

ptrus: :+1: looks good

PR #1129, opened 2022-09-19
client-sdk/go: add evm NewCreateTx

pro-wh:
match up with NewCallTx

context: we're making a simulation of the rust raw-eth-format tx decoding in go, so there's the desire to construct a evm.Create tx to represent the decoded tx without intending to send it over a runtime client


kostko (approved):

pro-wh:
consistent with NewCallTx, there is no test. required tests pass. thanks for reviewing. gonna rebase and merge

PR #1138, opened 2022-09-27
evm: don't encode contract create output

nhynes:
This PR changes the confidential EVM create function to return the address, as would a non-confidential EVM. This will cause the gateway and downstream tools like the block explorer to display the correct address in the tx receipt.

There is no security risk to exposing the contract address, as it can be easily computed by observing that a transaction deposited new code (meaning a contract has been created) and then computing the address using the public sender address and nonce.

This PR should go in ASAP, as it's very annoying for devs who aren't able to use the block explorer as effectively.

### DIFF
--- a/runtime-sdk/modules/evm/src/lib.rs
+++ b/runtime-sdk/modules/evm/src/lib.rs
@@ -366,11 +366,14 @@ impl<Cfg: Config> API for Module<Cfg> {
             return Ok(vec![]);
         }
 
-        let (init_code, tx_metadata) =
+        // Create output (the contract address) does not need to be encrypted because it's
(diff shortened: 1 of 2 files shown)


kostko:
Didn't we say this is going to be fixed in the gateway?

kostko (approved):

nhynes:
> Didn't we say this is going to be fixed in the gateway?

No? Maybe that was in a private discussion to which I was not privy. This one gets a runtime fix because the gateway is really just reading out what it gets from the runtime. There's no reason for this output to be encrypted, so we'd might as well fix it here so that other gateway implementations don't also trip up.

PR #1152, opened 2022-10-06
Expose x25519 key derivation and deoxysii to wasm contracts

jberci:


runtime-sdk/modules/contracts/src/abi/oasis/crypto.rs
```diff
@@ -121,6 +121,136 @@ impl<Cfg: Config> OasisV1<Cfg> {
...
+                                .as_slice(&memory)
+                                .map_err(|_| wasm3::Trap::Abort)?;
+                            let message = Region::from_arg(message)
+                                .as_slice(&memory)
+                                .map_err(|_| wasm3::Trap::Abort)?;
+                            let additional_data = Region::from_arg(additional_data)
+                                .as_slice(&memory)
+                                .map_err(|_| wasm3::Trap::Abort)?;
+                            func(key, nonce, message, additional_data).map_err(|e| {
+                                let err = match e {
```
kostko, line 218: This should only abort in case the input is obviously malformed (e.g. unexpected sizes), but not if there is a problem with authenticating/decrypting the data. In case of authentication errors, the error should be propagated so the caller can handle it.

runtime-sdk/modules/contracts/src/lib.rs
```diff
@@ -141,6 +141,26 @@ pub enum Error {
...
+    #[sdk_error(code = 28)]
+    CryptoMalformedNonce,
+
+    #[error("crypto: key derivation function failure")]
+    #[sdk_error(code = 29)]
+    CryptoKeyDerivationFunctionFailure,
+
+    #[error("crypto: decryption failure")]
+    #[sdk_error(code = 30)]
+    CryptoDecryptionFailure,
```
kostko, line 162: Is this one still needed?

runtime-sdk/modules/contracts/src/lib.rs
```diff
@@ -217,6 +240,9 @@ impl Default for GasCosts {
             wasm_crypto_signature_verify_ed25519: 500_000,
             wasm_crypto_signature_verify_secp256k1: 500_000,
             wasm_crypto_signature_verify_sr25519: 500_000,
+            wasm_crypto_x25519_derive_symmetric: 100_000,
+            wasm_crypto_deoxysii_base: 50_000,
+            wasm_crypto_deoxysii_byte: 3,
```
kostko, line 245: Are these based on updated benchmarks?

kostko (approved):

PR #1161, opened 2022-10-12
Further WASM module validation features

jberci:


runtime-sdk/modules/contracts/src/abi/oasis/validation.rs
```diff
@@ -0,0 +1,513 @@
...
+use walrus::{
+    ir::{self, dfs_in_order, Value, Visitor},
+    Module, ValType,
+};
+
+use crate::{abi::oasis::Info, Error};
+
+const EXPORT_SUB_VERSION_PREFIX: &str = "__oasis_sv_";
+const MAX_FUNCTION_DECLARATIONS: usize = 10_000;
+const MAX_LOCAL_DECLARATIONS: usize = 256_000;
```
kostko, line 12: These should be `contracts` module parameters instead of constants.

kostko (approved):

PR #1164, opened 2022-10-12
ts-web: mysterious ts-jest re-unification

pro-wh:
unexplained in #1119 why it got split into specific packages' node_modules. bringing it back


pro-wh:
tests pass, I'll sneak this into #1139

pro-wh:
or no, #1146 would be better

PR #1174, opened 2022-10-20
ts-web/core: improve TransactionWrapper type

lukaw3d:


client-sdk/ts-web/core/src/types.ts
```diff
@@ -354,7 +354,7 @@ export interface ConsensusStatus {
 /**
  * Transaction is an unsigned consensus transaction.
  */
-export interface ConsensusTransaction {
+export interface ConsensusTransaction<BODY = unknown> {
```
pro-wh, line 357: these types are generated from Go. I'd prefer that we not add this here. are you using these ConsensusTransaction s for things not covered in TransactionWrapper and visitTransaction?

client-sdk/ts-web/core/test/types.test.ts
```diff
@@ -9,16 +9,20 @@ describe('types', () => {
...
+        it('Should detect incorrect fields', async () => {
             const tw = transferWrapper()
             tw.setBody({
                 amount,
                 to,
                 // @ts-expect-error Expect typescript to detect incorrect field
                 otherField: 5,
             })
+            // @ts-expect-error Expect typescript to detect incorrect field
+            expect(tw.transaction.body?.otherField).toEqual(5)
```
pro-wh, line 25: could this have bone in the previous commit?

lukaw3d: In the previous commit
```ts
            expect(tw.transaction.body?.amount).toEqual(amount)
            expect(tw.transaction.body?.to).toEqual(to)
            expect(tw.transaction.body?.otherField).toEqual(5)
```
all trigger typescript errors

client-sdk/ts-web/core/src/types.ts
lukaw3d: Oh, alright

Web Wallet tests use transactionWrapper.transaction.body https://github.com/oasisprotocol/oasis-wallet-web/blob/0e7cab63e57216b9001dbd091a69b58f4ce73c1b/src/app/lib/transaction.test.ts#L36

pro-wh: ok we could add a typed body getter to TransactionWrapper

pro-wh: although: is this done only in tests? if the test really is meant not to take it for granted that e.g.

```ts
expect(tw.transaction.body).toHaveProperty('shares')
```

then wouldn't it be appropriate to have a typescript assertion only after that `expect`? :raised_eyebrow:

lukaw3d: Yep, only in tests. I suspect `toHaveProperty` check is only there because types were missing

I don't think adding a getter that will only be used by wallet tests is worth it - we can easily keep those tests as they are, or refactor to `expect(tw.transaction.body).toEqual({ shares: new Uint8Array([4, 226]) }) // 1250`

PR #1176, opened 2022-10-21
Add precompiles for key generation and signing

jberci:


runtime-sdk/src/crypto/signature/secp256k1.rs
```diff
@@ -99,3 +99,70 @@ impl cbor::Decode for PublicKey {
...
+
+    /// Return a byte representation of this signer's key.
+    pub fn to_bytes(&self) -> Vec<u8> {
+        self.sk.to_bytes().to_vec()
+    }
+
+    /// Public key corresponding to the signer.
+    pub fn public(&self) -> PublicKey {
+        let ep: k256::EncodedPoint = (&self.sk.verifying_key()).into();
+        PublicKey(ep)
```
nhynes, line 139:
```suggestion
        PublicKey(k256::EncodedPoint::from(&self.sk.verifying_key()))
```

runtime-sdk/src/crypto/signature/secp256k1.rs
```diff
@@ -99,3 +99,70 @@ impl cbor::Decode for PublicKey {
...
+    }
+
+    /// Generates a signature with the private key over the context and message.
+    pub fn context_sign(&self, context: &[u8], message: &[u8]) -> Result<Signature, Error> {
+        let mut digest = Sha512Trunc256::new();
```
nhynes, line 144: Let's try to make these signatures Eth-compatible. Eth uses `keccak256` as the hash function.

nhynes (dismissed):
Thanks for this PR. I think the functionality overall looks solid, but it seems quite like there's some abstraction that describes all three variants in a way that ties them all together and reduces the need for tests.

One approach is to just have the one enum and its `impl` but have big ol' `match` arms that handle each case. This keeps things easy to understand and high performance with a slight decrease in readability.

Another approach create a trait called `trait MemorySigner` (or something) containing provided implementations of all of these methods with associated types `type Signer: Signer` and `type Digest: Digest` (so like `let mut digest = Self::Digest::new()`). You could then pass around a `Box<dyn MemorySigner>` for polymorphism. This comes with cleaner separation of implementations, but less efficiency (not that it really matters when you're only calling the thing once).

A third approach that mixes the two above and is quite like what you're already doing here: enum dispatch for performance but a trait to keep the APIs in sync. There's even a macro for it ([enum_dispatch](https://crates.io/crates/enum_dispatch))! I'm not sure if this one works with associated types, though.

I'd suggest:
1. take a look at the different kinds of digests and signers you'll need (e.g., `k256`+`sha3`, `ed25519_dalek`+`sha2`) and see if they all implement a common trait (e.g., [digest::Digest](https://docs.rs/digest/0.10.5/digest/trait.Digest.html)). If they do, you can simplify the implementation using [associated types](https://doc.rust-lang.org/rust-by-example/generics/assoc_items/types.html).
2. Try to collect everything into a trait and use `Box<dyn Signer>` for polymorphism to keep things simple. If you find that doesn't work, then you can bring back your enum but also bind the implementations together using a trait.
3. Look at `enum_dispatch` for your own edification. A tiny performance benefit isn't worth the supply chain risk of an extra dependency.

runtime-sdk/modules/evm/src/precompile/confidential.rs
```diff
@@ -165,14 +210,234 @@ pub(super) fn call_deoxysii_open(
...
+            exit_status: ExitError::Other("method identifier out of bounds".into()),
+        })?
+        .try_into()
+        .map_err(|_| PrecompileFailure::Error {
+            exit_status: ExitError::Other("unknown signature type".into()),
+        })?;
+
+    let signer = signature::MemorySigner::from_bytes(sig_type, pk).map_err(|err| {
+        PrecompileFailure::Error {
+            exit_status: ExitError::Other(format!("error creating signer: {}", err).into()),
```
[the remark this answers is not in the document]
jberci: `ExitError::Other` takes a `Cow`. The docs say that converting a String into Cow doesn't do any copying tho, so I don't think it should be a problem?

runtime-sdk/src/crypto/signature/ed25519.rs
```diff
@@ -94,3 +95,76 @@ impl From<PublicKey> for CorePublicKey {
...
+        let mut digest = Sha512Trunc256::new();
+        for byte in &[context, message] {
+            digest.update(byte);
+        }
+        let message = digest.finalize();
+
+        let pk = ed25519_dalek::PublicKey::from(&self.sk);
+        let signature = self.sk.sign(&message, &pk);
+
+        Ok(signature.to_bytes().to_vec().into())
```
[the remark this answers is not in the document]
jberci: I think returning something other than `Signature` would break the interface a bit too much, since the other implementations also return a Signature. I could extend Signature's interface to make it easier to avoid copying?

runtime-sdk/src/crypto/signature/secp256k1.rs
jberci: I used `Sha512Trunc256` here because it was already used earlier up in the verification code. I can change both uses to keccak256, but I'm not sure if that would break anything else.

kostko: No this part should definitely not be changed. Note that this is the "context sign" function which is used with our domain-separated signatures. Raw signatures should be supported using a different method.

runtime-sdk/modules/evm/src/precompile/confidential.rs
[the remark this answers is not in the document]
nhynes: Oh, sorry. I missed that. This is fine, then.

runtime-sdk/src/crypto/signature/secp256k1.rs
nhynes: In that case, let's keep this as-is (after unifying the implementations, of course). And then make sure this is very specifically called out in the [library](https://github.com/oasisprotocol/sapphire-paratime/pull/80/files#diff-da309fcbe4ff8ed13416002a10d3221207a12cb9e55cfd8563a9c7eae27d20faR211) and docs that this is for the Oasis context since most people will not want this, but rather Eth-compatible signatures. But, for the record, I think that oasis-core using secp256k1+sha512/256 is just going to confuse people.

We can try our best to avoid confusion by calling `signMessageWithContext` `oasisContextSign` in the wrapper lib, and saying something like

> signPrehashed - signs a pre-hashed message using the provided key
>
> oasisContextSign - signs a message that intended for verification by the Oasis consensus layer or core modules. For Ethereum-compatible signatures, use `signPrehashed` on a `keccak256` digest.

But now I wonder why we're adding this precompile instead of providing a `sha512/256` digestion precompile and another for signing prehashed messages. The wrapper library can take care of any of the pitfalls of giving prehashed signing mechanisms directly to developers.

kostko: The oasis-specific one probably shouldn't be exposed as a precompile, this is used for signing transactions.

runtime-sdk/modules/evm/src/precompile/utils.rs
```diff
@@ -0,0 +1,62 @@
+//! Utilities useful in implementing precompiles.
+use std::cmp::min;
+
+use evm::{executor::stack::PrecompileFailure, ExitError};
+use num::{BigUint, ToPrimitive as _};
+
+/// Length of an EVM word, in bytes.
+pub const WORD: usize = 32;
+
+pub fn round_up(what: usize) -> usize {
```
nhynes, line 10:
```suggestion
pub fn round_up_nyble(what: usize) -> usize {
```

runtime-sdk/modules/evm/src/precompile/confidential.rs
```diff
@@ -165,14 +207,243 @@ pub(super) fn call_deoxysii_open(
...
+    }
+
+    let (method, input) =
+        utils::cut_uint_from_params(input, "method identifier out of bounds".into())?;
+    let (pk_len, input) =
+        utils::cut_uint_from_params(input, "private key length out of bounds".into())?;
+    let (context_len, input) =
+        utils::cut_uint_from_params(input, "context length out of bounds".into())?;
+    let (message_len, input) =
+        utils::cut_uint_from_params(input, "message length out of bounds".into())?;
```
nhynes, line 302: I thought we were going to sign prehashed except in the case of sr25519?

nhynes (commented):
This looks overall good. Just one main question, though: I see that the precompile is still about signing messages rather than signing digests using separate digest functions. Am I mistaken in thinking this? @kostko

runtime-sdk/modules/evm/src/precompile/utils.rs
jberci: Depends on what you mean by the u5. The function should round the number up to the nearest multiple of 32, so the bottom 5 bits are cleared.

runtime-sdk/modules/evm/src/precompile/confidential.rs
jberci: The way these crates are made, there's no way to just give them the bytes of an existing hash. I tried going the dummy hasher route, but that proved impossible too, so the current code has multiple combinations of signer+hash and always hashes the message itself.

runtime-sdk/src/crypto/signature/mod.rs
```diff
@@ -1,11 +1,81 @@
...
+    #[cbor(rename = "ed25519_oasis")]
+    Ed25519_Oasis,
+    #[cbor(rename = "ed25519_pure")]
+    Ed25519_Pure,
+    #[cbor(rename = "ed25519_sha512")]
+    Ed25519_Sha512,
+    #[cbor(rename = "secp256k1_oasis")]
+    Secp256k1_Oasis,
+    #[cbor(rename = "secp256k1_pure")]
+    Secp256k1_Pure,
```
kostko, line 27: Would it be better if this was called `_Sha256` instead of "pure"?

runtime-sdk/src/crypto/signature/mod.rs
```diff
@@ -66,6 +159,78 @@ impl PublicKey {
...
+                    if hash.len() != <Sha512 as sha2::digest::FixedOutput>::OutputSize::USIZE {
+                        return Err(Error::InvalidArgument);
+                    }
+                    let digest = digests::DummyDigest::<Sha512>::new_precomputed(hash);
+                    pk.verify_digest(digest, signature)
+                }
+                _ => Err(Error::InvalidArgument),
+            },
+            Self::Secp256k1(pk) => match signature_type {
+                SignatureType::Secp256k1_Keccak256 => {
```
kostko, line 220: What about prehashed with `Secp256k1_Sha256`? That should also work?

runtime-sdk/src/crypto/signature/mod.rs
```diff
@@ -66,6 +159,78 @@ impl PublicKey {
...
+                    if hash.len() != <sha3_0_9_1::Keccak256 as sha3_0_9_1::digest::FixedOutput>::OutputSize::USIZE {
+                        return Err(Error::InvalidArgument);
+                    }
+                    // Use SHA-256 for RFC6979 even if Keccak256 was used for the message.
+                    let digest = digests::DummyDigest::<sha2::Sha256>::new_precomputed(hash);
```
kostko, line 225: So in this case we don't actually care about what hash function was used to compute the hash as it seems we always use SHA-256 for inner hashing (and Keccak256 is only used for the length check)? Does this mean we could just use a generic `Secp256k1_PreHash` as long as the length matches (or what exactly is the requirement on the input)?

runtime-sdk/modules/evm/src/precompile/confidential.rs
```diff
@@ -1,21 +1,71 @@
...
+
+use ethabi::{ParamType, Token};
 use evm::{
     executor::stack::{PrecompileFailure, PrecompileOutput},
     Context, ExitError, ExitRevert, ExitSucceed,
 };
 use hmac::{Hmac, Mac, NewMac as _};
-use oasis_runtime_sdk::core::common::crypto::mrae::deoxysii::{DeoxysII, KEY_SIZE, NONCE_SIZE};
+
+use once_cell::sync::Lazy;
```
kostko, line 10: Group together with above.

runtime-sdk/modules/evm/src/precompile/confidential.rs
```diff
@@ -1,21 +1,71 @@
...
+    costs.insert(SignatureType::Ed25519_Pure, 35_000);
+    costs.insert(SignatureType::Ed25519_PrehashedSha512, 35_000);
+    costs.insert(SignatureType::Secp256k1_Oasis, 110_000);
+    costs.insert(SignatureType::Secp256k1_PrehashedKeccak256, 110_000);
+    costs.insert(SignatureType::Secp256k1_PrehashedSha256, 110_000);
+    Mutex::new(costs)
+});
+
+/// The costs of a message signing operation.
+static SIGN_MESSAGE_PREHASHED_COST: Lazy<Mutex<HashMap<SignatureType, (u64, u64)>>> =
```
kostko, line 45:
```suggestion
static SIGN_MESSAGE_COST: Lazy<Mutex<HashMap<SignatureType, (u64, u64)>>> =
```

runtime-sdk/modules/evm/src/precompile/confidential.rs
```diff
@@ -1,21 +1,71 @@
...
+        costs.insert(SignatureType::Ed25519_Pure, (75_000, 8));
+        costs.insert(SignatureType::Ed25519_PrehashedSha512, (75_000, 0));
+        costs.insert(SignatureType::Secp256k1_Oasis, (150_000, 8));
+        costs.insert(SignatureType::Secp256k1_PrehashedKeccak256, (150_000, 0));
+        costs.insert(SignatureType::Secp256k1_PrehashedSha256, (150_000, 0));
+        Mutex::new(costs)
+    });
+
+/// The costs of a signature verification operation.
+static VERIFY_MESSAGE_PREHASHED_COST: Lazy<Mutex<HashMap<SignatureType, (u64, u64)>>> =
```
kostko, line 58:
```suggestion
static VERIFY_MESSAGE_COST: Lazy<Mutex<HashMap<SignatureType, (u64, u64)>>> =
```

runtime-sdk/modules/evm/src/precompile/confidential.rs
```diff
@@ -1,21 +1,71 @@
...
+
 /// The base setup cost for encryption and decryption.
 const DEOXYSII_BASE_COST: u64 = 50_000;
 /// The cost for encryption and decryption per word of input.
 const DEOXYSII_WORD_COST: u64 = 100;
-/// Length of an EVM word, in bytes.
-const WORD: usize = 32;
+
+/// The cost of a key pair generation operation, per method.
+static KEYPAIR_GENERATE_BASE_COST: Lazy<Mutex<HashMap<SignatureType, u64>>> = Lazy::new(|| {
```
kostko, line 33: Why do these need a mutex, will they ever change?

runtime-sdk/modules/evm/src/precompile/confidential.rs
```diff
@@ -1,21 +1,71 @@
...
+    let mut costs = HashMap::with_capacity(6);
```
kostko, line 34: You could use `HashMap::from([(SignatureType::Ed25519_Oasis, 35_000), ( ... ), ...])` to avoid some code.

runtime-sdk/src/crypto/signature/digests.rs
```diff
@@ -0,0 +1,98 @@
+use k256::ecdsa::digest as ecdsa_digest;
+
+pub(crate) struct DummyDigest<D> {
```
kostko, line 3: Add a doc comment describing what this is and why it is needed.

kostko (approved):

runtime-sdk/src/crypto/signature/ed25519.rs
```diff
@@ -94,3 +115,63 @@ impl From<PublicKey> for CorePublicKey {
...
+        let pk = ed25519_dalek::PublicKey::from(&self.sk);
+        self.sk
+            .sign_prehashed(digest, &pk, None)
+            .map_err(|_| Error::SigningError)
+            .map(|sig| sig.to_bytes().to_vec().into())
+    }
+}
+
+impl super::Signer for MemorySigner {
+    fn new_from_seed(seed: &[u8]) -> Result<Self, Error> {
```
nhynes, line 138: this isn't really so much of `generate` as `derivePublicKey`. Maybe this should invoke `random_bytes` with `seed` as the pers str?

nhynes:
guess I was too late

PR #1188, opened 2022-11-05
ts-web/core: add debonding_amount to StakingTakeEscrowEvent

mitjat:
Updates the types with the field introduced in https://github.com/oasisprotocol/oasis-core/pull/5016

The type change is backwards-compatible, so it doesn't hurt much to have the new type in ts-web even before https://github.com/oasisprotocol/oasis-core/pull/5016 is widely deployed; users will just see `debonding_amount=0` when decoding an event from an old node.

### DIFF
--- a/client-sdk/ts-web/core/src/types.ts
+++ b/client-sdk/ts-web/core/src/types.ts
@@ -2965,7 +2965,14 @@ export interface StakingStakeThreshold {
  */
 export interface StakingTakeEscrowEvent {
     owner: Uint8Array;
+    /**
+     * amount is the sum of amounts slashed from active and debonding escrow balances.
+     */
     amount: Uint8Array;
+    /**
+     * debonding_amount is the amount slashed from the debonding escrow balance.
+     */
+    debonding_amount: Uint8Array;
 }
 
 /**


mitjat:
LMK if I should bump the version in `client-sdk/ts-web/core/package.json`

client-sdk/ts-web/core/src/types.ts
```diff
@@ -2965,7 +2965,14 @@ export interface StakingStakeThreshold {
  */
 export interface StakingTakeEscrowEvent {
     owner: Uint8Array;
+    /**
+     * amount is the sum of amounts slashed from active and debonding escrow balances.
```
pro-wh, line 2969: you should add these in oasis-core. our reflect-go program will automatically transfer them here

pro-wh (commented):
I guess this is okay, because it's backwards compatible. but otherwise our sdk is currently tracking oasis-core 22.1.x. switching to 22.2 is tracked in #1118 . let's get @kostko 's input on this before merging, in case it somehow conflicts with that work

don't bump the package.json version. we'll do that when we make another release. does anything want this right away?

you can add a fragment in the docs/changelog.md under Unreleased changes if you want.

mitjat:
Nobody needs the updated client right away. Since the only change introduced here will be autogenerated by `reflect-go` later as per @pro-wh, I'm abandoning this PR. Thanks for explaining.

PR #1200, opened 2022-11-09
docker: Add missing cmake dependency

kostko:


pro-wh (commented):
what's it needed for?

:+1: implemented as described

kostko:
Reproducible builds to not fail after Oasis Core 22.2.x (due to mbedtls-sys build process).

pro-wh:
thanks, someone had asked about mbedtls-sys on discord

PR #1210, opened 2022-11-18
ts-web/core: update types + client

pro-wh:
for 22.2.x


pro-wh:
:facepalm: there are two SGXQuotePolicy now

kostko (approved):

pro-wh:
adding changelog

PR #1216, opened 2022-12-02
evm: Implement call_format precompile

matevz:
TODO

### DIFF
--- a/runtime-sdk/modules/evm/src/lib.rs
+++ b/runtime-sdk/modules/evm/src/lib.rs
@@ -16,6 +16,7 @@ use evm::{
     Config as EVMConfig,
 };
 use once_cell::sync::OnceCell;
+use std::cell::RefCell;
 use thiserror::Error;
 
 use oasis_runtime_sdk::{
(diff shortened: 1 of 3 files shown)


nhynes:
What kind of app needs to know the callformat? It seems like something the contract either shouldn't care about, and if it does, it should let the user specify the desired behavior rather than inferring it from something that happens outside of the contract's universe.

matevz:
> What kind of app needs to know the callformat? It seems like something the contract either shouldn't care about, and if it does, it should let the user specify the desired behavior rather than inferring it from something that happens outside of the contract's universe.

Two reasons actually:
1. consistency with [Cipher's context object](https://github.com/oasisprotocol/oasis-sdk/blob/main/contract-sdk/src/context.rs) 
2. since encrypted transactions are not enforced, a way for dApp devs to make sure no private data is leaked by (accidentally?) unencrypted calls and transactions.

nhynes:
> consistency with Cipher's context object

We don't need consistency with an experimental ParaTime.

> since encrypted transactions are not enforced, a way for dApp devs to make sure no private data is leaked by (accidentally?) unencrypted calls and transactions.

I call using the encrypting cipher and then publish the results online. You can't stop people from compromising themselves. You have to go out of your way to use the plain cipher, so it's clearly intentional. The contract should be programmed not to release secrets to people who are going to compromise themselves.

matevz:
Superseeded by https://github.com/oasisprotocol/emerald-web3-gateway/pull/340

PR #1222, opened 2022-12-09
Comment where protobufjs-cli is used

lukaw3d:


pro-wh (approved):
yeah that's the place

PR #1240, opened 2023-01-25
[runtime-sdk] add txhash to Event

Andrew7234:
Adds the related tx hash, if it exists, to runtime events returned by the runtime client.

Open questions: Do we need to support `evm` and `contracts` modules? Do those events have a related transaction? If yes, how do we want to update the `evm.Event` and `contracts.Event` without breaking the api? (Currently they are cbor-unmarshalled directly so updating the `Event` is not possible. Ideally we would wrap the event in a larger struct similar to the other modules, but this seems like a breaking change)

Edit**: We will leave the module level `Event`s unchanged.

### DIFF
--- a/client-sdk/go/client/client.go
+++ b/client-sdk/go/client/client.go
@@ -337,7 +337,7 @@ func (rc *runtimeClient) GetTransactionsWithResults(ctx context.Context, round u
 
 		for _, rawEv := range raw.Events {
 			var ev types.Event
-			if err := ev.UnmarshalRaw(rawEv.Key, rawEv.Value); err != nil {
+			if err := ev.UnmarshalRaw(rawEv.Key, rawEv.Value, nil); err != nil {
 				continue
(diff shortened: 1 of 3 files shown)


pro-wh:
> wrap the event in a larger struct

that sounds good. these events also show up in TransactionWithResults.Events, where it's already evident what transaction it's from.

kostko (commented):
Yeah, so I think adding `TxHash` to `types.Event` is fine, but adding it to all of the specialized event types seems like it would be duplicating the transaction hash in places (e.g. as @pro-wh mentioned above) so I would rather not do that.

Andrew7234:
Thanks for the tips. I added TxHash to `types.Event`, which I believe is enough. I can then fetch all events with `GetEventsRaw` to get the events+txHash and then decode the events.

kostko (approved):

PR #1244, opened 2023-01-27
runtime-sdk: Allow custom modules to expose call result error

aefhm:
#### Goal
https://app.clickup.com/t/3c16qd5
https://app.clickup.com/t/863g4tpv3

#### Details
Allow modules to customize the disclosure of failures in the pursuit of improved development environment.

_Note_. This is a backwards compatible change and the default is still encryption.

#### TODO
- [x] Add `expose_failure` flag
- [x] Figure out merge/release steps. Presuming a bump on this dependency.

### DIFF
--- a/client-sdk/go/callformat/callformat.go
+++ b/client-sdk/go/callformat/callformat.go
@@ -2,7 +2,6 @@ package callformat
 
 import (
 	"crypto/rand"
-	"encoding/base64"
 	"fmt"
 
 	"github.com/oasisprotocol/deoxysii"
@@ -82,20 +81,19 @@ func DecodeResult(result *types.CallResult, meta interface{}) (*types.CallResult
 		// In case of plain-text data format, we simply pass on the result unchanged.
 		return result, nil
 	case *metaEncryptedX25519DeoxysII:
-		// Make sure the result makes sense in this context.
+		var envelope types.ResultEnvelopeX25519DeoxysII
 		switch {
 		case result.IsUnknown():
+			if err := cbor.Unmarshal(result.Unknown, &envelope); err != nil {
+				return nil, fmt.Errorf("callformat: malformed result envelope: %w", err)
+			}
 		case result.IsSuccess():
-			// Unexpected as a successful result shouldn't be plain.
-			return nil, fmt.Errorf("callformat: unexpected result: %s", base64.StdEncoding.EncodeToString(result.Ok))
+			if err := cbor.Unmarshal(result.Ok, &envelope); err != nil {
+				return nil, fmt.Errorf("callformat: malformed result envelope: %w", err)
+			}
 		default:
 			// Submission could fail before call format processing so the result would be plain.
 			return nil, result.Failed
-		}
-
-		var envelope types.ResultEnvelopeX25519DeoxysII
-		if err := cbor.Unmarshal(result.Unknown, &envelope); err != nil {
-			return nil, fmt.Errorf("callformat: malformed result envelope: %w", err)
 		}
 
 		// Open sealed envelope.
--- a/runtime-sdk/src/callformat.rs
+++ b/runtime-sdk/src/callformat.rs
@@ -197,39 +197,69 @@ pub fn encode_result<C: Context>(
     result: module::CallResult,
     metadata: Metadata,
 ) -> CallResult {
+    encode_result_ex(ctx, result, metadata, false /* expose_failure */)
+}
+
+/// Encode call results.
+///
+/// If `expose_failure` is set, then this method will not encrypt errors.
+pub fn encode_result_ex<C: Context>(
+    ctx: &C,
+    result: module::CallResult,
+    metadata: Metadata,
+    expose_failure: bool,
+) -> CallResult {
     match metadata {
         // In case of the plain-text data format, we simply pass on the data unchanged.
         Metadata::Empty => result.into(),
 
         // Encrypted data format using X25519 key exchange and Deoxys-II symmetric encryption.
         Metadata::EncryptedX25519DeoxysII { pk, sk, index } => {
-            // Generate nonce for the output as Round (8 bytes) || Index (4 bytes) || 00 00 00.
-            let mut nonce = Vec::with_capacity(deoxysii::NONCE_SIZE);
-            nonce
-                .write_u64::<BigEndian>(ctx.runtime_header().round)
-                .unwrap();
-            nonce
-                .write_u32::<BigEndian>(index.try_into().unwrap())
-                .unwrap();
-            nonce.extend(&[0, 0, 0]);
-            if ctx.is_simulation() {
-                // Randomize the lower-order bytes of the nonce to facilitate private queries.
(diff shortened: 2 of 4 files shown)


kostko (commented):
Hm, what is this about?

aefhm:
> Hm, what is this about?

Exposing/hoisting the [error status and messages](https://app.clickup.com/t/3c16qd5)

kostko:
Task linked: [CU-3c16qd5 expose error status and messages to devs and explorer (return real success/error status. not just Unknown)](https://app.clickup.com/t/3c16qd5)

aefhm:
I think technically this provides test coverage of the `CallResult::Failure` through a helper, but do you think we need another specific test? cc @nhynes 
https://github.com/oasisprotocol/oasis-sdk/blob/5e0353622f049fb53c52b3e7c8e54aa2d998bbdc/client-sdk/go/callformat/callformat.go#L93
https://github.com/oasisprotocol/oasis-sdk/blob/5e0353622f049fb53c52b3e7c8e54aa2d998bbdc/tests/e2e/evmtest.go#L927-L935

kostko:
Hm, so failures are never encrypted now? What if they contain sensitive data?

aefhm:
> Hm, so failures are never encrypted now?

Yes.

> What if they contain sensitive data?

DApp developers/contract writers would need to structure the smart contracts to prevent leaking sensitive data in an error/failure case.

kostko:
Hm I think this needs some more discussion as it is not like one can always control failures.

One option is also to reveal success/failure but encrypt any details?

What exactly is the goal here, making debugging easier?

aefhm:
> Hm I think this needs some more discussion as it is not like one can always control failures.

👍 Makes sense. I'll condense this discussion thread and corral some opinions.
> One option is also to reveal success/failure but encrypt any details?

Hmm. This plan would suggest maybe revealing the `module` and `code` but encrypting the `message` of a `Failure`?

> What exactly is the goal here, making debugging easier?

Yes. I think having the raw error message would make debugging/dev easier, and add more context or details to an explorer/BlockScout view, e.g. [`Error: (Awaiting internal transactions for reason)`](https://testnet.explorer.sapphire.oasis.dev/tx/0x06b2e5e89b1416e72e87e740adf56f23fe4a7c873294b40cf35f94b687d8140d)

kostko:
> and add more context or details to an explorer/BlockScout view

But one solution for that is that the caller would be able to login with their wallet and decrypt the responses shown in the block explorer? The client should be always able to decrypt the response to their own request, it's just that others can't do that.

I don't like just not encrypting responses, it seems strange to me, it's like TLS would only be used for sending requests, but responses would be plain text.

runtime-sdk/src/callformat.rs
```diff
@@ -196,42 +196,67 @@ pub fn encode_result<C: Context>(
...
+                CallResult::Ok(encrypt_result(ctx, result, pk, sk, index))
+            } else {
+                result
+            }
         }
     }
 }
 
+// Encrypt a result as cbor value.
+pub fn encrypt_result<C: Context>(
```
kostko, line 232: This seems specific to the X25519-Deoxys-II format, so maybe call it `encrypt_result_x25519_deoxysii`. Also mention the encryption scheme in the documentation comment.

runtime-sdk/src/callformat.rs
```diff
@@ -196,42 +196,67 @@ pub fn encode_result<C: Context>(
...
+            if result.is_success() {
+                CallResult::Ok(encrypt_result(ctx, result, pk, sk, index))
+            } else {
+                result
+            }
         }
     }
 }
 
+// Encrypt a result as cbor value.
```
kostko, line 231:
```suggestion
/// Encrypt a call result using the X25519-Deoxys-II encryption scheme.
```

runtime-sdk/src/callformat.rs
```diff
@@ -196,42 +196,67 @@ pub fn encode_result<C: Context>(
     ctx: &C,
     result: module::CallResult,
     metadata: Metadata,
+) -> CallResult {
+    encode_result_ex(ctx, result, metadata, false /* show_raw_errors */)
+}
+
+pub fn encode_result_ex<C: Context>(
```
kostko, line 203: Add documentation comment.

client-sdk/go/callformat/callformat.go
```diff
@@ -82,20 +81,23 @@ func DecodeResult(result *types.CallResult, meta interface{}) (*types.CallResult
...
+		// Submission could fail before call format processing so the result would be plain.
+		if result.Failed != nil {
 			return nil, result.Failed
 		}
 
 		var envelope types.ResultEnvelopeX25519DeoxysII
-		if err := cbor.Unmarshal(result.Unknown, &envelope); err != nil {
-			return nil, fmt.Errorf("callformat: malformed result envelope: %w", err)
+
+		if result.IsUnknown() {
```
kostko, line 91: Looks like these two would make more sense in a switch as only one of them should be true. Actually the failed one could also be handled here?

runtime-sdk/src/callformat.rs
```diff
@@ -196,42 +196,67 @@ pub fn encode_result<C: Context>(
...
+    ctx: &C,
+    result: module::CallResult,
+    metadata: Metadata,
+    show_raw_errors: bool,
```
nhynes, line 207:
```suggestion
    expose_failure: bool,
```
I recommend this because the encrypted errors aren't "cooked" but rather repacked/hidden.

runtime-sdk/src/callformat.rs
```diff
@@ -196,42 +196,67 @@ pub fn encode_result<C: Context>(
...
 ) -> CallResult {
     match metadata {
```
nhynes, line 209: if you don't want to use a separate function called `encrypt_whatever`, you can do something like this

```rust
if expose_failure && !result.is_success() {
    return result;
}
```
and then you're free of having to check elsewhere

nhynes (commented):
This LGTM to me except for a few comments from both Jernej and me. Otherwise, ready to merge!

runtime-sdk/src/callformat.rs
aefhm: The explicit variable usage is nice.

aefhm: Hmm, I think I like having the more explicit flow surrounding `expose_failure`.

runtime-sdk/src/callformat.rs
aefhm: Thanks. Taken.

client-sdk/go/callformat/callformat.go
aefhm: Ah, fair enough. `switch` it is.

kostko (commented):
This looks good to me now, thanks! Will let @nhynes also do another pass.

nhynes (approved):

kostko (approved):

PR #1256, opened 2023-02-04
evm: allow unencrypted calls

nhynes:
This PR changes confidential EVM to accept all calls instead of just enveloped ones. This makes it easier to accidentally send public transactions, but it makes it harder for third-party tooling to integrate. We could make the gateway more restrictive instead of the runtime. The gateway would see a mistakenly unencrypted tx anyway.


kostko (approved):
Looks good. What about queries?

PR #1268, opened 2023-02-20
evm: add curve25519 public key generation precompile

nhynes:
This PR adds a curve25519 public key generation precompile. This was added because it's not totally cool to use ed25519 keys for DH. Goes with https://github.com/oasisprotocol/sapphire-paratime/pull/116.


kostko (approved):
Looks good, but I would call it x25519 for consistency with other

PR #1299, opened 2023-03-16
Bump Go to 1.20.2

kostko:


client-sdk/go/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/oasisprotocol/oasis-sdk/client-sdk/go
 
-go 1.18
+go 1.19
```
pro-wh, line 3: some of these down here only upgrade to 1.19. do they need a different version? although I see oasis-core does the same with go.mod files saying 1.19 while using 1.20

pro-wh (approved):
if it works, yeah.

approving with https://github.com/oasisprotocol/oasis-core/pull/5228

client-sdk/go/go.mod
kostko: Yeah we support both 1.19 and 1.20.

PR #1301, opened 2023-03-19
runtime-sdk: Allow confidential gas estimation with zeroized caller

kostko:


runtime-sdk/src/modules/core/mod.rs
```diff
@@ -538,7 +558,6 @@ impl<Cfg: Config> Module<Cfg> {
         let bs_max_iters = Self::estimate_gas_search_max_iters(ctx);
 
         // Update the address used within the transaction when caller address is passed.
-        let mut extra_gas = 0;
         if let Some(caller) = args.caller.clone() {
             let address_spec = transaction::AddressSpec::Internal(caller);
             match args.tx.auth_info.signer_info.first_mut() {
```
pro-wh, line 563: we don't entirely wipe the signer info, we only replace the first one

kostko: Good point, although the (currently deployed) runtimes don't currently allow any additional signers we should still just get rid of those in case the caller is overriden.

kostko: Fixed.

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -484,11 +493,22 @@ impl<Cfg: Config> Module<Cfg> {
...
         ctx: &mut C,
         mut args: types::EstimateGasQuery,
     ) -> Result<u64, Error> {
+        let mut extra_gas = Self::get_estimate_gas_extra(ctx);
+        // In case the runtime is confidential we are unable to authenticate the caller so we must
+        // make sure to zeroize it to avoid leaking private information.
+        if ctx.is_confidential() {
+            args.caller = Some(
+                args.caller
+                    .unwrap_or(CallerAddress::Address(Default::default()))
```
pro-wh, line 507: :information_source: for sapphire we'll have to set `caller: Some(CallerAddress::EthAddr(some_dummy_value))` from the client

kostko: This is exactly what the `.zeroized()` call does ;-)

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -538,19 +558,17 @@ impl<Cfg: Config> Module<Cfg> {
...
-                }
-            }
+            args.tx.auth_info.signer_info = vec![transaction::SignerInfo {
+                address_spec: transaction::AddressSpec::Internal(caller),
+                nonce: args
+                    .tx
+                    .auth_info
+                    .signer_info
+                    .first()
+                    .map(|si| si.nonce)
```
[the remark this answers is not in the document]
kostko: We did this before, but it was not so explicit (e.g. in case the signer info was there, we only updated the address spec but left the nonce intact).

runtime-sdk/src/modules/core/mod.rs
kostko: Fixed, please take a look.

pro-wh: lg

runtime-sdk/src/modules/core/mod.rs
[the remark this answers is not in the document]
pro-wh: o right

PR #1309, opened 2023-03-23
runtime-sdk/modules/evm: Always propagate errors in plaintext

kostko:


pro-wh (commented):
this is probably okay, but @aefhm could you check?

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -703,14 +702,8 @@ impl<Cfg: Config> Module<Cfg> {
             // encrypt the call in the normal way.
             return evm_result;
         }
-        let call_result = match evm_result {
-            Ok(exit_value) => module::CallResult::Ok(exit_value.into()),
-            Err(e) => module::CallResult::Failed {
```
aefhm, line 708: In the not effectively (backwards compatible) case of `expose_failure=false`, we are technically still encrypting the error?
https://github.com/oasisprotocol/oasis-sdk/blob/de4cec5b08066e877311a8756a5b6fcec7ebe655/runtime-sdk/src/callformat.rs#L231

aefhm (approved):

runtime-sdk/modules/evm/src/lib.rs
kostko: Yes, but in this case it would never come to that (`expose_failure` is true here and the call result would now never contain an error). Or are you concerned about other uses?

pro-wh: should we maybe put an `unreachable!` in callformat then?

pro-wh (approved):
:+1: here

runtime-sdk/modules/evm/src/lib.rs
aefhm: > Yes, but in this case it would never come to that (expose_failure is true here and the call result would now never contain an error).

Agree.

> Or are you concerned about other uses?

I loosely recall that we were concerned about supporting previous versions of paratimes using this EVM module. I am comfortable not doing that and just moving ahead.

With that being said, then I support the above comment.

> should we maybe put an unreachable! in callformat then?

kostko: The unreachable path would be in case `encode_result_ex` was called with `expose_failure` set to `true` and one passing a failed `result`? It is true that this combination is not currently used anywhere (while the one that results in an Unknown is still used for non-EVM calls), I don't think we should just abort in that case?

aefhm: > unreachable path

My b. I forgot that non-EVM calls were still using the `Unknown` flow. That's what I was thinking of at first.
https://github.com/oasisprotocol/oasis-sdk/blob/de4cec5b08066e877311a8756a5b6fcec7ebe655/runtime-sdk/src/callformat.rs#L231

> The unreachable path would be in case `encode_result_ex` was called with `expose_failure `set to `true` and one passing a failed `result`?

Agree this is not used.

> I don't think we should just abort in that case?

Sounds good to me. I agree that this function (`encode_result_ex`) should be more complete.

PR #1321, opened 2023-04-11
runtime-sdk: Add support for estimating extra gas on failures

kostko:


pro-wh (approved):
ok

contract authors are advised not to allow the success/fail status of a contract call with zeroized sender to leak info :shrug: because node operator may set this "extra gas fail" arbitrarily high to make it very clear which one happened

kostko:
> contract authors are advised not to allow the success/fail status of a contract call with zeroized sender to leak info

A call failing was never secret for the caller anyway, it is just that the caller couldn't previously try it for the zeroized address case. Do you see an example where this could be a concern? Basically it would need to be something secret-dependent and also gated on a zeroized caller?

pro-wh:
> A call failing was never secret for the caller anyway

oh then never mind haha

PR #1338, opened 2023-04-20
runtime-sdk: Make extra gas estimation per method

kostko:
Testing has shown that having this enabled globally leads to bad estimates even for methods where regular estimation should work just fine. This makes the extra gas estimations per-method.


pro-wh (approved):
ok ya. thanks for the testing

PR #1352, opened 2023-04-27
[BREAKING] client-sdk: Fix ETH address format for test accounts

matevz:
Fix ETH address handling for non-secp256k1 test accounts.

Fixes https://github.com/oasisprotocol/cli/issues/63

### DIFF
--- a/client-sdk/go/helpers/address.go
+++ b/client-sdk/go/helpers/address.go
@@ -58,7 +58,7 @@ func ResolveAddress(net *config.Network, address string) (*types.Address, *ethCo
 		case addressExplicitTest:
 			// Test key.
 			if testKey, ok := testing.TestAccounts[data]; ok {
-				return &testKey.Address, &testKey.EthAddress, nil
+				return &testKey.Address, testKey.EthAddress, nil
 			}
 			return nil, nil, fmt.Errorf("unsupported test account: %s", data)
 		default:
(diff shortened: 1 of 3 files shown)


client-sdk/go/helpers/address.go
```diff
@@ -58,7 +58,11 @@ func ResolveAddress(net *config.Network, address string) (*types.Address, *ethCo
 		case addressExplicitTest:
 			// Test key.
 			if testKey, ok := testing.TestAccounts[data]; ok {
-				return &testKey.Address, &testKey.EthAddress, nil
```
pro-wh, line 61: what's in testKey.EthAddress if it's not secp256k1eth?

pro-wh:
what's the codepath that leads to this in the cli though?

client-sdk/go/helpers/address.go
pro-wh: should we just make the test key struct have a *ethCommon.Address instead then?

pro-wh (approved):
thanks

matevz:
Let's try to clean this and use `*ethCommon.Address` instead of zeros (actually leftovers from File wallet object before the cli/sdk split).

pro-wh (approved):
oh thanks

kostko (approved):

PR #1354, opened 2023-05-01
client/ts-web: add repo links to packages

lukaw3d:
Apparently https://github.com/oasisprotocol/oasis-sdk/pull/1099 was not enough for NPM to link back to the repo. It probably wants repo field too

### DIFF
--- a/client-sdk/ts-web/package-lock.json
+++ b/client-sdk/ts-web/package-lock.json
@@ -2200,9 +2200,9 @@
             }
         },
         "node_modules/async": {
-            "version": "3.2.0",
-            "resolved": "https://registry.npmjs.org/async/-/async-3.2.0.tgz",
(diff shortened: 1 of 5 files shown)


client-sdk/ts-web/package-lock.json
```diff
@@ -4,6 +4,7 @@
     "requires": true,
     "packages": {
         "": {
+            "name": "ts-web",
```
pro-wh, line 7: does it really add this? let me see if I can reproduce this locally

pro-wh: reproduced locally, with npm 9.6.5. weird, `npm install` neither adds nor removes this. I guess let's have it here since `npm audit fix` really did put it there. feel free to remove whenever. it's not really a named package. it's just a workspace for holding workspaces.

pro-wh (approved):
fields good, version bumps semver-seems-safe.

although I wonder if it has anything about us not having these fields in the `latest` tag on npm. but nothing against having these fields and security bumps even if it doesn't fix that.

client-sdk/ts-web/package-lock.json
lukaw3d: yeah :D even `npm ci` works in both cases

I'll remove it

PR #1246, opened 2023-01-28
ts-web/core: Run playground e2e in nodejs

lukaw3d:
In case we want to ensure ts-web/core works in nodejs too (with xhr2 polyfill)

PR #1364, opened 2023-05-11
ts-web/core: Run playground e2e in Node.js 18

lukaw3d:
Supersedes and closes https://github.com/oasisprotocol/oasis-sdk/pull/1246


kostko (approved):

client-sdk/ts-web/core/playground/src/startPlayground.mjs
```diff
@@ -1,10 +1,8 @@
 // @ts-check
 
-import * as oasis from './../..';
-
-const nic = new oasis.client.NodeInternal('http://localhost:42280');
-
-export const playground = (async function () {
+/** @param {import('./../..')} oasis */
+export async function startPlayground(oasis) {
```
pro-wh, line 4: we were interested in fixing the import so that we don't need to turn this into a function

client-sdk/ts-web/core/playground/src/startPlayground.mjs
```diff
@@ -2,7 +2,7 @@
 
 /** @param {import('./../..')} oasis */
 export async function startPlayground(oasis) {
-    const nic = new oasis.client.NodeInternal('http://localhost:42280');
+    const nic = new oasis.client.NodeInternal('http://127.0.0.1:42280');
```
pro-wh, line 5: :information_source: we had been listening on specifically ipv4 127.0.0.1 in the envoy proxy config, this fixes it not to try ipv6 localhost. thanks for the fix

client-sdk/ts-web/core/playground/src/startPlayground.mjs
pro-wh: oh or at least removing the parameter. dunno if you prefer it to be a function.

client-sdk/ts-web/core/playground/src/startPlayground.mjs
lukaw3d: I should make that as a separate commit

client-sdk/ts-web/core/playground/src/startPlayground.mjs
lukaw3d: Parameter: it was needed because nodejs and cypress having incompatible import syntaxes. But since this is an `npm workspace`, replacing "./../.." with "@oasisprotocol/client" should work and be compatible.

I'll remove the parameter, but I do prefer it as a function

client-sdk/ts-web/core/playground/src/startPlayground.mjs
lukaw3d: https://github.com/oasisprotocol/oasis-sdk/pull/1364/commits/a86df5d7fedf51ab36736a9deed11b3d83001303

client-sdk/ts-web/core/playground/src/startPlayground.mjs
pro-wh: thanks. yeah, we avoid side effects everywhere else too, so it's consistent to move it into a function

PR #1367, opened 2023-05-15
client-sdk/go/types: Use non-wrapped public keys as return values

kostko:


pro-wh (commented):
what was the purpose of this PublicKey struct? does anything else use it?

pro-wh (approved):
oh ok, feel free to convert other uses over to unwrapped too then

kostko:
> oh ok, feel free to convert other uses over to unwrapped too then

I didn't find any other uses where this would be relevant though. Let me know if there is something you had in mind and we can fix it in a followup PR.

Issue #1371, opened 2023-05-22
crypto libraries review


pro-wh:
map of crypto libraries

![image](https://github.com/oasisprotocol/oasis-sdk/assets/38229038/a2e770de-fecf-451f-8d77-e7d1e4707d52)

PR #1372, opened 2023-05-24
ts-web: trying out other crypto libraries

pro-wh:
cross reference #1371

### DIFF
--- a/client-sdk/ts-web/rt/src/signature_secp256k1.ts
+++ b/client-sdk/ts-web/rt/src/signature_secp256k1.ts
@@ -1,5 +1,5 @@
+import {secp256k1} from '@noble/curves/secp256k1';
 import * as oasis from '@oasisprotocol/client';
-import * as elliptic from 'elliptic';
 
 export interface Signer {
     public(): Uint8Array;
@@ -11,8 +11,6 @@ export interface ContextSigner {
     sign(context: string, message: Uint8Array): Promise<Uint8Array>;
 }
 
-const SECP256K1 = new elliptic.ec('secp256k1');
-
 export async function verify(
     context: string,
     message: Uint8Array,
@@ -20,9 +18,7 @@ export async function verify(
     publicKey: Uint8Array,
 ) {
     const signerMessage = await oasis.signature.prepareSignerMessage(context, message);
-    const publicKeyA = Array.from(publicKey);
-    // @ts-expect-error acceptance of array-like encoded public key is not modeled
-    return SECP256K1.verify(signerMessage, signature, publicKeyA);
+    return secp256k1.verify(signature, signerMessage, publicKey);
 }
 
 export class BlindContextSigner implements ContextSigner {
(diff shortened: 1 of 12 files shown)


CedarMist:
I second this pull request, the `@noble/` ECC related libraries are much cleaner than `elliptic`.

client-sdk/ts-web/rt/src/signature_secp256k1.ts
```diff
@@ -44,28 +38,28 @@ export class BlindContextSigner implements ContextSigner {
     }
 }
 
-export class EllipticSigner implements Signer {
-    key: elliptic.ec.KeyPair;
+export class NobleSigner implements Signer {
+    key: Uint8Array;
 
-    constructor(key: elliptic.ec.KeyPair, note: string) {
+    constructor(key: Uint8Array, note:string) {
```
pro-wh, line 44: wait why is this space gone

pro-wh:
rebased, bumped versions

pro-wh:
gonna see about using `@noble/curves` for ed25519 as well

pro-wh:
opening this up for review, without any tweetnacl removal. I looked and I ended up getting all paranoid about ADR-0009

major change: EllipticSigner from client-rt is replaced with NobleSigner, with a similar interface. it's not tested against anything external though, we only check that the sdk can verify its own signature

pro-wh:
ok good, secp256k1 NobleSigner matches up with the runtime-sdk rust code

```ts
import * as oasis from '@oasisprotocol/client';
import * as oasisRT from '@oasisprotocol/client-rt';

describe('secp256k1', () => {
    it('is compatible with runtime-sdk', async () => {
        // Based on `test_memory_signer` in
        // runtime-sdk/src/crypto/signature/mod.rs.
        const priv = await oasis.hash.hash(oasis.misc.fromString('memory signer test'));
        expect(oasis.misc.toHex(priv)).toEqual('4844bcd03d683656f9b7b656f5a53883f7aad7a74d8781fd953d77b5e59956de');
        const signer = oasisRT.signatureSecp256k1.NobleSigner.fromPrivate(priv, 'this key is not important');
        const contextSigner = new oasisRT.signatureSecp256k1.BlindContextSigner(signer);
        const pk = contextSigner.public();
        expect(oasis.misc.toHex(pk)).toEqual('0305327494815250eb0850beb36e78c512a1a42d37503f18843d5eb55554742117');
        const ctx = 'oasis-core/test: context';
        const message = oasis.misc.fromString('this is a message');
        const signature = await contextSigner.sign(ctx, message);
        expect(oasis.misc.toHex(signature)).toEqual('304502210080e4f0e7a94baea44dcded16fb543642291aad5d7ab77606f4fce4cfc5b119260220010c0b40a09803494d7339bb52fb044028c9a58b0964610d863a50274264bd18');
        expect(await oasisRT.signatureSecp256k1.verify(ctx, message, signature, pk)).toEqual(true);
    });
});
```

gotta sort out the global `crypto.getRandomValues` in jest

pro-wh:
adding this to the test :upside_down_face: 

```js
import {webcrypto} from 'crypto';

if (typeof crypto === 'undefined') {
    // @ts-expect-error there are some inconsequential type differences
    globalThis.crypto = webcrypto;
}
```

pro-wh:
> other projects using Node will have to do similar patches?

ts-web targets web only, it's based on grpc-web, which internally uses XMLHttpRequest

I think our last recommendation was to use `--experimental-global-webcrypto`. see more hacks here https://github.com/oasisprotocol/oasis-sdk/blob/main/client-sdk/ts-web/core/playground/e2e-test-nodejs.js

pro-wh:
web wallet builds successfully with this version and minor change:

```diff
diff --git a/src/app/lib/transaction.ts b/src/app/lib/transaction.ts
index ff834281..c46f2cc1 100644
--- a/src/app/lib/transaction.ts
+++ b/src/app/lib/transaction.ts
@@ -19,7 +19,7 @@ export const signerFromPrivateKey = (privateKey: Uint8Array) => {
 }
 
 export const signerFromEthPrivateKey = (ethPrivateKey: Uint8Array) => {
-  return oasisRT.signatureSecp256k1.EllipticSigner.fromPrivate(ethPrivateKey, 'this key is not important')
+  return oasisRT.signatureSecp256k1.NobleSigner.fromPrivate(ethPrivateKey, 'this key is not important')
 }
 
 /** Transaction Wrapper */
```

PR #1380, opened 2023-05-30
wip: emit Transfer events for fee payments/disbursments

ptrus:


ptrus:
https://app.codecov.io/gh/oasisprotocol/oasis-sdk/pull/1380/indirect-changes?src=pr&el=tree-more&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=oasisprotocol :sob:

runtime-sdk/src/modules/accounts/mod.rs
```diff
@@ -952,6 +952,32 @@ impl module::TransactionHandler for Module {
         Ok(())
     }
 
+    fn after_handle_call<C: TxContext>(ctx: &mut C) -> Result<(), modules::core::Error> {
```
kostko, line 955: Can this result in the move from tx fee accumulator to block fee accumulator not happening in case the transaction fails before call dispatch (but fees have been charged as they are charged in `authenticate_tx`)? For example, in case the (encrypted) call cannot be decoded or the `before_handle_call` validation fails.

We should also add a test for this case.

kostko: I guess the problem is also about emitting events as they will be emitted in batch context instead of tx? Maybe we could add another hook (or call `after_handle_call` with an additional result argument) which would be called from `dispatch_tx_opts` in the failed call handler?

ptrus: Oh right, I also wanted to emit the fee-events as transaction tags not as block tags. But I missed that `after_handle_call` is not always called.

So likely I would need a new hook that is called with transaction context, regardless of any errors during the tx processing, then I can `emit_unconditonally` events there.

ptrus: I didn't see your reply before posting my last two replies :grin: I think updating the `after_handle_call` should be best?

kostko: Yeah probably should be. Maybe with an additional argument signalling the status (success vs. failure).

ptrus: Updated, and also added a test (see 2nd commit).

runtime-sdk/src/module.rs
```diff
@@ -382,9 +382,8 @@ pub trait TransactionHandler {
     /// Perform any action after call, within the transaction context.
     ///
     /// If an error is returned the transaction call fails and updates are rolled back.
-    fn after_handle_call<C: TxContext>(_ctx: &mut C) -> Result<(), modules::core::Error> {
+    fn after_handle_call<C: TxContext>(_ctx: &mut C, _result: &CallResult) {
```
kostko, line 385: So this can no longer abort the call, any reason for removing the return value (I know it was previously not used, but it may be useful in certain cases to be able to do that)? If this is removed the documentation comment above should be updated.

kostko (commented):
Thanks, looks good now, just one minor comment about the `after_handle_call` API change.

kostko (approved):

PR #1408, opened 2023-07-10
evm: add sha512-256 precompile

ptrus:


runtime-sdk/modules/evm/src/precompile/mod.rs
```diff
@@ -99,14 +100,14 @@ impl<Cfg: Config, B: EVMBackendExt> PrecompileSet for Precompiles<'_, Cfg, B> {
     }
 
     fn is_precompile(&self, address: H160) -> bool {
-        // All Ethereum precompiles are zero except for the last byte, which is no more than five.
-        // Otherwise, when confidentiality is enabled, Oasis precompiles start with one and have a last byte of no more than four.
+        // All Ethereum precompiles are zero except for the last byte, which is no more than six.
```
ptrus, line 103: Is having this additional `is_precompile` fn really a worth optimization (vs just letting the above `match` handle this)?

runtime-sdk/modules/evm/src/precompile/standard.rs
```diff
@@ -293,6 +293,19 @@ fn calculate_modexp_gas_cost(
     Ok(gas)
 }
 
+pub(super) fn call_sha512_256(handle: &mut impl PrecompileHandle) -> PrecompileResult {
+    record_linear_cost(handle, handle.input().len() as u64, 60, 12)?; // TODO: sensible gas costs, not sure what to set.
```
ptrus, line 297: Any suggestions for gas costs? Current values were copied from sha256.

runtime-sdk/modules/evm/src/precompile/mod.rs
kostko: Yeah I think it would be better to have a common match function instead of this, if possible to do nicely.

ptrus: Unless I'm missing something, just removing the `is_precompile` function would already work like that, no?

runtime-sdk/modules/evm/src/precompile/mod.rs
```diff
@@ -86,6 +86,7 @@ impl<Cfg: Config, B: EVMBackendExt> PrecompileSet for Precompiles<'_, Cfg, B> {
             (0, 3) => standard::call_ripemd160(handle),
             (0, 4) => standard::call_datacopy(handle),
             (0, 5) => standard::call_bigmodexp(handle),
+            (0, 6) => standard::call_sha512_256(handle),
```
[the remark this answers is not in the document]
ptrus: :+1: Moved it to `0x01 ... 0x01 0x01` 

Added some documentation to indicate that the standard package implements the standard precompiles defined in EVM spec.

runtime-sdk/modules/evm/src/precompile/auxiliary.rs
```diff
@@ -0,0 +1,44 @@
...
+    })
+}
+
+#[cfg(test)]
+mod test {
+    use super::super::test::*;
+
+    #[test]
+    fn test_sha512_256() {
+        let input = "38d18acb67d25c8bb9942764b62f18e17054f66a817bd4295423adf9ed98873e000000000000000000000000000000000000000000000000000000000000001b38d18acb67d25c8bb9942764b62f18e17054f66a817bd4295423adf9ed98873e789d1dd423d25f0772d2748d60f7e4b81bb14d086eba8e8e8efb6dcff8a4ae02";
```
ptrus, line 30: Would probably be nice to have E2E test in an actual .sol contract for all/most precompiles?

runtime-sdk/modules/evm/src/precompile/auxiliary.rs
kostko, line 1: I would just move this into a `sha512` module instead of `auxiliary` (with #1412 we will also have a `subcall` module/precompile).

runtime-sdk/modules/evm/src/precompile/auxiliary.rs
```diff
@@ -0,0 +1,44 @@
...
+    executor::stack::{PrecompileHandle, PrecompileOutput},
+    ExitSucceed,
+};
+use ripemd160::Digest as _;
+use sha2::Sha512Trunc256;
+
+use super::{record_linear_cost, PrecompileResult};
+
+pub(super) fn call_sha512_256(handle: &mut impl PrecompileHandle) -> PrecompileResult {
+    record_linear_cost(handle, handle.input().len() as u64, 60, 12)?; // TODO: sensible gas costs, not sure what to set.
```
kostko, line 12: Probably should be benchmarked against SHA256 and set based on that?

ptrus: Ok, did some benchmarks. Seems like the base cost is a bit higher, and then it grows similarly to sha256. I updated the values and pushed the benchmarks. There's also a doc comment pointing to the benchmarks, for anyone revisiting this in future.

comp

runtime-sdk/modules/evm/src/precompile/sha512.rs
```diff
@@ -0,0 +1,46 @@
...
+    executor::stack::{PrecompileHandle, PrecompileOutput},
+    ExitSucceed,
+};
+use ripemd160::Digest as _;
+use sha2::Sha512Trunc256;
+
+use super::{record_linear_cost, PrecompileResult};
+
+pub(super) fn call_sha512_256(handle: &mut impl PrecompileHandle) -> PrecompileResult {
+    // Costs were computed by benchmarking and comparing to SHA256 and using the already set SHA256 costs (unsure how those were obtained).
```
kostko, line 12: > (unsure how those were obtained)

I believe those come from Ethereum.

kostko (approved):

runtime-sdk/modules/evm/src/precompile/sha512.rs
ptrus: Ah thanks, indeed. Updated the comment.

PR #1411, opened 2023-07-11
runtime-sdk: Decouple storage from Context, add thread-local store

kostko:


runtime-sdk/src/storage/current.rs
```diff
@@ -0,0 +1,514 @@
+//! A store attached to the current thread.
```
ptrus, line 1: This looks ok :+1: (i checked this a bit more in depth, didn't fully check all places that were updated that use storage, but seems a straightforward update in most places)

runtime-sdk/src/subcall.rs
```diff
@@ -0,0 +1,214 @@
...
+}
+
+/// Subcall validator.
+pub trait Validator {
+    /// Validate a subcall before it is performed.
+    fn validate(&self, info: &SubcallInfo) -> Result<(), Error>;
+}
+
+/// A validator which allows everything.
+pub struct AllowAllValidator;
```
ptrus, line 26: I guess the `subscall` dispatch will be used in other places in a future PR as well. Currently, it's only used in a single place, so the abstraction seems a bit arbitrary (e.g. the `Validator`  trait which is only implemented by the `AllowAllValidator` ).

ptrus (approved):

runtime-sdk/src/subcall.rs
kostko: Yeah, see #1412 (WIP).

runtime-sdk/src/storage/mkvs.rs
```diff
@@ -55,6 +54,11 @@ impl<M: mkvs::MKVS> NestedStore for MKVSStore<M> {
         self.parent
     }
 
+    fn rollback(self) -> Self::Inner {
```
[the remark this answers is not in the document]
kostko: Fixed.

runtime-sdk/src/storage/current.rs
```diff
@@ -0,0 +1,514 @@
...
+            },
+            transactions: vec![],
+        };
+
+        CURRENT.with(|c| {
+            c.try_borrow_mut()
+                .expect("must not re-enter from with block")
+                .push(current)
+        });
+        let _guard = CurrentStoreGuard; // Ensure current store is popped once we return.
```
peternose, line 75: Could add the same comment to the `CurrentStoreGuard`.

runtime-sdk/src/storage/current.rs
```diff
@@ -0,0 +1,514 @@
...
+
+impl Drop for TransactionGuard {
+    fn drop(&mut self) {
+        let level = CURRENT.with(|c| {
+            let mut current_ref = c.borrow_mut();
+            let current = current_ref.last_mut().expect("must enter context");
+            current.transactions.len()
+        });
+
+        // If transaction hasn't been either committed or reverted, rollback.
```
peternose, line 30: Could add the same comment to the `TransactionGuard`.

Issue #1402, opened 2023-07-03
EVM: add ECADD, ECMUL & ECPAIRING opcodes for Ethereum ZK compatibility


CedarMist:
To be able to use Ethereum compatible zkSNARK circuits, and other cool tech, it would be useful for the EVM to include the ECADD, ECMUL and ECPAIRING opcodes.

To be more specific, implement [EIP-196](https://eips.ethereum.org/EIPS/eip-196) and [EIP-197](https://eips.ethereum.org/EIPS/eip-197).

There are several rust implementations of these opcodes and the underlying pairing & curve operations which could be used.

Curve implementations:

 * https://github.com/paritytech/bn
 * https://github.com/zcash-hackworks/bn
 * https://github.com/M4tsuri/Cryptix/tree/main
 * https://github.com/arkworks-rs/curves/tree/master/bn254

EVM opcode implementations:

 * https://github.com/ewasm/ethereum-bn128.rs/blob/master/src/lib.rs
 * https://github.com/AcalaNetwork/Acala/blob/34b1d74b46f160cbe86e99d6aae0b588b175b099/modules/evm/src/precompiles/bn128.rs#L63

PR #1416, opened 2023-07-25
runtime-sdk/evm/precomile: Add Ethereum alt-bn128 precompiles

peternose:
Related work:
https://github.com/paritytech/frontier/blob/master/frame/evm/precompile/bn128/src/lib.rs
https://eips.ethereum.org/EIPS/eip-1108
https://eips.ethereum.org/EIPS/eip-196
https://eips.ethereum.org/EIPS/eip-197

Also see this GitHub issue:
[oasis-sdk#1402](https://github.com/oasisprotocol/oasis-sdk/issues/1402)


runtime-sdk/modules/evm/Cargo.toml
```diff
@@ -18,8 +18,14 @@ thiserror = "1.0"
...
 ripemd160 = { version = "0.9", default-features = false }
-k256 = { version = "0.10.4", default-features = false, features = ["keccak256", "ecdsa"] }
+k256 = { version = "0.10.4", default-features = false, features = [
+    "keccak256",
+    "ecdsa",
+] }
 sha3 = { version = "0.10", default-features = false }
+substrate-bn = "0.6.0"
+serde = { version = "1.0.144", features = ["derive"] }
+serde_json = { version = "1.0.87", features = ["raw_value"] }
```
kostko, line 28: Are these only needed for tests? In that case they should be moved under `dev-dependencies` below.

runtime-sdk/modules/evm/src/precompile/standard.rs
kostko, line 1: Could we instead move these into their own `bn128` module as this one is getting quite long.

kostko (approved):

PR #1440, opened 2023-08-08
evm: add sha512 precompile

ptrus:


runtime-sdk/modules/evm/src/precompile/sha512.rs
```diff
@@ -23,6 +23,20 @@ pub(super) fn call_sha512_256(handle: &mut impl PrecompileHandle) -> PrecompileR
     })
 }
 
+pub(super) fn call_sha512(handle: &mut impl PrecompileHandle) -> PrecompileResult {
+    // Same costs as SHA512_256.
```
peternose, line 27: I guess this is ok, no need to update `bench_sha256_sha512_256`.

runtime-sdk/modules/evm/src/precompile/mod.rs
```diff
@@ -128,6 +128,7 @@ impl<Cfg: Config, B: EVMBackendExt> PrecompileSet for Precompiles<'_, Cfg, B> {
             // Oasis-specific, general.
             (1, 1, 1) => sha512::call_sha512_256(handle),
             (1, 1, 2) => subcall::call_subcall(handle, self.backend),
+            (1, 1, 3) => sha512::call_sha512(handle),
```
peternose, line 131: You could also move it one place up, as the subcalls haven't been deployed yet :)

peternose (approved):

runtime-sdk/modules/evm/src/precompile/mod.rs
ptrus: I'd need to recompile the subcall solidity example :sob:

peternose: No need to. Thought it was simple.

kostko (approved):

runtime-sdk/modules/evm/src/precompile/mod.rs
kostko: There is now a makefile so as long as you have `solc` installed it should be simple.

ptrus: Ok, updated.

PR #1459, opened 2023-08-22
runtime-sdk/modules/evm: Fix incorrect contract address computation

kostko:


runtime-sdk/modules/evm/src/backend.rs
```diff
@@ -471,8 +486,13 @@ impl<'ctx, 'backend, 'config, C: TxContext, Cfg: Config> StackState<'config>
...
     fn inc_nonce(&mut self, address: H160) -> Result<(), ExitError> {
-        // Do not increment the origin nonce as that has already been handled by the SDK.
-        if address == self.origin() {
+        let ctx = self.backend.ctx.borrow_mut();
+
+        // Do not increment the origin nonce as that has already been handled by the SDK. But do
+        // record that the nonce should be incremented based on EVM semantics so we can adjust any
+        // results from the `basic` method.
+        if address == self.origin() && !ctx.is_simulation() {
+            self.substate.origin_nonce_incremented = true;
```
peternose, line 495: Should this be a counter?

runtime-sdk/modules/evm/src/backend.rs
```diff
@@ -471,8 +486,13 @@ impl<'ctx, 'backend, 'config, C: TxContext, Cfg: Config> StackState<'config>
     }
 
     fn inc_nonce(&mut self, address: H160) -> Result<(), ExitError> {
-        // Do not increment the origin nonce as that has already been handled by the SDK.
-        if address == self.origin() {
+        let ctx = self.backend.ctx.borrow_mut();
+
+        // Do not increment the origin nonce as that has already been handled by the SDK. But do
```
peternose, line 491: Have no idea where this is handled. I only see one call to `Cfg::Accounts::inc_nonce` and that is few lines below..

runtime-sdk/modules/evm/src/backend.rs
kostko: No, as the nonce is only incremented once. This is basically because the EVM assumes it is the entire runtime handling also transaction logic. 

But in our case the SDK (e.g. the `accounts` module) is handling nonce increments.

runtime-sdk/modules/evm/src/backend.rs
kostko: You mean the actual nonce increment performed by the SDK? The accounts module is handling nonce checks and increments in the authentication handler.

peternose: I see. The module is using `checked_add`, that's why it is not calling `inc_nonce` (or `inc_nonces`, if implemented).

Not related to this PR, but `is_empty` calls `basic` twice.

peternose (approved):

PR #1460, opened 2023-08-22
runtime-sdk: Always rollback any changes to storage from queries

kostko:
Note that in practice this is a no-op as Oasis Core will perform the rollback in any case. This just makes it explicit.


peternose (approved):
> Note that in practice this is a no-op as Oasis Core will perform the rollback in any case. This just makes it explicit.

I would add this as a comment, but saying that this is probably a no-op (as we can change the core code in the future).

peternose (approved):

PR #1465, opened 2023-08-28
runtime-sdk: Do not emit gas used events for subcalls

kostko:


pro-wh (approved):
good simplification :tipping_hand_person:

PR #1466, opened 2023-08-29
runtime-sdk: Add support for undelegations query

kostko:


runtime-sdk/src/modules/consensus_accounts/types.rs
```diff
@@ -76,12 +84,23 @@ pub struct DelegationInfo {
 /// Extended information about a delegation.
 #[derive(Clone, Debug, Default, cbor::Encode, cbor::Decode)]
 pub struct ExtendedDelegationInfo {
```
peternose, line 86: Should/can we rename to `DelegationInfo`?

peternose (approved):

PR #1467, opened 2023-08-29
tools/gen_runtime_vectors: Add (un)delegate

matevz:


tools/gen_runtime_vectors/main.go
```diff
@@ -193,6 +193,80 @@ func main() {
...
+						origTo       *ethCommon.Address
+						chainContext string
+						valid        bool
+					}{
+						// Valid Delegate: Alice -> Bob's native address on ParaTime
+						{testing.Bob.Address.String(), nil, context.ChainContext, true},
+						// Valid Delegate: Alice -> Dave's native address on ParaTime
+						{testing.Dave.Address.String(), nil, context.ChainContext, true},
+						// Valid Delegate: Alice -> Dave's ethereum address on ParaTime
+						{testing.Dave.EthAddress.String(), testing.Dave.EthAddress, context.ChainContext, true},
```
kostko, line 209: Just note that delegating to Ethereum addresses makes absolutely no sense as those are not controllable in the consenus layer.

kostko (approved):

tools/gen_runtime_vectors/main.go
matevz: Is there a reason to keep this at all or should we just remove non-consensus addresses? If so, I would also remove orig_to from meta for delegate/undelegate?

matevz: I removed the non-consensus addresses now. Still not sure, if we may keep the support for signing those or they open potential vulnerabilities by supporting it.

kostko (approved):

PR #1473, opened 2023-09-01
client-sdk/go: Update validity rules for config identifiers

kostko:


client-sdk/go/config/identifier.go
```diff
@@ -14,6 +14,8 @@ func ValidateIdentifier(id string) error {
 		return fmt.Errorf("identifier cannot be empty")
 	case len(id) > 64:
 		return fmt.Errorf("identifier must be less than 64 characters long")
+	case id == "default":
+		return fmt.Errorf("identifier cannot be the string 'default'")
 	case !validID.MatchString(id):
 		return fmt.Errorf("identifier must only contain lower-case letters, numbers and _")
```
peternose, line 20: The error message could also be updated.

client-sdk/go/config/identifier.go
```diff
@@ -5,7 +5,7 @@ import (
 	"regexp"
 )
 
-var validID = regexp.MustCompile(`^[a-z0-9_]+$`)
+var validID = regexp.MustCompile(`^[a-z0-9][a-z0-9_]+$`)
```
peternose, line 8: Should we use * instead of +?

client-sdk/go/config/identifier.go
kostko: Fixed.

client-sdk/go/config/identifier.go
kostko: Yeah, fixed and added tests.

peternose (approved):

PR #1479, opened 2023-09-08
runtime-sdk/modules/evm: Simplify revert reason encoding

### LINKED ISSUE #1478: EVM: revert data is not handled uniformly
Via: https://github.com/oasisprotocol/oasis-web3-gateway/issues/429

When the EVM reverts it returns an arbitrary length array of bytes. When using Solidity this follows the convention of ABI encoding it with the signature `Error(string)`. However, other custom errors can be returned, or a contract may decide to return arbitrary non ABI-encoded data for some reason.

Custom error example in Solidity:
```solidity
    error Blah(uint);
    function testRevert () external view {
        revert Blah(123);
    }
```

The problem is the following code intercepts `Error(string)` encoded reverts and returns just the string. Otherwise it returns up to 1024 bytes of the revert data, as a base64 encoded string. This means it's not possible to know if the data returned is the raw revert data but base64 encoded, or a string extracted from the raw revert data.

The web3 gateway then tries to re-encode the error into the raw format returned from the EVM so it can be used by Ethereum clients which follow the ABI encoded error spec. This means, instead of getting a nice custom ABI encoded error type, you get Error(string) with a bunch of base64 encoded data - making custom errors difficult / annoying to use with Sapphire.

https://github.com/oasisprotocol/oasis-sdk/blob/957380713924e01aed725c020579f7b3cd049786/runtime-sdk/modules/evm/src/lib.rs#L222-L256

While we can handle this entirely in web3-gateway in a clunky fashion, by detecting if the revert string is a valid base64 encoding and then return the custom error instead of wrapping it again in Error(string) encoding, there is overlap with valid revert strings and base64 data.

I suggest returning the raw revert data, base64 encoded, without doing custom parsing. Which removes the special case from both runtime-sdk/evm & oasis-web3-gateway. Existing reverts with string messages will continue to work as expected, and custom reverts will start working. But - this will cause a small amount of time when errors are broken until all the web3 gateways are upgraded after runtime-sdk.

Alternatively, we can do the clunky workaround, and add a workaround to a workaround (presumably the workaround was 'I want simple strings returned in error messages')

kostko:
Fixes #1478 

Instead of attempting to decode simple ABI-encoded error strings, just return the entire data, base64-encoded.


runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -214,47 +214,10 @@ impl From<evm::ExitFatal> for Error {
...
-            if length > MAX_REASON_SIZE {
-                length = MAX_REASON_SIZE;
-            }
-            let reason =
-                String::from_utf8_lossy(&data[FIELD_REASON_START..FIELD_REASON_START + length]);
-            Err(Error::Reverted(reason.to_string()))
-        }
+        // Encode raw revert reason as Base64, up to the maximum length of 1024 bytes.
+        evm::ExitReason::Revert(_) => Err(Error::Reverted(base64::encode(
+            &data[..data.len().clamp(0, 1024)],
```
CedarMist, line 219: I expect it to be unlikely that there will be 1kb of errors returned via revert.  After base64 encoding the full return value will be approx 1400 bytes.

Is there a specific reason that reverts are limited to 1kb but success isn't?

CedarMist (approved):

runtime-sdk/modules/evm/src/lib.rs
kostko: Good point, let me think about it.

ptrus (approved):

PR #1483, opened 2023-09-12
runtime-sdk: Implement automatic MigrationHandler derivation

kostko:
TODO
* [x] Tests for the new `MigrationHandler` derivation.
* [x] Also derive the `Module` trait.


runtime-sdk-macros/src/module_derive/migration_handler.rs
```diff
@@ -0,0 +1,167 @@
...
+            syn::ImplItem::Fn(ref f) => {
+                // Check whether a `migration` attribute is set for the method.
+                if let Some(attrs) = parse_attrs(&f.attrs) {
+                    self.migrate_fns.push(MigrateFn {
+                        ident: f.sig.ident.clone(),
+                        from_version: attrs.from_version,
+                        item,
+                    });
+
+                    None
```
peternose, line 49:
```suggestion
                    None // Take the item.
```

runtime-sdk-macros/src/module_derive/migration_handler.rs
```diff
@@ -0,0 +1,167 @@
...
+
+    fn derive(&mut self, generics: &syn::Generics, ty: &Box<syn::Type>) -> TokenStream {
+        let genesis_ty = if let Some(genesis_ty) = &self.genesis_ty {
+            genesis_ty
+        } else {
+            return quote! {};
+        };
+
+        let mut seen_versions = HashSet::new();
+        let (migrate_fns, mut migrate_arms): (Vec<_>, Vec<_>) = self.migrate_fns.iter().map(|f| {
```
peternose, line 66: Should we order `migrate_fns` first?

runtime-sdk-macros/src/module_derive/migration_handler.rs
```diff
@@ -0,0 +1,167 @@
...
+                    genesis: Self::Genesis,
+                ) -> bool {
+                    let mut version = meta.versions.get(Self::NAME).copied().unwrap_or_default();
+                    if version == Self::VERSION {
+                        return false; // Already the latest version.
+                    }
+
+                    #(#migrate_arms)*
+
+                    if version != Self::VERSION {
```
peternose, line 111: Is there a problem if one defines migration from `Self::VERSION` to `Self::VERSION + 1`? Should we exclude this migration?

runtime-sdk-macros/src/lib.rs
```diff
@@ -91,6 +91,17 @@ pub fn handler(_args: TokenStream, input: TokenStream) -> TokenStream {
     input
 }
 
+/// A helper attribute for `#[sdk_derive(...)]`. It doesn't do anything on its own;
+/// it only marks functions that represent a module state migration.
+///
+/// The permitted forms are:
+///  - `#[migration(init)]`: Marks the initial (genesis) migration.
+///  - `#[migration(from = 1)]`: Marks a migration from version 1.
```
peternose, line 99:
```suggestion
///  - `#[migration(from = v)]`: Marks a migration from version v to v+1, where v is a non-negative integer.
```

runtime-sdk-macros/src/module_derive/mod.rs
```diff
@@ -678,12 +335,136 @@ mod tests {
...
+                            fn my_method_call() -> () {}
+                        }
+                    };
+                )
+            }),
+        );
+    }
+
+    #[test]
+    fn generate_module_impl() {
```
peternose, line 382: These tests should be elsewhere, but I guess is fine for now.

runtime-sdk-macros/src/module_derive/mod.rs
```diff
@@ -1,272 +1,59 @@
...
-                fn supported_methods() -> Vec<core_types::MethodHandlerInfo> {
-                    vec![ #(
-                        core_types::MethodHandlerInfo {
-                            kind: #handler_kinds,
-                            name: #handler_names.to_string(),
-                        },
-                    )* ]
-                }
+        for deriver in derivers.iter_mut() {
+            item = deriver.preprocess(item.take().expect("no deriver claimed the item"));
```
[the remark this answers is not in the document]
kostko: But we continue the outer loop? So the expectation is that "no deriver claimed the item" so far (and it is still there). But also feel free to suggest something that is more clear.

runtime-sdk-macros/src/module_derive/migration_handler.rs
kostko: Ideally we could exclude this at compile time, but unfortunately it seems this doesn't seem straightforward due to `Self` use in const functions.

peternose (approved):

PR #1484, opened 2023-09-12
ts-web: eliminate `video: false`, which is now default

pro-wh:
to be put on that branch after dependabot rebases it


pro-wh:
oop

PR #1500, opened 2023-09-18
runtime-sdk/modules/evm: Revert "drop signed queries in new contracts"

kostko:
Given further thought this seems too much like a foot gun and may need some more thought.


pro-wh:
https://github.com/oasisprotocol/oasis-sdk/pull/1481 cross reference

peternose (approved):

PR #1510, opened 2023-09-23
runtime-sdk: Add support for (un)delegation receipts

kostko:
This adds support for (un)delegation receipts so contracts can more easily do delegation accounting.

TODO
* [x] More tests.


runtime-sdk/src/modules/consensus_accounts/mod.rs
```diff
@@ -450,34 +461,77 @@ impl<Accounts: modules::accounts::API, Consensus: modules::consensus::API>
...
+        let store_receipt = body.receipt > 0;
+        if store_receipt {
+            <C::Runtime as Runtime>::Core::use_tx_gas(ctx, params.gas_costs.store_receipt)?;
+        }
 
         // Check whether delegate is allowed.
         if params.disable_delegate {
             return Err(Error::Forbidden);
         }
+        // Make sure receipts can only be requested internally (e.g. via subcalls).
```
nhynes, line 473: why is this required? what's wrong with external receipt storage other than it being weird?

runtime-sdk/src/modules/consensus_accounts/mod.rs
```diff
@@ -450,34 +461,77 @@ impl<Accounts: modules::accounts::API, Consensus: modules::consensus::API>
...
         // Signer.
         let signer = &ctx.tx_auth_info().signer_info[0];
         let from = signer.address_spec.address();
-        let nonce = signer.nonce;
-        Self::delegate(ctx, from, nonce, body.to, body.amount)
+        let nonce = if store_receipt {
+            body.receipt // Use receipt identifier as the nonce.
+        } else {
+            signer.nonce // Use signer nonce as the nonce.
+        };
```
nhynes, line 485:
```suggestion
        let nonce = store_receipt.then_some(body.receipt).unwrap_or(signer.nonce);
```
FYI

runtime-sdk/src/modules/consensus_accounts/mod.rs
```diff
@@ -450,34 +461,77 @@ impl<Accounts: modules::accounts::API, Consensus: modules::consensus::API>
...
         let signer = &ctx.tx_auth_info().signer_info[0];
         let to = signer.address_spec.address();
-        let nonce = signer.nonce;
-        Self::undelegate(ctx, body.from, nonce, to, body.shares)
+        let nonce = if store_receipt {
+            body.receipt // Use receipt identifer as the nonce.
+        } else {
+            signer.nonce // Use signer nonce as the nonce.
+        };
+        Self::undelegate(ctx, body.from, nonce, to, body.shares, store_receipt)
```
nhynes, line 515: this is a fair amount of duplicated code. can you split out the common stuff into a function that takes the other functions as input and then does the verification and nonce setup?

runtime-sdk/src/modules/consensus_accounts/mod.rs
```diff
@@ -679,14 +772,35 @@ impl<Accounts: modules::accounts::API, Consensus: modules::consensus::API>
         let result: ReclaimEscrowResult = cbor::from_value(result).unwrap();
         let debonding_shares = result.debonding_shares.try_into().unwrap();
 
-        state::add_undelegation(
+        let receipt = if context.receipt {
```
nhynes, line 775:
```suggestion
        let receipt = context.receipt.then_some(context.nonce).unwrap_or_default();
```

runtime-sdk/src/modules/consensus_accounts/mod.rs
kostko: Nothing wrong beyond what you said.

runtime-sdk/src/modules/consensus_accounts/mod.rs
kostko: Duplication between delegate and undelegate you mean?

tests/e2e/contracts/delegation/delegation.sol
```diff
@@ -0,0 +1,252 @@
...
+            }
+        }
+
+        // Compute how much we get from the pool and transfer the amount.
+        uint128 transferAmount = (pending.shares * pool.totalAmount) / pool.totalShares;
+        pending.to.transfer(transferAmount);
+
+        delete(pendingUndelegations[receiptId]);
+    }
+}
```
nhynes, line 252: strictly speaking, for just testing, I would have passed the cbor directly into the contract and have it proxy the subcall instead. would have been a bit clearer what's going on, too.

nhynes (commented):
looks normal to me

tests/e2e/contracts/delegation/delegation.sol
kostko: Yeah thought about that, I just wanted to write some Solidity to see if the API was viable. Any thoughts about the API, does it make sense?

CedarMist:
There's one thing I need to confirm:

The same `receiptId` will be used twice for the `UndelegateStart` & `UndelegateDone` receipts.

`add_undelegation` batches them together on a per-epoch basis, so if I call `undelegate` twice in a row with two sequential receipt IDs, fetching the two `undelegateStart` tickets will return the same `endReceiptId` - which will be the `receiptId` of the first `undelegate` call.

kostko:
Yes, note that this is batched per epoch per address the contract is undelegating from.

tests/e2e/contracts/delegation/delegation.sol
ptrus: Could use an existing solidity cbor lib? That would make it clearer to show how it would look like in practice. Unless there's too much work incorporating that into the existing unit tests.

CedarMist: I'm in the process of re-working the CBOR stuff (it was an easy fix to support 64bit ints, but need to do better parsing).

I'm happy with the unit test as a starting point and will be submitting a PR for sapphire-paratime repo when the updated sapphire-dev image is available.

ptrus (approved):

tests/e2e/contracts/delegation/delegation.sol
kostko: > Could use an existing solidity cbor lib? That would make it clearer to show how it would look like in practice. Unless there's too much work incorporating that into the existing unit tests.

Not easily for these E2E tests, we would need to adapt the pipeline a bit.

runtime-sdk/src/modules/consensus_accounts/types.rs
```diff
@@ -29,13 +29,68 @@ pub struct Withdraw {
...
+#[derive(Clone, Debug, Default, cbor::Encode, cbor::Decode)]
+pub struct TakeReceipt {
+    pub kind: ReceiptKind,
+    pub id: u64,
+}
+
+/// A receipt.
+#[derive(Clone, Debug, Default, cbor::Encode, cbor::Decode)]
+#[cfg_attr(test, derive(PartialEq, Eq))]
+pub struct Receipt {
```
peternose, line 74: I prefer splitting this into multiple structs, as it is hard to know which fields are/can be set by different kinds.

kostko: Yeah it is just slightly annoying with the current deserializer.

PR #1517, opened 2023-10-03
evm: add sha384 precompile

nhynes:
This enhances the SHA512 precompile suite with the recently very popular SHA384 routine.


runtime-sdk/modules/evm/src/precompile/sha512.rs
kostko, line 1: Should this module be renamed given that it is no longer just `sha512`? Maybe something like `hash` or `sha2` or similar?

nhynes: I didn't mind because sha384 is just 0.75 sha512s in a trenchcoat. Up to you.

runtime-sdk/modules/evm/src/precompile/mod.rs
```diff
@@ -133,6 +133,7 @@ impl<Cfg: Config, B: EVMBackendExt> PrecompileSet for Precompiles<'_, Cfg, B> {
             (1, 1, 1) => sha512::call_sha512_256(handle),
             (1, 1, 2) => sha512::call_sha512(handle),
             (1, 1, 3) => subcall::call_subcall(handle, self.backend),
+            (1, 1, 4) => sha512::call_sha384(handle),
```
peternose, line 136: I think you can also move this up (and reorder) as oasis specific group hasn't been deployed yet.

kostko: It has been deployed to Testnet.

nhynes: I've stopped developing on networks because they broke their testnet. I know I have no right to be mad, but it's annoying nonetheless. I'd rather avoid that if the consequence is just a few numbers being out of order :)

peternose: Agree, I forgot that it was already deployed.

kostko (approved):

PR #1518, opened 2023-10-03
runtime-sdk: bump crypto deps

nhynes:
I had some trouble getting `p384` working because `sha2`, `ed25519_dalek`, and a few other crypto deps were a version behind. Such is the state of Rust crypto. Anyway, this PR does not change any functionality, but I did remove one test for rejecting all-zero ed25519 private keys because `[0u8; 32]` is a valid private key, just not one you'd want to use.

### DIFF
--- a/runtime-sdk/src/crypto/signature/ed25519.rs
+++ b/runtime-sdk/src/crypto/signature/ed25519.rs
@@ -1,11 +1,9 @@
 //! Ed25519 signatures.
 use std::convert::TryInto;
 
-use curve25519_dalek::{
-    digest::{consts::U64, Digest},
-    edwards::CompressedEdwardsY,
-};
-use sha2::Sha512Trunc256;
+use curve25519_dalek::{digest::consts::U64, edwards::CompressedEdwardsY};
+use ed25519_dalek::Signer as _;
+use sha2::{Digest as _, Sha512_256};
 
 use oasis_core_runtime::common::crypto::signature::{
     PublicKey as CorePublicKey, Signature as CoreSignature,
@@ -85,7 +83,9 @@ impl PublicKey {
             .as_ref()
             .try_into()
             .map_err(|_| Error::MalformedSignature)?;
-        let pk = ed25519_dalek::PublicKey::from_bytes(self.as_bytes())
+        let pk: ed25519_dalek::VerifyingKey = self
+            .as_bytes()
+            .try_into()
             .map_err(|_| Error::MalformedPublicKey)?;
         pk.verify_prehashed(digest, None, &sig)
             .map_err(|_| Error::VerificationFailed)
@@ -118,7 +118,7 @@ impl From<PublicKey> for CorePublicKey {
 
 /// A memory-backed signer for Ed25519.
 pub struct MemorySigner {
-    sk: ed25519_dalek::ExpandedSecretKey,
+    sk: ed25519_dalek::SigningKey,
 }
 
 impl MemorySigner {
@@ -126,9 +126,8 @@ impl MemorySigner {
     where
         D: ed25519_dalek::Digest<OutputSize = U64>,
     {
-        let pk = ed25519_dalek::PublicKey::from(&self.sk);
         self.sk
-            .sign_prehashed(digest, &pk, None)
+            .sign_prehashed(digest, None)
             .map_err(|_| Error::SigningError)
             .map(|sig| sig.to_bytes().to_vec().into())
     }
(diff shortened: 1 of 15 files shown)


runtime-sdk/src/crypto/signature/ed25519.rs
```diff
@@ -118,33 +118,29 @@ impl From<PublicKey> for CorePublicKey {
...
-        let esk = ed25519_dalek::ExpandedSecretKey::from(&sk);
-        Ok(Self { sk: esk })
+        Self::from_bytes(seed)
     }
 
     fn from_bytes(bytes: &[u8]) -> Result<Self, Error> {
         Ok(Self {
-            sk: ed25519_dalek::ExpandedSecretKey::from_bytes(bytes)
-                .map_err(|_| Error::MalformedPrivateKey)?,
+            sk: bytes.try_into().map_err(|_| Error::MalformedPrivateKey)?,
```
kostko, line 143: I believe this (and corresponding `to_bytes` due to changed `sk` type) is breaking in case someone stored the result (key) in storage and/or is expecting the key format to be the old one. Not sure why we previously returned the expanded key (which is the wrong thing to do), I guess these weird ed25519 key variants contributed to this not being cought during review. :facepalm:

nhynes: that's okay, I can make it backwards compatible and add a test for it

runtime-sdk/src/crypto/signature/ed25519.rs
```diff
@@ -118,60 +118,180 @@ impl From<PublicKey> for CorePublicKey {
 
 /// A memory-backed signer for Ed25519.
 pub struct MemorySigner {
-    sk: ed25519_dalek::ExpandedSecretKey,
+    key: Key,
+}
+
+/// The original version of this signer returned the "expanded" signing key from `to_bytes`.
+/// If the contract passes in one of these, it's gets the same thing out. New invocations get the "proper" version.
```
nhynes, line 125: I propose that this is an okay thing to do because it was never documented what exactly is returned by the precompile 😐.

runtime-sdk/src/crypto/signature/ed25519.rs
```diff
@@ -118,60 +118,180 @@ impl From<PublicKey> for CorePublicKey {
...
+    key: Key,
+}
+
+/// The original version of this signer returned the "expanded" secret key from `to_bytes`.
+/// If the contract passes in one of these, it's gets the same thing out. New invocations get the "proper" version.
+enum Key {
+    Expanded {
+        esk: ed25519_dalek::hazmat::ExpandedSecretKey,
+        /// The hash output that is used to create the "expanded" secret key. It is stored to return from `from_bytes` because `esk` is a function of the hash, not the hash directly.
+        hash: [u8; 64],
```
peternose, line 130: Hash should be zeroized on drop.

runtime-sdk/src/crypto/signature/ed25519.rs
```diff
@@ -118,60 +118,180 @@ impl From<PublicKey> for CorePublicKey {
...
+    },
+    Regular(ed25519_dalek::SigningKey),
+}
+
+impl Key {
+    fn from_bytes(bytes: &[u8]) -> Result<Self, Error> {
+        if bytes.len() == 64 {
```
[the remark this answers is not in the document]
nhynes: I generally don't `match` on non-string literals to avoid the extra indentation and syntax noise, but I'll change it to a `match` since you prefer it.

runtime-sdk/src/crypto/signature/ed25519.rs
nhynes: It belongs on `Key`, but I can make the comment better.

runtime-sdk/src/crypto/signature/ed25519.rs
```diff
@@ -118,60 +118,184 @@ impl From<PublicKey> for CorePublicKey {
...
+}
+
+/// The original version of the Ed25519 signer returned the "expanded" secret key from `to_bytes`.
+/// A contract may have stored the "expanded" key and expects its use to continue to succeed.
+/// For backwards compatibility, the signer works with both "expanded" and regular keys.
+/// New invocations receive a regular/proper key, and from-"expanded" ones get the old behavior.
+enum Key {
+    Expanded {
+        esk: ed25519_dalek::hazmat::ExpandedSecretKey,
+        /// The hash output that is used to create the "expanded" secret key. It is stored to return from `from_bytes` because `esk` is a function of the hash, not the hash directly.
```
kostko, line 131: Can you wrap this line? Too bad that `esk` doesn't have a `to_bytes`.

runtime-sdk/src/crypto/signature/ed25519.rs
```diff
@@ -118,60 +118,184 @@ impl From<PublicKey> for CorePublicKey {
...
+        match bytes.len() {
+            // It's a new/correct-style secret key.
+            32 => bytes
+                .try_into()
+                .map(ed25519_dalek::SigningKey::from_bytes)
+                .map(Self::Regular)
+                .map_err(|_| Error::MalformedPrivateKey),
+            // It's an "expanded" secret key, which is treated as the output of a 64-byte hash function.
+            64 => {
+                let hash: [u8; 64] = bytes.try_into().unwrap();
```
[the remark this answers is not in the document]
nhynes: this operation is infallible. it's just a length check followed by `copy_from_slice`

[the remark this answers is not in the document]
kostko: Oops, I forgot that the earlier match is checking the length, nevermind :blush:

kostko (approved):

PR #1519, opened 2023-10-03
runtime-sdk: add p384 signing & verification

nhynes:


runtime-sdk/src/crypto/signature/secp256r1.rs
```diff
@@ -41,22 +41,10 @@ impl PublicKey {
         message: &[u8],
         signature: &Signature,
     ) -> Result<(), Error> {
-        // Note that we must use Sha512_256 instead of our Hash here,
```
nhynes, line 44: The comment didn't seem to match the code, so I updated it to match the p384 implementation.

kostko: What do you mean it didn't reflect the code? The new implementation is better though, but it still must use `Sha512_256` instead of our `Hash` for the reason it was mentioned earlier.

nhynes: I guess I didn't understand what `Hash` meant since it wasn't defined in the file and I couldn't tell what it's the "same thing" as.

kostko (commented):
If you want this to be accessible from EVM, make sure to also update the gas costs in the confidential precompiles. Ideally you would add some benchmarks to establish proper costs (see how it was done for the others).

Otherwise looks good.

nhynes:
The benches look like
```
test precompile::confidential::test::bench_sign_secp256r1_prehashed_sha256   ... bench:     288,916 ns/iter (+/- 1,086)
test precompile::confidential::test::bench_sign_secp384r1_prehashed_sha384   ... bench:   1,214,954 ns/iter (+/- 6,736)
test precompile::confidential::test::bench_verify_secp256r1_prehashed_sha256 ... bench:     277,456 ns/iter (+/- 1,579
)
test precompile::confidential::test::bench_verify_secp384r1_prehashed_sha384 ... bench:   1,178,009 ns/iter (+/- 4,977
)
```

runtime-sdk/src/crypto/signature/secp384r1.rs
```diff
@@ -0,0 +1,144 @@
...
+    /// Return a byte representation of this public key.
+    pub fn as_bytes(&self) -> &[u8] {
+        self.0.as_bytes()
+    }
+
+    /// Construct a public key from a slice of bytes.
+    pub fn from_bytes(bytes: &[u8]) -> Result<Self, Error> {
+        let ep = p384::EncodedPoint::from_bytes(bytes).map_err(|_| Error::MalformedPublicKey)?;
+        if !ep.is_compressed() {
+            // This should never happen due to the size check above.
```
peternose, line 27: Which size check above?
Can we remove this size check also on similar points?

runtime-sdk/src/crypto/signature/secp384r1.rs
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    /// Verify a signature.
+    pub fn verify(
+        &self,
+        context: &[u8],
+        message: &[u8],
+        signature: &Signature,
+    ) -> Result<(), Error> {
+        self.verify_digest(
```
peternose, line 40: I would change Secp256r1 and Secp256k1 a bit, just to be consistent:
- Remove `digest` from `verify`.
- Update `Signer`.
```
    fn sign(&self, context: &[u8], message: &[u8]) -> Result<Signature, Error> {
        let digest = Sha512_256::new()
            .chain_update(context)
            .chain_update(message);
            ...
    }
```

peternose:
> so about 4.8x as slow

Did you mean 4.2x?

runtime-sdk/src/crypto/signature/secp384r1.rs
nhynes: The size check I deleted but the hunk didn't get staged 😛 
The other curves implement a quick size check in `from_bytes`, so I updated them

nhynes:
> Did you mean 4.2x?

It looks that way from the one sample, but it's highly variable, and sometimes I get 4.8x, so I took the max so that we can safely reduce it later.

kostko (approved):

peternose (approved):

PR #1529, opened 2023-10-16
tools/gomod-updater: Fix Go Dependabot PR's automatically

ptrus:
Makes dependabot useful for go projects in this repo:

- Adds tools/gomod-updater which supports updating the provided package in all go projects in the repo that use the package.
- Setups a Github workflow that runs on all dependabot PR's and updates the package in all projects.

See example: https://github.com/ptrus/oasis-sdk/pull/3


kostko (approved):
Cool!

PR #1534, opened 2023-10-17
gomod-updater: tidy all projects

ptrus:


peternose:
Sometimes, `go mod tidy` returns different results if we delete indirect imports. Just saying, no need to change.

PR #1564, opened 2023-11-21
runtime/consensus: Add support for querying round roots

ptrus:


runtime-sdk/src/modules/consensus/mod.rs
```diff
@@ -231,6 +255,27 @@ impl Module {
         // Set genesis parameters.
         Self::set_params(genesis.parameters);
     }
+
+    #[handler(call = "consensus.RoundRoot", internal)]
+    fn internal_round_root<C: TxContext>(
```
peternose, line 260: Why do we return only one root, if we already fetched both of them?

runtime-sdk/src/modules/consensus/types.rs
```diff
@@ -0,0 +1,27 @@
+use oasis_core_runtime::common::namespace::Namespace;
+
+/// Kind of root.
+#[derive(Clone, Debug, Default, cbor::Encode, cbor::Decode)]
+#[cfg_attr(test, derive(PartialEq, Eq))]
+#[repr(u8)]
+pub enum RootKind {
+    #[default]
+    Invalid = 0,
```
peternose, line 9: Can we remove this?

runtime-sdk/src/modules/consensus/mod.rs
ptrus: This (internal) call version is intended to be used from smart contract modules (via the subcalls feature). It was suggested to keep the response as simple as possible to make the cbor parsing in contracts (solidity) easier. 

Higher level code (e.g. a runtime module) would use the `round_roots` method directly and it would get the more structured result with both hashes.

runtime-sdk/src/modules/consensus/types.rs
ptrus: This follows the same pattern as it is already used by the other internal call in the code (`consensus.TakeReceipt`). Not actually sure, but I guessed we have the invalid as the default variant as an additional safeguard, since the cbor encoding can be a bit messy in soldity. Or my other guess would be it has to do with gas costs in evm.

@kostko please confirm what is the reasoning and if it makes sense to have it here as well.

runtime-sdk/src/modules/consensus/mod.rs
peternose: > It was suggested to keep the response as simple as possible to make the cbor parsing in contracts (solidity) simpler.

Makes sense.

> This (internal)-call version is intended to be used from smart contract modules (via the subcalls feature). 

Yes, and the caller will now have to pay double gas for the same operation. That's why we could return both roots.

kostko: What is the use case where the caller would actually need both roots for the same round? Usually you would fetch the roots to verify a specific Merkle proof.

tests/e2e/evmtest.go
```diff
@@ -1492,6 +1492,54 @@ func DelegationReceiptsTest(_ *RuntimeScenario, log *logging.Logger, conn *grpc.
...
+	if err != nil {
+		return fmt.Errorf("failed to unpack test_consensus_round_root result: %w", err)
+	}
+	stateHash := results[0].([]byte)
+	if len(stateHash) != 34 { // 2 bytes CBOR header + 32 bytes hash.
+		return fmt.Errorf("invalid test_consensus_round_root response, expected state hash, got: %v", stateHash)
+	}
+
+	/*
+		// Round Roots not exposed in the go roothash client at the moment :(
```
kostko, line 1528: Feel free to add the query.

runtime-sdk/src/modules/consensus/types.rs
kostko: I'm not sure why this would be useful, we could just drop the invalid variant and start the valid ones with 1 (as it is now).

ptrus: Ok I removed the Invalid variant here.

tests/e2e/evmtest.go
ptrus: Will do but don't want to block this PR until the next oasis-core release.

kostko (approved):

PR #1591, opened 2023-12-12
runtime-sdk/evm: Emit Transfer event for contract-initiated transfers

mitjat:
Previously, any native token transfers that were initiated from within the EVM (typically, via a `(address).transfer(amount)` in Solidity) did _not_ generate a `Transfer` event in the `accounts` SDK module.
As a result, it was hard to impossible to track native token balance changes from outside the EVM. The behavior is also asymmetric with transfers that are initiated via the `accounts` module; those do emit the `accounts.Transfer` event.

This PR changes the `evm` module so that within-EVM native token transfers also emit an `accounts.Transfer` event.


runtime-sdk/modules/evm/src/backend.rs
```diff
@@ -554,12 +554,13 @@ impl<'ctx, 'backend, 'config, C: TxContext, Cfg: Config> StackState<'config>
     }
 
     fn transfer(&mut self, transfer: Transfer) -> Result<(), ExitError> {
+        let mut ctx = self.backend.ctx.borrow_mut();
```
kostko, line 557: Ah I just remembered why we used the "silent" transfer. The ctx you are using is not substate-aware so you cannot use it like this as the substate (EVM's notion of context) containing the internal transfer action may be reverted, but I think your implementation will incorrectly emit an event in this case.

kostko (changes requested):
I believe the current implementation will result in some potential issues during substate reverts, see comment.

kostko (commented):
Please rebase.

runtime-sdk/modules/evm/src/backend.rs
mitjat: Now rebased on top of your https://github.com/oasisprotocol/oasis-sdk/pull/1601 that should resolve this. Thank you!

kostko (approved):

PR #1601, opened 2023-12-21
runtime-sdk: Refactor state handling

kostko:
The `Context` is now only used to serve read-only environment state (e.g. information about the current block) and provide the `Runtime` type for static dispatch. All mutable state (including storage) has been moved into a new type, `State`, which supports nesting and transactions (similar to the previous `CurrentStore` which only provided this for storage).

This allows many things to be simplified and a better integration with the EVM's substates (which now no longer require hacks to handle events and messages).


runtime-sdk/src/state.rs
```diff
@@ -0,0 +1,1538 @@
...
+        matches!(self.mode, Mode::Execute)
+    }
+
+    /// Whether there is an active transaction in the current environment.
+    pub fn is_transaction(&self) -> bool {
+        self.tx.is_some()
+    }
+
+    /// An active transaction's index (order) within the block. If there is no active transaction,
+    /// this method will return zero.
```
ptrus, line 98: Should this and other tx-only methods panic if not called from within a transaction environment (like tx_auth_info does)?

runtime-sdk/src/callformat.rs
```diff
@@ -275,9 +276,9 @@ pub fn encrypt_result_x25519_deoxysii<C: Context>(
         .write_u32::<BigEndian>(index.try_into().unwrap())
         .unwrap();
     nonce.extend(&[0, 0, 0]);
-    if ctx.is_simulation() {
-        // Randomize the lower-order bytes of the nonce to facilitate private queries.
-        OsRng.fill_bytes(&mut nonce[deoxysii::NONCE_SIZE - 3..]);
+    if CurrentState::with(|state| !state.env().is_execute()) {
```
ptrus, line 279: This used to only check if it's simulation, was previous check wrong? Or do we only end up here either in simulation or execute? (and not check?)

kostko: The new check is _not_ execute (there is an `!`), so it covers simulation mode and be same as before, additionally covering check and pre-schedule (which should likely never reach this stage, but it seems like a good default to have).

ptrus: > additionally covering check and pre-schedule (which should likely never reach this stage, but it seems like a good default to have).

Yeah that's what i thought

runtime-sdk/modules/evm/src/state.rs
```diff
@@ -70,26 +71,27 @@ where
 
     // These values are used to derive the confidential store nonce:
     let round = ctx.runtime_header().round;
-    let instance_count: usize = {
-        // One Context is used per tx batch, so the instance count will monotonically increase.
-        let cnt = *ctx
-            .value(CONTEXT_KEY_CONFIDENTIAL_STORE_INSTANCE_COUNT)
+    let instance_count: usize = CurrentState::with(|state| {
+        // One state is used per tx batch, so the instance count will monotonically increase.
```
mitjat, line 75: Shooting from the hip: You mention in the PR description that `State` "supports nesting and transactions". Is this comment still right? There's one State that keeps track of forks and reverts, rather than a tree of States?

kostko: Values behave differently as each block-wide value needs to exist for the duration of the whole block and implements its own semantics.

So block-wide values are not affected by reverts and there is also no clear need for this. We should probably update the documentation on value accessors to make this clear.

kostko: Updated documentation for `block_value` and `local_value` methods.

runtime-sdk/src/state.rs
```diff
@@ -0,0 +1,1545 @@
...
+        };
+        // Apply options to allow customization of the root state.
+        opts.apply(&mut root);
+
+        CURRENT.with(|c| {
+            c.try_borrow_mut()
+                .expect("must not re-enter from with block")
+                .push(root)
+        });
+        let _guard = CurrentStateGuard; // Ensure current state is popped once we return.
```
pro-wh, line 705: I usually see these things with a symmetry to them, such that creating one is combined with pushing the state

runtime-sdk/src/dispatcher.rs
```diff
@@ -272,64 +271,55 @@ impl<R: Runtime> Dispatcher<R> {
         }
         let tx_auth_info = tx.auth_info.clone();
         let is_read_only = tx.call.read_only;
+        let call = tx.call.clone(); // TODO: Avoid clone.
```
pro-wh, line 274: what's needed here? dispatch_tx_call needing a copy as "call" as well as the whole tx within the options tx meta?

pro-wh (commented):
new interface looks good, and tests pass. still going through some parts

runtime-sdk/src/state.rs
```diff
@@ -0,0 +1,1545 @@
...
+        if let Some(ref parent) = self.parent {
+            parent.level() + 1
+        } else {
+            0
+        }
+    }
+}
+
+thread_local! {
+    static CURRENT: RefCell<Vec<State>> = RefCell::new(Vec::new());
```
pro-wh, line 601: can you summarize what mechanisms use the state .parent chain and what mechanisms use the CURRENT vector?

kostko: The `CURRENT` vector is there so one can do `CurrentState::enter` multiple times (each `enter` results in a new element in this vector and the last element is the active one), starting with a fresh root state. The `.parent` chain is then derived from one such root state and is what is actually used most of the time (e.g. in practical use in the SDK there are no nested enters). All mechanisms only look at the last state in the `CURRENT` vector which is only used to support nested enters.

Alternatively we could forbid nested `CurrentState::enter` (e.g. panic if entered again instead of stacking), but supporting it may be useful (e.g. at least in tests).

runtime-sdk/src/state.rs
kostko: Yeah, it is pushed just above.

runtime-sdk/src/callformat.rs
```diff
@@ -275,9 +276,9 @@ pub fn encrypt_result_x25519_deoxysii<C: Context>(
         .write_u32::<BigEndian>(index.try_into().unwrap())
         .unwrap();
     nonce.extend(&[0, 0, 0]);
-    if ctx.is_simulation() {
-        // Randomize the lower-order bytes of the nonce to facilitate private queries.
-        OsRng.fill_bytes(&mut nonce[deoxysii::NONCE_SIZE - 3..]);
```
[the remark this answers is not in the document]
kostko: Not sure why this was limited to just the low-order bytes. More entropy seems better which is why I changed this.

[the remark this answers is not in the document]
pro-wh: oh I checked above the context and it says

> // Generate nonce for the output as Round (8 bytes) || Index (4 bytes) || 00 00 00.

I guess as long as nothing tries to read that data :shrug: <sup>:shrug:</sup>

runtime-sdk/src/state.rs
pro-wh: what is nested enter used for? something like when you're in the middle of a transaction in a block, and you want to simulate something happening before the block started?

`.parent` chain is used for things like smart contracts invoking other methods in my reading, is that right

kostko: In practice nested enter is not used outside tests. Besides what you said it can also specify a completely different state store.

The "`.parent` state chain" is used for everything currently (simulation, subcalls from contracts, regular tx execution, ...).

runtime-sdk/src/callformat.rs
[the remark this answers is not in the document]
kostko: That scheme is needed during block execution as nonces need to be deterministic to reduce the possibility of nonce reuse (even though we are using a deterministic scheme). There is no reason why one would want to "parse"/interpret a nonce.

For non-execution modes, there is no reason for the nonce to be deterministic and having it be completely random using local secure entropy seems safer.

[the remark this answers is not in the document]
pro-wh: mind if we branch before the various nonce.write_*() and nonce.extend() then

runtime-sdk/src/state.rs
pro-wh: ah completely different state store. thanks for the summary, resolving

pro-wh (approved):
my questions are answered, approving

runtime-sdk/src/callformat.rs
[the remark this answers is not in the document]
kostko: Good idea, will change.

runtime-sdk/src/state.rs
```diff
@@ -0,0 +1,1545 @@
...
+    where
+        F: FnOnce() -> Rs,
+        Rs: Into<TransactionResult<R>>,
+    {
+        let level = Self::with(|state| {
+            state.open();
+            opts.apply(state);
+            state.level()
+        });
+        let _guard = TransactionGuard(level); // Ensure transaction is always closed.
```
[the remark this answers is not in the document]
kostko: This is to cleanup on panic.

runtime-sdk/src/state.rs
```diff
@@ -0,0 +1,1545 @@
...
+            ..self
+        }
+    }
+
+    /// Apply the requested modifications to the given state.
+    fn apply(self, state: &mut State) {
+        if let Some(mode) = self.mode {
+            // Change mode.
+            state.env.mode = mode;
+            state.rng.as_mut().unwrap().set_mode(mode);
```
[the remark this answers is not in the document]
kostko: The reason for this is due to the same `Options` being used for root init and for child states. We would only want to set the rng mode on root state init and leave it unchanged from there onwards.

Let me think about refactoring the API a bit.

tests/runtimes/benchmarking/src/runtime/mod.rs
```diff
@@ -58,23 +59,25 @@ pub struct Module<Accounts: modules::accounts::API> {
 
 // Impls.
 impl<Accounts: modules::accounts::API> Module<Accounts> {
-    fn tx_accounts_mint<C: TxContext>(ctx: &mut C, body: types::AccountsMint) -> Result<(), Error> {
+    fn tx_accounts_mint<C: Context>(_ctx: &C, body: types::AccountsMint) -> Result<(), Error> {
         // XXX: no gas costs atm.
 
-        Accounts::mint(ctx, ctx.tx_caller_address(), &body.amount)?;
+        let caller = CurrentState::with(|state| state.env().tx_caller_address());
```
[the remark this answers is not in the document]
kostko: We could even add some macros to make this nicer to use as it is used a lot.

runtime-sdk/src/state.rs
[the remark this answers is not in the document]
kostko: Changed so that the `RootRng` takes a mode argument at construction time, please take a look.

runtime-sdk/src/state.rs
```diff
@@ -0,0 +1,1545 @@
...
+    /// If the closure returns `TransactionResult::Commit(R)` then the child state is committed,
+    /// otherwise the child state is rolled back.
+    pub fn with_transaction_opts<F, R, Rs>(opts: Options, f: F) -> R
+    where
+        F: FnOnce() -> Rs,
+        Rs: Into<TransactionResult<R>>,
+    {
+        let level = Self::with(|state| {
+            state.open();
+            opts.apply(state);
```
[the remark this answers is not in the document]
kostko: Done, also renamed to `state.init` to make it more clear that this initializes the state with the given options (as it should not really be called multiple times).

runtime-sdk/src/callformat.rs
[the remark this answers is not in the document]
kostko: Done.

runtime-sdk/src/state.rs
```diff
@@ -0,0 +1,1573 @@
...
+        Self::with(|state| f(state.store()))
+    }
+
+    /// Run a closure with the environment of the currently active state.
+    ///
+    /// # Panics
+    ///
+    /// This method will panic if called outside `CurrentState::enter` or if any transaction methods
+    /// are called from the closure.
+    pub fn with_env<F, R>(f: F) -> R
```
peternose, line 823: Optional: There are still ~40 lines where this could be used instead of `with`, just search for `state.env()`.

runtime-sdk/src/state.rs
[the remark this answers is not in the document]
peternose: Looks good, although a cleaner solution would try to get rid of `valid` and change `RootRng` to a trait which is implemented twice, once for a valid and once for an invalid generator. But maybe that complicates other things.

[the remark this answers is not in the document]
peternose: Looks good, although a cleaner solution would try to get rid of `valid` and change `RootRng` to a trait which is implemented twice, once for a valid and once for an invalid generator. But maybe that complicates other things.

runtime-sdk/src/state.rs
peternose: Have the same question, as the caller cannot distinguish between the right and the default index.

runtime-sdk/src/state.rs
kostko: Should be fixed now.

peternose: I still see ~20 lines that could be changed.

runtime-sdk/src/state.rs
kostko: Changed, please take a look.

peternose (approved):

PR #1612, opened 2024-01-19
runtime-sdk: Support storage update gas costs

kostko:
Introduce the notion of storage gas which is charged for each byte of a key/value pair that is updated in storage. For removals only the size of the key is charged. The reason for this is to reflect actual costs that storage updates have (as any state update increases the size of the write log).

Storage gas is handled in a somewhat special way where any gas already used by transaction execution is taken into account. This means that storage gas is only charged in case a transaction would use more storage gas than execution gas in order to limit storage growth.

### DIFF
--- a/runtime-sdk/modules/contracts/src/abi/oasis/test.rs
+++ b/runtime-sdk/modules/contracts/src/abi/oasis/test.rs
@@ -5,7 +5,6 @@ use oasis_runtime_sdk::{
     error::Error as _,
     modules,
     modules::core,
-    state::CurrentState,
     testing::mock,
     types::{address::Address, transaction::CallFormat},
 };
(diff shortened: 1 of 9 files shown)


CedarMist:
Ethereum has the concept of 'hot' vs 'cold' storage I/O, which is like a thru-cache which refunds gas if the item exists in cache (so doing a 'hot' write costs less than doing a 'cold' write as many hot writes only result in the equivalent of a single cold write). These have been slowly tuned over time to reflect the average characteristics of nodes so for them (Ethereum etc.) it makes sense.

Wouldn't it make sense to modify the cost of the SLOAD & SSTORE operations than to add a Sapphire-specific gas calculation?

kostko:
> Wouldn't it make sense to modify the cost of the SLOAD & SSTORE operations than to add a Sapphire-specific gas calculation?

This mechanism is more general (works for all SDK modules) and to me easier to reason about (cost per byte which actually ends up in the write log).

ptrus (approved):
I think this makes sense since it will be much easier to reason about and properly charge storage usage in transactions.

PR #1638, opened 2024-02-13
client-sdk/config: Add Pontus-X chain

mitjat:
Motivation: The config being expanded here is used by Nexus to know what human-readable configs mean by the string "pontusx".

### DIFF
--- a/client-sdk/go/config/default.go
+++ b/client-sdk/go/config/default.go
@@ -95,6 +95,23 @@ var DefaultNetworks = Networks{
 							},
 						},
 					},
+
+					// Pontus-X on Testnet.
+					"pontusx": {
+						Description: "Pontus-X Testnet",
+						ID:          "0000000000000000000000000000000000000000000000004febe52eb412b421",
+						Denominations: map[string]*DenominationInfo{
+							NativeDenominationKey: {
+								Symbol:   "EUROe",
+								Decimals: 18,
+							},
+							// The consensus layer denomination when deposited into the runtime.
+							"TEST": {
+								Symbol:   "TEST",
+								Decimals: 18,
+							},
+						},
+					},
 				},
 			},
 		},


client-sdk/go/config/default.go
```diff
@@ -95,6 +95,18 @@ var DefaultNetworks = Networks{
...
 						},
 					},
+
+					// Pontus-X on Testnet.
+					"pontusx": {
+						Description: "Pontus-X Testnet",
+						ID:          "0000000000000000000000000000000000000000000000004febe52eb412b421",
+						Denominations: map[string]*DenominationInfo{
+							NativeDenominationKey: {
+								Symbol:   "PNTX",
```
mitjat, line 105: Other than the ID, the values are placeholders. Reviewers, please advise.
I don't know their preferred currency symbol or the number of decimals.

kostko: The number of decimals is correct, symbol should be "EUROe" and there is also a second denomination with identifier "TEST" (this is the consensus layer denomination when deposited into the runtime):
```golang
"TEST": {
    Symbol: "TEST",
    Decimals: 18,
}
```

mitjat: Huh, that might take some tweaks in Nexus still. We mostly support multiple runtime currencies, but it's very untested, and in some places we explicitly only care about the native denomination.

> "TEST" (this is the consensus layer denomination when deposited into the runtime

I take this to mean that TEST is the native denomination. This comment supports that: https://github.com/oasisprotocol/pontusx-paratime/blob/main/runtime/src/lib.rs#L233-L234 (`consensus_denomination()` resolves to `"TEST"`).
But code trumps comments, and this implies that `TEST` is not the native denomination:
https://github.com/oasisprotocol/pontusx-paratime/blob/main/runtime/src/lib.rs#L179-L182

I am confused. I thought "native denomination" was the "main" one, and was the one that was exchanged with the consensus layer: If somebody deposits consensus ROSE into the runtime, you receive NATIVE_DENOMINATION in the runtime, and vice versa with withdrawals. So that's not the case? How does one convert from runtime TEST (after depositing) to EUROe?

Can EVM contract interact with the `TEST` tokens? I assume not, because [the evm config](https://github.com/oasisprotocol/pontusx-paratime/blob/main/runtime/src/lib.rs#L258-L262) doesn't mention them, so it stands to reason that evm only "sees" the native denomination.

kostko: > I take this to mean that TEST is the native denomination.

No. (That linked comment is actually wrong and is a leftover from being copied over from Sapphire where this is the case.)

When you deposit consensus layer TEST into this runtime you get TEST inside the runtime and not the native denomination. There is no built-in conversion between TEST and EUROe.

> Can EVM contract interact with the TEST tokens?

Currently not, as the EVM is limited to a single native denomination. When ERC-20 mapping is supported, this additional denomination can be visible as an ERC-20 token inside the EVM. So currently this deposited TEST is not very useful on its own.

kostko (approved):

PR #1644, opened 2024-02-16
client-sdk/go: Fix denomination resolution

kostko:
When parsing the config file the denomination ID may get transformed to lower case which will prevent the denomination from being resolved correctly.

### DIFF
--- a/client-sdk/go/config/paratime.go
+++ b/client-sdk/go/config/paratime.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/oasisprotocol/oasis-core/go/common"
 )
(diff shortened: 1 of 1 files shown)


ptrus (approved):

client-sdk/go/config/paratime.go
```diff
@@ -155,8 +156,8 @@ func (p *ParaTime) GetDenominationInfo(d string) *DenominationInfo {
 	var di *DenominationInfo
 	if len(d) == 0 {
 		di = p.Denominations[NativeDenominationKey]
-	} else {
-		di = p.Denominations[d]
+	} else if di = p.Denominations[d]; di == nil {
```
peternose, line 159:
```go
	if d == "" {
		d = NativeDenominationKey
	}

	if di, ok := p.Denominations[d]; ok {
		return di
	}
	if di, ok := p.Denominations[strings.ToLower(d)]; ok {
		return di
	}
```

PR #1669, opened 2024-03-26
runtime-sdk/modules/evm: Add support for simulating CREATE

kostko:


runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -415,20 +389,44 @@ impl<Cfg: Config> API for Module<Cfg> {
             tx_metadata,
         ) = Self::decode_simulate_call_query(ctx, call)?;
 
-        let call_tx = transaction::Transaction {
+        let (method, body) = match address {
```
ptrus, line 392: Could remember/store into a variable if this is a call or create, to avoid matching on address below again (and repeating the comment)

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -473,8 +466,62 @@ impl<Cfg: Config> API for Module<Cfg> {
...
+        caller: H160,
+        value: U256,
+        init_code: Vec<u8>,
+        estimate_gas: bool,
+    ) -> Result<Vec<u8>, Error> {
+        Self::evm_execute(
+            ctx,
+            caller,
+            |exec, gas_limit| {
+                let address = exec.create_address(evm::CreateScheme::Legacy {
```
ptrus, line 505: This can likely be moved into `exit_reason.is_succeed()` branch? (or does the executor require that this is called before transact_create?)

ptrus (approved):

runtime-sdk/modules/evm/src/lib.rs
kostko: I believe there is a problem with address derivation as it depends on the nonce which can change during execution.

runtime-sdk/modules/evm/src/lib.rs
kostko: Good idea, refactored using a closure, take a look.

runtime-sdk/modules/evm/src/lib.rs
ptrus: Maybe add a comment for future reference.

runtime-sdk/modules/evm/src/lib.rs
ptrus: Looks good, thx!

runtime-sdk/modules/evm/src/lib.rs
kostko: Done.

CedarMist (changes requested):
It seems errors & reversions aren't being handled when simulating a contract creation via eth_call.

Oh, my bad, need to update oasis-web3-gateway to specify nil as the `to` field.

CedarMist (approved):
https://github.com/oasisprotocol/oasis-web3-gateway/pull/546
https://github.com/oasisprotocol/sapphire-paratime/pull/294

Issue #1673, opened 2024-03-29
runtime-sdk/evm: EIP-1559 `resolved_gas_price` is incorrect


CedarMist:
This is preventing [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559) transactions from being submitted as, unless `max_priority_fee_per_gas` is also set to 100 GWei the transaction will be rejected for having a too low gas price.

https://github.com/oasisprotocol/oasis-sdk/blob/a9c70cbf17cd8df87760bfcb35ed4f325864e57b/runtime-sdk/modules/evm/src/raw_tx.rs#L101-L105

According to the spec there is a `expected_base_fee_per_gas`, and the `max_fee_per_gas` field of the transaction must be at least that amount. In Sapphire's case that would be 100 Gwei.

The `priority_fee_per_gas` is then calculated as:

```python
priority_fee_per_gas = min(transaction.max_priority_fee_per_gas, transaction.max_fee_per_gas - block.base_fee_per_gas)
# signer pays both the priority fee and the base fee
effective_gas_price = priority_fee_per_gas + block.base_fee_per_gas
```

This causes a failure in the oasis-web3-gateway tests:

```
{"caller":"api.go:437","err":"client: transaction check failed: runtime error: module: core code: 20 message: gas price too low","level":"debug","method":"eth_sendRawTransaction","module":"eth_rpc","msg":"failed to submit transaction","ts":"2024-03-29T15:20:48.196176379Z"}
    tx_test.go:104: 
        	Error Trace:	/home/runner/work/oasis-web3-gateway/oasis-web3-gateway/tests/rpc/tx_test.go:104
        	            				/home/runner/work/oasis-web3-gateway/oasis-web3-gateway/tests/rpc/tx_test.go:137
        	Error:      	Expected nil, but got: &rpc.jsonError{Code:-32000, Message:"client: transaction check failed: runtime error: module: core code: 20 message: gas price too low", Data:interface {}(nil)}
        	Test:       	TestContractFailCreation
        	Messages:   	send transaction failed
--- FAIL: TestContractFailCreation (0.01s)
```

Caused by code like this:

```go
	// Create transaction
	tx := types.NewTx(&types.DynamicFeeTx{
		Nonce:     nonce,
		Value:     value,
		GasFeeCap: new(big.Int).SetUint64(GasLimit),
		Data:      code,
	})
```

PR #1674, opened 2024-03-30
runtime-sdk/modules/evm: Fix EIP-1559 gas price calculation

kostko:
Closes: #1673

### DIFF
--- a/runtime-sdk/modules/evm/src/raw_tx.rs
+++ b/runtime-sdk/modules/evm/src/raw_tx.rs
@@ -30,6 +30,7 @@ pub fn recover_low(
 pub fn decode(
     body: &[u8],
     expected_chain_id: Option<u64>,
+    min_gas_price: u128,
 ) -> Result<transaction::Transaction, anyhow::Error> {
     let (
         chain_id,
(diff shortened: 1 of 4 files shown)


runtime-sdk/modules/evm/src/raw_tx.rs
```diff
@@ -99,9 +100,13 @@ pub fn decode(
             let sig_recid = k256::ecdsa::RecoveryId::new(eth_tx.odd_y_parity, false);
             let message = ethereum::EIP1559TransactionMessage::from(eth_tx);
 
-            // Base fee is zero. Allocate only priority fee.
-            let resolved_gas_price =
-                std::cmp::min(message.max_fee_per_gas, message.max_priority_fee_per_gas);
+            let min_gas_price = min_gas_price.into();
```
peternose, line 103: I would rename a bit, to be in line with the EIP.

```rust
            let base_fee_per_gas = min_gas_price.into();
            let priority_fee_per_gas = std::cmp::min(
                message.max_priority_fee_per_gas,
                message.max_fee_per_gas.saturating_sub(base_fee_per_gas),
            );
            let effective_gas_price = priority_fee_per_gas.saturating_add(base_fee_per_gas);
```

Do we need to error if the following doesn't hold?
```rust
balance >= gas_limit * max_fee_per_gas
max_fee_per_gas >= max_priority_fee_per_gas
max_fee_per_gas >= block.base_fee_per_gas
```

peternose (approved):
The code looks okay, but I haven't tested whether this solves the problem.

runtime-sdk/modules/evm/src/raw_tx.rs
kostko: We don't need to check these things here as it will be handled by the core module during transaction checks. This is transaction decoding so it should do as little as possible.

peternose: I think that this decoding hides `max_priority_fee_per_gas` and `max_fee_per_gas` from the core, so the second inequality cannot be checked. So in this case, EIP would reject the transaction and we don't. No sure, if this is relevant, though.

kostko: Fixed.

peternose (approved):

PR #1703, opened 2024-04-30
.github/dependabot: update all go projects in a single PR

ptrus:
Also use the new [multi-directory-beta-feature](https://github.blog/changelog/2024-04-29-dependabot-multi-directory-configuration-public-beta-now-available/) to bump go dependency for all projects in a single PR.

This also deprecates the hacky gomod-updater tool.


.github/workflows/ci-dependabot-fixup.yml
```diff
@@ -1,118 +0,0 @@
...
-        file_list+=$(find . -type f -name 'go.mod' ! -path "$client_sdk" | awk -vORS=, '{ print $1 }' | sed 's/,$/\n/')
-        tools/gomod-updater/gomod-updater ${{ steps.extract.outputs.repo }} ${{ steps.extract.outputs.version }} --packages "$file_list"
-
-    - name: Commit and push all changed files
-      if: steps.check.outputs.skip != 'true'
-      env:
-        CI_COMMIT_MESSAGE: Dependabot dependencies fixup 👷
-        CI_COMMIT_AUTHOR: oasisprotocol-bot
-      run: |
-        git config --global user.name "${{ env.CI_COMMIT_AUTHOR }}"
```
lukaw3d, line 115: This was a bit more risky for injections than `$CI_COMMIT_AUTHOR`, but it's a constant `CI_COMMIT_AUTHOR: oasisprotocol-bot`

.github/workflows/ci-dependabot-fixup.yml
```diff
@@ -1,118 +0,0 @@
...
-
-    # Dependabot titles are:
-    #      <prefix>: bump <dependancy name> from <frmo_version> to <to_version> in <path>
-    # e.g. client-sdk/go: bump github.com/ethereum/go-ethereum from 1.12.1 to 1.13.3 in /client-sdk/go
-    # as long as the (configurable) <prefix> is without whitespace, the bellow parsing should work.
-    - name: Try extracting package name and version
-      if: steps.check.outputs.skip != 'true'
-      id: extract
-      run: |
-        title="${{ github.event.pull_request.title }}"
```
lukaw3d, line 92: This was unsafe - dependabot could had injected code with a title `aaaa"; echo "inject`
https://github.com/oasisprotocol/oasis-rosetta-gateway/pull/527

and we were using `pull_request_target` :scream:  https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token
> When a workflow is triggered by the [pull_request_target](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) event, the GITHUB_TOKEN is granted read/write repository permission, even when it is triggered from a public fork.

and this doesn't apply because it's about **base ref** https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions#responding-to-events
> For workflows initiated by Dependabot (github.actor == 'dependabot[bot]') using the pull_request_target event, if the base ref of the pull request was created by Dependabot (github.actor == 'dependabot[bot]'), the GITHUB_TOKEN will be read-only and secrets are not available.

PR #1713, opened 2024-04-30
CI: fix command injection plus other changes

pro-wh:
* command injection through PR title


pro-wh:
oh it's deleted on main

ptrus:
> oh it's deleted on main

Avtually might need to bring it back, will apply these changes if I do. Thanks!

pro-wh:
thanks, yeah if it's needed

PR #1742, opened 2024-05-07
runtime-sdk/modules/core: Add internal calls for call data key and epoch

jberci:


runtime-sdk/src/modules/core/mod.rs
```diff
@@ -431,6 +431,9 @@ const CONTEXT_KEY_PRIORITY: &str = "core.Priority";
 const CONTEXT_KEY_SENDER_META: &str = "core.SenderMeta";
 const CONTEXT_KEY_EPOCH_CHANGED: &str = "core.EpochChanged";
 
+const CALL_CALLDATA_PUBLIC_KEY_GAS: u64 = 100;
+const CALL_CURRENT_EPOCH_GAS: u64 = 10;
```
kostko, line 435: Move these into `GasCosts` in parameters.

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -839,20 +861,27 @@ impl<Cfg: Config> Module<Cfg> {
...
+    }
+
+    /// Retrieve the current epoch.
+    #[handler(call = "core.CurrentEpoch", internal)]
+    fn internal_call_current_epoch<C: Context>(
+        ctx: &C,
+        _args: (),
+    ) -> Result<types::CurrentEpochCallResponse, Error> {
+        <C::Runtime as Runtime>::Core::use_tx_gas(CALL_CURRENT_EPOCH_GAS)?;
+        Ok(types::CurrentEpochCallResponse { epoch: ctx.epoch() })
```
kostko, line 884: I think it would be easier (e.g. for parsing in Solidity) if it just returned an `u64` instead of a map.

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -839,20 +864,24 @@ impl<Cfg: Config> Module<Cfg> {
...
+        ctx: &C,
+        _args: (),
+    ) -> Result<types::CallDataPublicKeyQueryResponse, Error> {
+        <C::Runtime as Runtime>::Core::use_tx_gas(Cfg::GAS_COST_CALL_CALLDATA_PUBLIC_KEY)?;
+        Self::calldata_public_key_common(ctx)
+    }
+
+    /// Retrieve the current epoch.
+    #[handler(call = "core.CurrentEpoch", internal)]
+    fn internal_call_current_epoch<C: Context>(ctx: &C, _args: ()) -> Result<u64, Error> {
```
kostko, line 882:
```suggestion
    fn internal_current_epoch<C: Context>(ctx: &C, _args: ()) -> Result<u64, Error> {
```

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -839,20 +864,24 @@ impl<Cfg: Config> Module<Cfg> {
...
-                }
-                _ => Error::Abort(err.into()),
-            })?;
+        Self::calldata_public_key_common(ctx)
+    }
 
-        Ok(types::CallDataPublicKeyQueryResponse { public_key, epoch })
+    /// Retrieve the public key for encrypting call data (internally exposed call).
+    #[handler(call = "core.CallDataPublicKey", internal)]
+    fn internal_call_calldata_public_key<C: Context>(
```
kostko, line 872:
```suggestion
    fn internal_calldata_public_key<C: Context>(
```

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -581,6 +587,25 @@ impl<Cfg: Config> Module<Cfg> {
     type Parameters = Parameters;
     type Genesis = Genesis;
 
+    fn calldata_public_key_common<C: Context>(
```
kostko, line 590: I would move this down to be near calldata public key method implementations.

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -420,6 +420,12 @@ pub trait Config: 'static {
     ///
     /// Note that execution of such transactions is allowed to access confidential state.
     const ALLOW_INTERACTIVE_READ_ONLY_TRANSACTIONS: bool = false;
+
+    /// The gas cost of the internal call to retrieve the current calldata public key.
+    const GAS_COST_CALL_CALLDATA_PUBLIC_KEY: u64 = 100;
+
```
kostko, line 426:
```suggestion
```

kostko (commented):
Some minor nits, otherwise looks good to me.

PR #1797, opened 2024-05-29
rename pontusx_test to pontusx

Andrew7234:
As discussed on slack, the `_test` is indicative of the paratime being on testnet, and `pontusx` alone might be preferable

### DIFF
--- a/client-sdk/go/config/default.go
+++ b/client-sdk/go/config/default.go
@@ -120,8 +120,8 @@ var DefaultNetworks = Networks{
 						ConsensusDenomination: "TEST",
 					},
 
-					// Pontus-X Testnet on Testnet.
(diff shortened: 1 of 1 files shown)


kostko (approved):

Andrew7234:
Hm I'd still prefer to get this merged; the Nexus issues have been resolved and this PR will allow Nexus to support a mainnet pontusx runtime with a simple config change when that time comes.

kostko (changes requested):
I don't think we should do this. Also the CLI 0.9.0 now already uses `pontusx_{test,dev}`.

The main reason is that having it explicit forces one to both be aware that there are multiple pontusx variants and to think about which paratime should actually be selected.

PR #1836, opened 2024-06-20
ts-web: removed unnecessary async from hash function

CedarMist:
The `hash.ts:hash` function is unnecessarily async, this has a knock-on effect making many simple functions also async where really they're deterministic.

This change means I can derive addresses in a `constructor`, such as:

```typescript
export class OasisSigner {
  public readonly signer: oasis.signature.Signer;
  public readonly address: Uint8Array;
  public readonly addressBech32: string;

  constructor(public readonly client: OasisClient, public readonly secret: Uint8Array) {
    this.signer = oasis.signature.NaclSigner.fromSecret(secret, 'this key is not important');
    this.address = oasis.staking.addressFromPublicKey(this.signer.public())
    this.addressBech32 = oasis.staking.addressToBech32(this.address);
  }
...
}
```

You can't have async code in constructors, so the alternative would be an async factory that passes the pre-calculated addresses as args to the constructor which is annoying.

The following functions can have their async nature removed due to this change:

* address.fromData
* common.openSignedEntity
* common.openMultiSignedNode
* consensus.openSignedTransaction
* consensus.hashSignedTransaction
* genesis.chainContext
* signature.prepareSignerMessage
* signature.verify
* signature.openSigned
* signature.openMultiSigned
* staking.addressFromPublicKey
* staking.addressFromRuntimeID
* staking.commonPoolAddress
* staking.feeAccumulatorAddress
* staking.governanceDepositsAddress

### DIFF
--- a/client-sdk/ts-web/core/src/address.ts
+++ b/client-sdk/ts-web/core/src/address.ts
@@ -3,18 +3,11 @@ import {bech32} from 'bech32';
 import * as hash from './hash';
 import * as misc from './misc';
 
-export async function fromData(
-    contextIdentifier: string,
-    contextVersion: number,
-    data: Uint8Array,
-) {
+export function fromData(contextIdentifier: string, contextVersion: number, data: Uint8Array) {
     const versionU8 = new Uint8Array([contextVersion]);
     return misc.concat(
         versionU8,
-        (await hash.hash(misc.concat(misc.fromString(contextIdentifier), versionU8, data))).slice(
-            0,
-            20,
-        ),
+        hash.hash(misc.concat(misc.fromString(contextIdentifier), versionU8, data)).slice(0, 20),
     );
 }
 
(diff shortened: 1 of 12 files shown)


pro-wh:
the original design for this API was to allow a webcrypto implementation. it was never due to implementations making it necessary/unnecessary. the polyfill is incidentally synchronous.

but it's been a long time, and there's no progress toward webcrypto supporting sha-512/256. I see this proposal as "let's take that out of consideration and adjust the design." which naturally means let's make it sync for something that we can easily do sync. which seems fine

pro-wh (commented):
can you adjust the rt library too

CedarMist:
Oh, `window.crypto.subtle.digest` is async! wtf

pro-wh (approved):
(after deleting zero occurrences of subtle) goodbye subtle API!

kostko (approved):

PR #1850, opened 2024-07-02
Add method access control module

jberci:


runtime-sdk/src/modules/access/test.rs
```diff
@@ -0,0 +1,142 @@
...
+                gas: 1000,
+                consensus_messages: 0,
+            },
+            ..Default::default()
+        },
+    };
+    let call = tx.call.clone();
+
+    // Try to transfer.
+    CurrentState::with_transaction_opts(Options::new().with_tx(tx.into()), || {
```
kostko, line 116: Can we instead go through the dispatcher using the mock signer as this will test the usual dispatch flow. See e.g. [accounts tests](https://github.com/oasisprotocol/oasis-sdk/blob/dfa1d9bbe7a60ac088926a60d97488f83883a071/runtime-sdk/src/modules/accounts/test.rs#L1287) for details.

runtime-sdk/src/modules/access/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+    const NAME: &'static str = MODULE_NAME;
+    const VERSION: u32 = 1;
+    type Error = Error;
+    type Event = ();
+    type Parameters = ();
+    type Genesis = ();
+}
+
+impl<Cfg: Config> module::TransactionHandler for Module<Cfg> {
+    fn before_handle_call<C: Context>(
```
kostko, line 54: Using encrypted calldata can bypass this restriction. Please add a test for this.

runtime-sdk/src/dispatcher.rs
```diff
@@ -250,6 +250,10 @@ impl<R: Runtime> Dispatcher<R> {
             }
         }
 
+        if let Err(e) = R::Modules::before_authorized_call_dispatch(ctx, &call) {
+            return (e.into_call_result(), callformat::Metadata::Empty);
```
kostko, line 254:
```suggestion
            return (e.into_call_result(), call_format_metadata);
```

runtime-sdk/src/modules/access/mod.rs
```diff
@@ -0,0 +1,76 @@
+//! Method access control module.
+
```
kostko, line 2:
```suggestion
```

runtime-sdk/src/modules/access/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+    type Event = ();
+    type Parameters = ();
+    type Genesis = ();
+}
+
+impl<Cfg: Config> module::TransactionHandler for Module<Cfg> {
+    /*fn before_handle_call<C: Context>(
+        _ctx: &C,
+        call: &transaction::Call,
+    ) -> Result<(), modules::core::Error> {*/
```
kostko, line 57:
```suggestion
```

runtime-sdk/src/modules/access/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+    fn before_authorized_call_dispatch<C: Context>(
+        _ctx: &C,
+        call: &transaction::Call,
+    ) -> Result<(), modules::core::Error> {
+        let tx_caller_address = CurrentState::with_env(|env| env.tx_caller_address());
+        #[allow(clippy::borrow_interior_mutable_const)]
+        if Cfg::METHOD_AUTHORIZATIONS.is_authorized(&call.method, &tx_caller_address) {
+            Ok(())
+        } else {
+            Err(modules::core::Error::MalformedTransaction(
```
kostko, line 67: I would rather use `InvalidArgument` here.

runtime-sdk/src/modules/access/types.rs
kostko, line 1: Add some unit tests for these types.

runtime-sdk/src/testing/mock.rs
```diff
@@ -179,6 +180,8 @@ pub fn transaction() -> transaction::Transaction {
 pub struct CallOptions {
     /// Transaction fee.
     pub fee: transaction::Fee,
+    /// Plain of encrypted call.
```
kostko, line 183:
```suggestion
    /// Should the call be encrypted.
```

runtime-sdk/src/testing/mock.rs
```diff
@@ -232,14 +236,45 @@ impl Signer {
...
     {
+        let inner_call = transaction::Call {
+            format: transaction::CallFormat::Plain,
+            method: method.to_owned(),
+            body: cbor::to_value(body),
+            ..Default::default()
+        };
+        let call = if !opts.encrypted {
+            inner_call
+        } else {
```
kostko, line 247: It doesn't look like this is transparently handling encrypted results?

runtime-sdk/src/testing/mock.rs
```diff
@@ -232,14 +236,45 @@ impl Signer {
         C: Context,
         B: cbor::Encode,
     {
+        let inner_call = transaction::Call {
```
peternose, line 239: No need to create `inner_call`, if you move inside `if` block.

runtime-sdk/src/dispatcher.rs
```diff
@@ -250,6 +250,10 @@ impl<R: Runtime> Dispatcher<R> {
             }
         }
 
+        if let Err(e) = R::Modules::before_authorized_call_dispatch(ctx, &call) {
```
peternose, line 253: It looks like we have 2 method authorizations, one here and one few lines above (`method_authorizer`). A bit confusing. Can this be merged?

runtime-sdk/src/modules/access/types.rs
```diff
@@ -0,0 +1,72 @@
...
+pub enum Authorization {
+    /// Control a statically configured set of methods, each with a
+    /// statically configured set of addresses that are allowed to call it.
+    FilterOnly(Methods),
+}
+
+impl Authorization {
+    /// Return a new access control configuration.
+    pub fn new() -> Self {
+        Self::FilterOnly(BTreeMap::new())
```
peternose, line 43: Use default.

runtime-sdk/src/dispatcher.rs
kostko: I don't think so. One is for caller-specified dispatch options, the other is a hook for that runtime modules can implement. So they serve two different purposes and we need both.

runtime-sdk/src/testing/mock.rs
jberci: It doesn't, but it calls `Dispatch::dispatch_tx`, which doesn't automatically encrypt the result object. I added a return value to both runtime methods in the access module's tests to check.

runtime-sdk/src/testing/mock.rs
jberci: The value of `inner_call` is used in both branches of the `if`, in one it's wrapped and in the other it's just moved to a new variable, I wanted to avoid repeating the constructor. I restructured the if slightly, maybe it's a bit more readable now.

kostko (approved):

peternose (approved):

PR #1884, opened 2024-07-15
runtime-sdk/modules/rofl: Reduce coupling between ROFL and RONL

kostko:
The previous solution required the same SDK versions for ROFL and RONL which seems like an artificial constraint that could get annoying in practice. This should reduce developer friction.

### DIFF
--- a/runtime-sdk/src/modules/rofl/app/mod.rs
+++ b/runtime-sdk/src/modules/rofl/app/mod.rs
@@ -7,14 +7,14 @@ use tokio::sync::mpsc;
 
 use crate::{
     core::{
+        common::version,
         config::Config,
-        consensus::roothash,
+        consensus::{roothash, verifier::TrustRoot},
         dispatcher::{PostInitState, PreInitState},
         rofl, start_runtime,
     },
     crypto,
     types::transaction,
-    Runtime,
 };
 
 mod client;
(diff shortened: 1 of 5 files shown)


runtime-sdk/src/modules/rofl/app/mod.rs
```diff
@@ -31,14 +31,18 @@ pub use env::Environment;
...
-    /// The runtime must have the ROFL module enabled so that the application can register to it.
-    type AttachTo: Runtime;
+    /// ROFL application version.
+    const VERSION: version::Version;
 
     /// Identifier of the application (used for registrations).
     fn id() -> AppId;
 
+    /// Return the consensus layer trust root for this runtime; if `None`, consensus layer integrity
+    /// verification will not be performed.
```
peternose, line 41: This looks dangerous. Should we rather force apps to implement this, even if they only have to return `None`, so that they explicitly ignore verification?

peternose (approved):

runtime-sdk/src/modules/rofl/app/mod.rs
kostko: Good point, let me fix this, the PR already got merged.

peternose: Sorry, didn't see that you have auto-merge enabled. If needed, we can change this in another PR.

kostko: #1885

PR #1906, opened 2024-07-24
Multiple gas and fee proxy related fixes

kostko:


runtime-sdk/src/modules/rofl/app/client.rs
```diff
@@ -195,6 +195,11 @@ where
                     propagate_failures: false,
                 })
                 .await?;
+
+            // The estimate may be off due to current limitations in confidential gas estimation.
+            // Inflate the estimated gas by 20%.
+            let gas = gas.saturating_mul(120).saturating_div(100);
```
peternose, line 201:
```suggestion
            let gas = gas.saturating_add(gas.saturating_mul(20).saturating_div(100));
```
In theory, this could result in a lower gas.

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -603,7 +603,7 @@ impl<Cfg: Config> Module<Cfg> {
         ctx: &C,
         mut args: types::EstimateGasQuery,
     ) -> Result<u64, Error> {
-        let mut extra_gas = 0;
+        let mut extra_gas = 0u64;
```
peternose, line 606: This change is not needed, I think.
This line could be moved lower.

kostko: It is needed because the compiler needs type annotations to resolve the right methods.

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -1040,6 +1043,32 @@ impl<Cfg: Config> Module<Cfg> {
 
         Ok(())
     }
+
+    fn compute_signature_verification_cost(signer_info: &[SignerInfo]) -> Option<u64> {
+        let params = Self::params();
```
kostko, line 1048: Ah this needs to be moved to an argument due to nested env.

runtime-sdk/src/modules/rofl/app/client.rs
kostko: Significantly? The 20% is a rough estimate anyway so it shouldn't matter if it is not exact.

peternose: Significantly yes, as in theory this value could reduce for a factor of 100, e.g., if `gas = maxu64`, then new gas is `gas = maxu64/100`. In practice probably not possible.

PR #1934, opened 2024-08-20
runtime-sdk/modules/rofl: Add YAML serialization for ROFL fee policy

kostko:


client-sdk/go/modules/rofl/policy.go
```diff
@@ -39,8 +43,46 @@ type AllowedEndorsement struct {
...
+	// FeePolicyInstancePays is a fee policy where the application enclave pays the gas fees.
+	FeePolicyInstancePays FeePolicy = 1
 	// FeePolicyEndorsingNodePays is a fee policy where the endorsing node pays the gas fees.
 	FeePolicyEndorsingNodePays FeePolicy = 2
 )
+
+// UnmarshalYAML implements yaml.Unmarshaler.
+func (fp *FeePolicy) UnmarshalYAML(value *yaml.Node) error {
+	switch value.ShortTag() {
+	case "!!str":
```
peternose, line 55: Why do we need to support this case?  Since encoding is not unique, one might mix different encodings :thinking: 

```yaml
fee_policy1: instance
fee_policy2: 1
```

peternose (approved):

client-sdk/go/modules/rofl/policy.go
kostko: Just for compatibility. Maybe no need to support it.

kostko: Fixed.

peternose (approved):

PR #1946, opened 2024-08-23
rofl-utils: Get rid of ureq fork, bring mbedtls support locally

kostko:
This removes the need to maintain an ureq fork and brings in mbedtls support into the `rofl-utils` crate (also see discussion in https://github.com/algesten/ureq/pull/791). There is some code duplication with `ureq` TLS, but on the other hand we have our own connectors/transports for SGX anyway.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -199,7 +199,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "726535892e8eae7e70657b4c8ea93d26b8553afb1ce617caee529ef96d7dee6c"
 dependencies = [
  "proc-macro2 1.0.86",
- "quote 1.0.36",
+ "quote 1.0.37",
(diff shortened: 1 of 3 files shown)


peternose (commented):
NIce :+1: 

Do you have any idea why request to `https://www.bing.com/` fails with `Io(Custom { kind: InvalidData, error: AlertReceived(HandshakeFailure) })`?

kostko:
> Do you have any idea why request to `https://www.bing.com/` fails with `Io(Custom { kind: InvalidData, error: AlertReceived(HandshakeFailure) })`?

Tried to debug it a bit, but didn't find the cause. Likely some suite is not supported?

peternose (approved):
> Likely some suite is not supported?

Probably. We can debug latter if this becomes a problem for our developers.

PR #1961, opened 2024-08-29
runtime-sdk: Disallow zero transfers of unknown denoms

kostko:


pro-wh (commented):
the details of this change:

1. if you try to transfer zero from an account that doesn't even have an entry for that denomination, it fails with insufficient balance
2. except for if the above is done from EVM, in which case it succeeds instead
3. the above does not emit a transfer event
4. if you try to transfer zero from an account that does have an entry for that denomination, it succeeds
5. the above does not create the entry in the recipient account
6. the above does not emit a transfer event
7. transferring the last of a denomination does not remove the entry for that denomination

question: would this interfere with the special zero-fee transactions like deposit? in those cases you have no fee denomination in your account.

kostko:
> question: would this interfere with the special zero-fee transactions like deposit?

Possibly. I think this introduces many edge cases that need to be worked around so I'm not convinced if it is really worth the effort? Frontends could just be more careful in displaying denominations?

pro-wh:
ya maybe allowing zero transfers is normal enough. I think ERC-20 doesn't forbid 0 transfers, and they should still emit a Transfer event. and a lot of the system has the type of assumptions that an empty account is equivalent to having 0 of a given token. lotta things to be consistent with

PR #1982, opened 2024-09-12
ROFL instructions for Localnet

matevz:


docs/rofl/prerequisites.md
```diff
@@ -77,13 +77,28 @@ nightly-2022-08-22-x86_64-unknown-linux-gnu (overridden by '/code/rust-toolchain
 rustc 1.65.0-nightly (c0941dfb5 2022-08-21)
 ```
 
+ROFL binaries should be compiled for [MUSL C standard library]. You will need
```
kostko, line 80: I'm not sure why they should be compiled for musl, because this can actually cause some problems. We could mention that they need to be compiled for musl in case of Localnet, but outside that it doesn't make sense to require musl?

examples/runtime-sdk/rofl-oracle/src/main.rs
```diff
@@ -90,5 +93,6 @@ impl OracleApp {
 }
 
 fn main() {
+    println!("ROFL oracle app loaded!");
```
kostko, line 96: Drop this.

examples/runtime-sdk/rofl-oracle/src/main.rs
```diff
@@ -50,17 +53,17 @@ impl OracleApp {
         let observation = tokio::task::spawn_blocking(move || -> Result<_> {
             // Request some data from Coingecko API.
             let rsp: serde_json::Value = rofl_utils::https::agent()
-                .get("https://api.coingecko.com/api/v3/simple/price?ids=oasis-network&vs_currencies=USD")
+                .get("https://www.binance.com/api/v3/ticker/price?symbol=ROSEUSDT")
```
kostko, line 56: Why this change?

examples/runtime-sdk/rofl-oracle/src/main.rs
```diff
@@ -12,21 +12,24 @@ impl App for OracleApp {
...
+//     fn consensus_trust_root() -> Option<TrustRoot> {
+//         // The trust root below is for Sapphire Testnet at consensus height 22110615.
+//         Some(TrustRoot {
+//             height: 22110615,
+//             hash: "95d1501f9cb88619050a5b422270929164ce739c5d803ed9500285b3b040985e".into(),
+//             runtime_id: "000000000000000000000000000000000000000000000000a6d1e3ebf60dff6c".into(),
+//             chain_context: "0b91b8e4e44b2003a7c5e23ddadb5e14ef5345c0ebcb3ddcae07fa2f244cab76"
+//                 .to_string(),
+//         })
+//     }
```
kostko, line 32: I am not convinced this should be the default as then people will omit the trust root.

kostko (commented):
The "How do ROFL Apps Work" currently has this text:
> For the rest of this chapter we will be using Sapphire Testnet which has all of the required functionality.

This should be changed to somehow say that we will be deploying either on Localnet or on Sapphire Testnet. An alternative would be to completely separate and have two chapters:
* Deploying an UNSAFE debug version on Localnet.
* Deploying in production to Testnet and Mainnet.

I think the current mixing of the two is somewhat confusing.

examples/runtime-sdk/rofl-oracle/src/main.rs
matevz: Coingecko has a very low rate limit for the free/demo plan (30 requests/min) which doesn't work long for the Localnet's 1s block time. Binance does not have such a limit for getting price quotes.

examples/runtime-sdk/rofl-oracle/src/main.rs
matevz: CLI defining the trust root would work best. Could Oasis CLI also verify that the ROFL .orc is ready to roll once it's compiled (e.g. check if the trust root is not None, which app ID is defined etc.)?

examples/runtime-sdk/rofl-oracle/src/main.rs
kostko: We could also add a rate limit in the demo, there is no need to query so often.

docs/rofl/prerequisites.md
matevz: For Testnet and Mainnet production builds the build target is `x86_64-fortanix-unknown-sgx` right? I don't see other environments developers would want to run their ROFLs?

matevz: Agree, I'll change the wording ;)

kostko: Yeah so the issue is that for Mainnet and Testnet the ELF binary doesn't really matter and requiring MUSL for those doesn't make sense. The orcs currently must include the ELF binaries as well -- this will change in Oasis Core 24.3 where ELF binaries will become optional and not having them will automatically force the use of SGX/TDX.

examples/runtime-sdk/rofl-oracle/src/main.rs
matevz: Switched back.

docs/rofl/deployment.md
```diff
@@ -0,0 +1,365 @@
...
+This way, your ROFL client will sync more quickly and not want to start on any
+other network or ParaTime. Read the [Consensus Trust Root] chapter to learn more
+about obtaining a correct block for the root of trust.
+
+[Consensus Trust Root]: trust-root.md
+
+## Register the App
+
+Before the ROFL app can authenticate it needs to be registered as an app on the
+network. Anyone with enough stake can register an app by using the CLI.
```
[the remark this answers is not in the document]
matevz: This is actually ParaTime related, so we could have it set differently on Sapphire and Pontus-x for example both running on Testnet. Currently it's 10k on Testnet and Localnet Sapphire. We don't have ROFLs enabled on Mainnet Sapphire yet. I'll mention this in the paragraph.

FYI I opened https://github.com/oasisprotocol/oasis-sdk/issues/1993 and https://github.com/oasisprotocol/cli/issues/290 for developers to discover the thresholds for themselves.

examples/runtime-sdk/rofl-oracle/src/main.rs
matevz: FYI opened https://github.com/oasisprotocol/cli/issues/291 to address this.

abukosek (approved):

matevz:
Let's merge this so we have a documented ROFL Localnet usage and open subsequent followup PRs.

ptrus (approved):

PR #1996, opened 2024-09-20
runtime-sdk/src/modules/core: add "core.PublicKey" call

CedarMist:
Provides an easy way to retrieve both the KeypairId and the Keypair which sign the CallDataPublicKey.

These, combined with the runtime ID, should be enough to validate the CallDataPublicKey signature.

Would probably be useful to include the runtime ID too, am leaving this in draft until I can get e2e tests setup.


CedarMist:
So, ran into some problems.

> So secrets::get_public_ephemeral_key uses `kdf.sign_pubic_key`, which signs the key using `kdf.inner.signer`
> But... I can't get the inner.signer public key through the keymanager API, and it's not the long term key pair (which I assumed had been signing the ephemeral keypairs) that I need - it's the signer
> `signer` is derived from derive_signing_key with the master secret & runtime ID, and the `ekiden-derive-signing-key` namespace, whereas long-term keys are derived using `ekiden-derive-runtime-secret` namespace via `get_or_create_longterm_keys`

CedarMist:
Relies on https://github.com/oasisprotocol/oasis-core/pull/5865 being merged into oasis-core, so we can use it in new oasis-sdk.

runtime-sdk/src/modules/core/types.rs
```diff
@@ -41,6 +45,17 @@ pub struct CallDataPublicKeyQueryResponse {
...
 }
 
+/// Response to the public key query.
+#[derive(Clone, Debug, Default, cbor::Encode, cbor::Decode)]
+pub struct KeyManagerPublicKeyQueryResponse {
+    /// ID of the public key which signs the call data public keys
+    pub key_pair_id: KeyPairId,
+    /// Runtime signing key which signs the call data public keys
+    pub public_key: PublicKey,
+    /// For reference, which runtime ID is this for?
```
peternose, line 55: When querying the key manager for (long or short term) keys you need to specify the runtime id (e.g. Sapphire) so that the key manager node can verify if the enclave, which is making the call, is allowed to query keys for the given runtime.

runtime-sdk/src/modules/core/types.rs
```diff
@@ -41,6 +45,17 @@ pub struct CallDataPublicKeyQueryResponse {
     pub epoch: u64,
 }
 
+/// Response to the public key query.
+#[derive(Clone, Debug, Default, cbor::Encode, cbor::Decode)]
+pub struct KeyManagerPublicKeyQueryResponse {
```
peternose, line 50: Response to the public key query should probably only contain the public key. The request should contain the other two fields. So the name is a bit misleading (and long) and could be changed.

runtime-sdk/src/modules/core/mod.rs
```diff
@@ -846,6 +848,24 @@ impl<Cfg: Config> Module<Cfg> {
         <C::Runtime as Runtime>::Modules::check_invariants(ctx)
     }
 
+    fn keymanager_public_key_common<C: Context>(
+        ctx: &C,
+    ) -> Result<types::KeyManagerPublicKeyQueryResponse, Error> {
+        let key_manager = ctx
+            .key_manager()
+            .ok_or_else(|| Error::InvalidArgument(anyhow!("key manager not available")))?;
+        let runtime_id = key_manager.runtime_id().ok_or_else(|| Error::InvalidArgument(anyhow!("cannot get runtime_id")))?;
```
peternose, line 857: Is this line needed?

runtime-sdk/src/modules/core/types.rs
CedarMist: I'm following the same naming scheme as `CallDataPublicKeyQueryResponse`.

It needs to return `key_pair_id` to be able to verify the signed ephemeral key, as that's not known ahead of time.

The `runtime_id` should be moved to `CallDataPublicKeyQueryResponse` as it's the runtime which we're encrypting the calldata for.

PR #2010, opened 2024-09-30
Bump Oasis Core to 24.3

kostko:


runtime-sdk/src/modules/rofl/app/client.rs
```diff
@@ -185,9 +186,21 @@ where
...
             latest_round: Arc::new(AtomicU64::new(0)),
+            rpc: Arc::new(RpcClient::new_runtime(
+                state.host.clone(),
+                ENCLAVE_RPC_ENDPOINT_RONL,
+                session::Builder::default()
+                    .use_endorsement(true)
+                    .quote_policy(None)
+                    .local_identity(state.identity.clone())
+                    .remote_enclaves(None),
+                8,
```
ptrus, line 199: A comment explaining the chosen session parameters would be nice.

ptrus (approved):

runtime-sdk/src/modules/rofl/app/client.rs
kostko: Adjusted and added a comment.

PR #2013, opened 2024-10-01
client-sdk: allow insecure connections to localhost address

ptrus:
This usecase came up in https://github.com/oasisprotocol/oasis-web3-gateway/pull/626 since we want to expose the node from docker via a port (since this works better cross-platform than exposing the socket via a volume).

The alternative possible solution would be to provide a flag that allows insecure connections or connections with self-signed certs (something like --insecure in curl).

BTW we might want to do the same in oasis-core: https://github.com/oasisprotocol/oasis-core/blob/280d2ef1947bc9baaeb390c3131a08b87824c9ef/go/oasis-node/cmd/common/grpc/grpc.go#L83-L89
since ideally, both `oasis-node` and `oasis-cli` would work with local-net without much hassle.

cc @matevz 

TODO: 
- [x] use the method from oasis-core: https://github.com/oasisprotocol/oasis-core/compare/ptrus/feature/grpc-loopback-insecure?expand=1

### DIFF
--- a/client-sdk/go/config/network.go
+++ b/client-sdk/go/config/network.go
@@ -3,7 +3,6 @@ package config
 import (
 	"fmt"
 	"net/url"
-	"strings"
 
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/hash"
 )
@@ -135,9 +134,4 @@ func (n *Network) Validate() error {
 	}
 
 	return nil
-}
-
-// IsLocalRPC checks whether the RPC endpoint points to a local UNIX socket.
-func (n *Network) IsLocalRPC() bool {
-	return strings.HasPrefix(n.RPC, "unix:")
 }
--- a/client-sdk/go/connection/connection.go
+++ b/client-sdk/go/connection/connection.go
@@ -105,7 +105,7 @@ func Connect(ctx context.Context, net *config.Network) (Connection, error) {
 // omitting the chain context check.
 func ConnectNoVerify(_ context.Context, net *config.Network) (Connection, error) {
 	var dialOpts []grpc.DialOption
-	switch net.IsLocalRPC() {
+	switch cmnGrpc.IsLocalAddress(net.RPC) {
 	case true:
 		// No TLS needed for local nodes.
 		dialOpts = append(dialOpts, grpc.WithTransportCredentials(insecure.NewCredentials()))


client-sdk/go/config/network.go
```diff
@@ -137,7 +138,17 @@ func (n *Network) Validate() error {
...
+		return true
+	}
+
+	// RPC addresses are without scheme.
+	host, _, err := net.SplitHostPort(n.RPC)
+	if err != nil {
+		return false
+	}
+
+	return net.ParseIP(host).IsLoopback() || host == "localhost"
```
matevz, line 153: You may also have virutal hosts or non-standard names for 127.0.0.1 right?

```suggestion
	ip, err := net.LookupIP(host)
	if err != nil || ip.length==0 {
		return false
	}
	return ip.IsLoopback()
```

kostko (approved):

ptrus:
Please check again, I updated it a bit to be in line with the gRPC address spec defined at: https://github.com/grpc/grpc/blob/master/doc/naming.md

kostko (approved):

matevz (approved):
I can confirm Oasis CLI using this patch works fine with Envoy running on `emerald-localnet` and `sapphire-localnet`.

ptrus:
Updated code once again, now it reuses the method from oasis-core (added in: https://github.com/oasisprotocol/oasis-core/pull/5878)

kostko (commented):
Looks good, so we need to wait until a new Oasis Core gets released?

ptrus:
Yes, but if we don't plan to release it anytime soon, then we could copy the method for now. 

But we will need to release the updated `oasis-sdk` and `oasis-cli` anyway if we want a working `cli` with the localnet container, so we could wait for `oasis-core` as well if there are any plans for it in the near future.

@matevz  not sure how urgently we want this for different components
- For `oasis-cli` to work with localnet  - we need  `oasis-sdk` and `oasis-cli` releases (and potentially `oasis-core` unless we duplicate the code here)
- For `oasis-node` to work with localnet -  we need `oasis-core` release

matevz:
Sure, we can wait for the point release of the `oasis-core` and keep a clean code here.

kostko:
Oasis Core 24.3 has been released, please rebase, then we can tag the Go Client SDK.

PR #2034, opened 2024-10-17
Add support for TDX

kostko:


tools/orc/main.go
```diff
@@ -426,12 +426,26 @@ func showComponent(bnd *bundle.Bundle, comp *bundle.Component, legacy bool) {
...
 	indent := "  "
 
-	fmt.Printf("%sExecutable:     %s\n", indent, comp.Executable)
+	if comp.Executable != "" {
+		fmt.Printf("%sExecutable:     %s\n", indent, comp.Executable)
+	}
+
+	fmt.Printf("%sTEE kind:       %s\n", indent, comp.TEEKind())
+
+	switch {
```
peternose, line 435: This `switch` could be removed since checks are already done in helper functions. Updated code would (in theory) allow having a component with SGX and TDX.

peternose (approved):

PR #2040, opened 2024-10-22
runtime-sdk: Add rofl.StakeThresholds query

### LINKED ISSUE #1993: Add ROFL minimum staking requirements query
Add the minimum staking requirements (currently set to 10k on Testnet) query for registering a new ROFL so the Oasis CLI and other apps can show it.

Related https://github.com/oasisprotocol/cli/issues/290

matevz:
This PR adds `rofl.StakeThresholds` query to runtime-sdk and client-sdk/go.

Fixes #1993

### DIFF
--- a/client-sdk/go/modules/rofl/rofl.go
+++ b/client-sdk/go/modules/rofl/rofl.go
@@ -18,10 +18,11 @@ var (
 	methodRegister = types.NewMethodName("rofl.Register", Register{})
 
 	// Queries.
-	methodApp          = types.NewMethodName("rofl.App", AppQuery{})
-	methodAppInstance  = types.NewMethodName("rofl.AppInstance", AppInstanceQuery{})
-	methodAppInstances = types.NewMethodName("rofl.AppInstances", AppQuery{})
-	methodParameters   = types.NewMethodName("rofl.Parameters", nil)
+	methodApp             = types.NewMethodName("rofl.App", AppQuery{})
+	methodAppInstance     = types.NewMethodName("rofl.AppInstance", AppInstanceQuery{})
+	methodAppInstances    = types.NewMethodName("rofl.AppInstances", AppQuery{})
+	methodParameters      = types.NewMethodName("rofl.Parameters", nil)
+	methodStakeThresholds = types.NewMethodName("rofl.StakeThresholds", nil)
 )
 
 // V1 is the v1 rofl module interface.
@@ -45,6 +46,9 @@ type V1 interface {
 
 	// AppInstances queries the registered instances of the given application.
 	AppInstances(ctx context.Context, round uint64, id AppID) ([]*Registration, error)
+
+	// StakeThresholds queries the stake information for managing ROFL.
+	StakeThresholds(ctx context.Context, round uint64) (*StakeThresholds, error)
 
 	// Parameters queries the module parameters.
 	Parameters(ctx context.Context, round uint64) (*Parameters, error)
@@ -108,6 +112,16 @@ func (a *v1) AppInstances(ctx context.Context, round uint64, id AppID) ([]*Regis
 		return nil, err
 	}
 	return instances, nil
+}
+
+// Implements V1.
+func (a *v1) StakeThresholds(ctx context.Context, round uint64) (*StakeThresholds, error) {
+	var thresholds StakeThresholds
+	err := a.rc.Query(ctx, round, methodStakeThresholds, nil, &thresholds)
+	if err != nil {
+		return nil, err
+	}
+	return &thresholds, nil
 }
 
 // Implements V1.
--- a/client-sdk/go/modules/rofl/types.go
+++ b/client-sdk/go/modules/rofl/types.go
@@ -136,3 +136,8 @@ type Event struct {
 	AppUpdated *AppUpdatedEvent
 	AppRemoved *AppRemovedEvent
 }
+
+// StakeThresholds contains staking thresholds for managing ROFL.
+type StakeThresholds struct {
+	AppCreate *types.BaseUnits `json:"app_create"`
+}
--- a/runtime-sdk/src/modules/rofl/config.rs
+++ b/runtime-sdk/src/modules/rofl/config.rs
@@ -16,6 +16,8 @@ pub trait Config: 'static {
     const GAS_COST_CALL_AUTHORIZED_ORIGIN_NODE: u64 = 2000;
     /// Gas cost of rofl.AuthorizedOriginEntity call.
     const GAS_COST_CALL_AUTHORIZED_ORIGIN_ENTITY: u64 = 2000;
+    /// Gas cost of rofl.StakeThresholds call.
+    const GAS_COST_CALL_STAKE_THRESHOLDS: u64 = 10;
 
     /// Amount of stake required for maintaining an application.
     ///
--- a/runtime-sdk/src/modules/rofl/mod.rs
+++ b/runtime-sdk/src/modules/rofl/mod.rs
@@ -492,6 +492,29 @@ impl<Cfg: Config> Module<Cfg> {
         Self::get_instances(args.id)
     }
 
+    /// Returns the minimum stake thresholds for managing ROFL.
+    #[handler(query = "rofl.StakeThresholds")]
+    fn query_stake_thresholds<C: Context>(
+        _ctx: &C,
+        _args: (),
+    ) -> Result<types::StakeThresholds, Error> {
+        Ok(types::StakeThresholds {
+            app_create: Cfg::STAKE_APP_CREATE,
+        })
+    }
+
+    /// Returns the minimum stake thresholds for managing ROFL.
+    #[handler(call = "rofl.StakeThresholds", internal)]
+    fn internal_query_stake_thresholds<C: Context>(
+        ctx: &C,
+        _args: (),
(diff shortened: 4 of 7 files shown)


client-sdk/go/modules/rofl/rofl.go
```diff
@@ -18,10 +18,11 @@ var (
...
 
 	// Queries.
-	methodApp          = types.NewMethodName("rofl.App", AppQuery{})
-	methodAppInstance  = types.NewMethodName("rofl.AppInstance", AppInstanceQuery{})
-	methodAppInstances = types.NewMethodName("rofl.AppInstances", AppQuery{})
-	methodParameters   = types.NewMethodName("rofl.Parameters", nil)
+	methodApp            = types.NewMethodName("rofl.App", AppQuery{})
+	methodAppInstance    = types.NewMethodName("rofl.AppInstance", AppInstanceQuery{})
+	methodAppInstances   = types.NewMethodName("rofl.AppInstances", AppQuery{})
+	methodStakeAppCreate = types.NewMethodName("rofl.StakeAppCreate", StakeAppCreate{})
```
kostko, line 24:
```suggestion
	methodStakeAppCreate = types.NewMethodName("rofl.StakeAppCreate", nil)
```

client-sdk/go/modules/rofl/rofl.go
```diff
@@ -110,6 +114,16 @@ func (a *v1) AppInstances(ctx context.Context, round uint64, id AppID) ([]*Regis
 	return instances, nil
 }
 
+// Implements V1.
+func (a *v1) StakeAppCreate(ctx context.Context, round uint64) (*types.BaseUnits, error) {
+	var bu types.BaseUnits
+	err := a.rc.Query(ctx, round, methodStakeAppCreate, StakeAppCreate{}, &bu)
```
kostko, line 120:
```suggestion
	err := a.rc.Query(ctx, round, methodStakeAppCreate, nil, &bu)
```

client-sdk/go/modules/rofl/types.go
```diff
@@ -100,6 +100,9 @@ type Registration struct {
 	ExtraKeys []types.PublicKey `json:"extra_keys"`
 }
 
+// StakeAppCreate are the stake app create for the rofl module.
+type StakeAppCreate struct{}
+
```
kostko, line 105: Remove.

runtime-sdk/src/modules/rofl/mod.rs
```diff
@@ -492,6 +492,12 @@ impl<Cfg: Config> Module<Cfg> {
         Self::get_instances(args.id)
     }
 
+    /// Returns the minimum amount of required stake to create a new application.
+    #[handler(query = "rofl.StakeAppCreate")]
+    fn query_stake_app_create<C: Context>(_ctx: &C) -> Result<token::BaseUnits, Error> {
```
kostko, line 497:
```suggestion
    fn query_stake_app_create<C: Context>(_ctx: &C, _args: ()) -> Result<token::BaseUnits, Error> {
```

kostko (commented):
An alternative would be to have a query `rofl.StakeThresholds` which would return something like:
```rust
pub struct StakeThresholds {
    pub create: token::BaseUnits,
}
```

runtime-sdk/src/modules/rofl/types.rs
```diff
@@ -109,3 +109,9 @@ pub struct AppInstanceQuery {
     /// Runtime Attestation Key.
     pub rak: PublicKey,
 }
+
+/// Minimum stake thresholds for managing ROFL application.
```
peternose, line 113:
```suggestion
/// Stake thresholds for managing ROFL application.
```

Optionally you could also generalize a bit more and rename to `StakeThresholds`, change query to `rofl.StakeThresholds` and field `create` to `app_create`.

client-sdk/go/modules/rofl/types.go
```diff
@@ -136,3 +136,8 @@ type Event struct {
 	AppUpdated *AppUpdatedEvent
 	AppRemoved *AppRemovedEvent
 }
+
+// AppStakeThresholds contains minimum stake thresholds for managing ROFL application.
+type AppStakeThresholds struct {
+	Create *types.BaseUnits
```
kostko, line 142:
```suggestion
	Create *types.BaseUnits `json:"create"`
```

runtime-sdk/src/modules/rofl/mod.rs
```diff
@@ -492,6 +492,17 @@ impl<Cfg: Config> Module<Cfg> {
         Self::get_instances(args.id)
     }
 
+    /// Returns the minimum amount of required stake for managing ROFL application.
+    #[handler(query = "rofl.AppStakeThresholds")]
```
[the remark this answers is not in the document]
kostko: > unfortunately

There are good reasons why this is not the case by default:
* Queries can be non-deterministic which could cause discrepancies and round failures.
* Queries can be expensive.
* Queries don't usually account for gas which is needed for safe operation from calls.

We could however extend the procedural macros to support an attribute that would expose a specific query as an internal call as well. But this would just save some typing and would still need to be explicit to be safe (with gas accounting handled by the query impl when run in a call context).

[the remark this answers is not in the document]
CedarMist: Yup this was just a note, and a question really about did we want to expose this to solidity

runtime-sdk/src/modules/rofl/mod.rs
```diff
@@ -492,6 +492,17 @@ impl<Cfg: Config> Module<Cfg> {
...
+    fn query_app_stake_thresholds<C: Context>(_ctx: &C, _args: ()) -> Result<types::AppStakeThresholds, Error> {
+        Ok(
+            types::AppStakeThresholds{
+                create: Cfg::STAKE_APP_CREATE,
+                ..Default::default()
```
kostko, line 501:
```suggestion
```

runtime-sdk/src/modules/rofl/mod.rs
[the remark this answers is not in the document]
matevz: I don't see a reason why not to.

tests/e2e/rofl/tests.go
```diff
@@ -105,6 +107,18 @@ func CreateUpdateRemoveTest(ctx context.Context, env *scenario.Env) error {
 func QueryTest(ctx context.Context, env *scenario.Env) error {
 	rf := rofl.NewV1(env.Client)
 
+	// Query for stake app create.
+	s, err := rf.StakeThresholds(ctx, client.RoundLatest)
+	if err != nil {
+		return err
+	}
+	if !reflect.DeepEqual(rofl.StakeThresholds{AppCreate: &types.BaseUnits{
+		Amount:       *quantity.NewFromUint64(1_000_000_000_000),
```
kostko, line 116: I believe the default is zero, so if you want something else you need to modify either this or [this](https://github.com/oasisprotocol/oasis-sdk/blob/main/tests/runtimes/components-ronl/src/lib.rs).

runtime-sdk/src/modules/rofl/mod.rs
```diff
@@ -492,6 +492,29 @@ impl<Cfg: Config> Module<Cfg> {
...
+    fn query_stake_thresholds<C: Context>(
+        _ctx: &C,
+        _args: (),
+    ) -> Result<types::StakeThresholds, Error> {
+        Ok(types::StakeThresholds {
+            app_create: Cfg::STAKE_APP_CREATE,
+        })
+    }
+
+    /// Returns the minimum stake thresholds for managing ROFL. (internally exposed call).
```
kostko, line 506:
```suggestion
    /// Returns the minimum stake thresholds for managing ROFL.
```

runtime-sdk/src/modules/rofl/config.rs
```diff
@@ -16,6 +16,8 @@ pub trait Config: 'static {
     const GAS_COST_CALL_AUTHORIZED_ORIGIN_NODE: u64 = 2000;
     /// Gas cost of rofl.AuthorizedOriginEntity call.
     const GAS_COST_CALL_AUTHORIZED_ORIGIN_ENTITY: u64 = 2000;
+    /// Gas cost of rofl.StakeThresholds call.
+    const GAS_COST_CALL_STAKE_THRESHOLDS: u64 = 100;
```
matevz, line 20: I have no idea what gas cost should we apply for fetching a constant from the config?

peternose: Probably 20? See https://github.com/oasisprotocol/oasis-sdk/blob/main/runtime-sdk/src/modules/core/mod.rs#L426-L429

Why would one pay for the gas if the value could be hard-coded?

kostko: The value can change. And yeah something small seems appropriate.

tests/e2e/rofl/tests.go
matevz: Set to 1000 for testing it.

runtime-sdk/src/modules/rofl/config.rs
matevz: Set to 10 to match `GAS_COST_CALL_CURRENT_EPOCH`.

kostko (approved):

PR #2069, opened 2024-11-19
examples: Add TDX-based oracle

kostko:


client-sdk/ts-web/rt/playground/sample-run-network.sh
```diff
@@ -38,7 +40,55 @@ jq '
...
+# Run test runner to generate genesis document.
 "$TEST_NET_RUNNER" \
     --fixture.file "$FIXTURE_FILE" \
-    --basedir /tmp/oasis-net-runner-sdk-rt \
-    --basedir.no_temp_dir
+    --basedir "$WORKDIR" \
+    --basedir.no_temp_dir &
+RUNNER_PID=$!
+
+# Below is a workaround for there being no way to change the default max tx size which
```
ptrus, line 50: Do we have an issue in oasis-core, for adding support to this vis a fixture?

ptrus (approved):

client-sdk/ts-web/rt/playground/sample-run-network.sh
kostko: Good point, I'll open an issue. Core otherwise has the increased max tx size in master.

kostko: Done, https://github.com/oasisprotocol/oasis-core/issues/5960.

ptrus: Thx!

PR #2073, opened 2024-11-20
runtime-sdk: Add sr25519 support in EVM precompiles

### LINKED ISSUE #1968: sr25519 precompile for EVM sdk
Re: #304

In [evm/src/precompile/confidential.rs](https://github.com/oasisprotocol/oasis-sdk/blob/main/runtime-sdk/modules/evm/src/precompile/confidential.rs) we support Ed25519, Secp256k1, Secp256r1 and Secp384r1.

Sr25519 is also supported in [crypto/signature/mod.rs](https://github.com/oasisprotocol/oasis-sdk/blob/main/runtime-sdk/src/crypto/signature/mod.rs)

For Polkadot compatibility, we should support sr25519 signing & verification, as all the necessary support is there.

However, it seems substrate key derivation is not compatible with our ADR-8. To make sure it's compatible with Polkadot it will be necessary to take test vectors from Polkadot ecosystem to be sure signatures, keygen & verification works correctly.

kostko:
Fixes #1968

### DIFF
--- a/runtime-sdk/modules/evm/src/precompile/confidential.rs
+++ b/runtime-sdk/modules/evm/src/precompile/confidential.rs
@@ -44,6 +44,7 @@ static KEYPAIR_GENERATE_BASE_COST: Lazy<HashMap<SignatureType, u64>> = Lazy::new
         (SignatureType::Secp256k1_PrehashedSha256, 1_500),
         (SignatureType::Secp256r1_PrehashedSha256, 4_000),
         (SignatureType::Secp384r1_PrehashedSha384, 18_000),
+        (SignatureType::Sr25519_Pure, 1_000),
     ])
 });
 
@@ -58,6 +59,7 @@ static SIGN_MESSAGE_COST: Lazy<HashMap<SignatureType, (u64, u64)>> = Lazy::new(|
         (SignatureType::Secp256k1_PrehashedSha256, (3_000, 0)),
         (SignatureType::Secp256r1_PrehashedSha256, (9_000, 0)),
         (SignatureType::Secp384r1_PrehashedSha384, (43_200, 0)),
+        (SignatureType::Sr25519_Pure, (1_500, 8)),
     ])
 });
 
@@ -72,6 +74,7 @@ static VERIFY_MESSAGE_COST: Lazy<HashMap<SignatureType, (u64, u64)>> = Lazy::new
         (SignatureType::Secp256k1_PrehashedSha256, (3_000, 0)),
         (SignatureType::Secp256r1_PrehashedSha256, (7_900, 0)),
         (SignatureType::Secp384r1_PrehashedSha384, (37_920, 0)),
+        (SignatureType::Sr25519_Pure, (2_000, 8)),
     ])
 });
 
@@ -680,21 +683,6 @@ mod test {
         .expect("call should return something")
         .expect_err("call should fail");
 
-        // Unsupported method.
-        let params = ethabi::encode(&[
-            Token::Uint(6.into()), // sr25519 is not yet supported.
-            Token::Bytes(b"01234567890123456789012345678901".to_vec()),
-        ]);
-        call_contract(
-            H160([
-                0x01, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0x05,
-            ]),
-            &params,
-            10_000_000,
-        )
-        .expect("call should return something")
-        .expect_err("call should fail");
-
         // Working test.
         let params = ethabi::encode(&[
             Token::Uint(SignatureType::Ed25519_Oasis.as_int().into()),
@@ -773,13 +761,18 @@ mod test {
         bench_keypair_generate(b, SignatureType::Secp384r1_PrehashedSha384);
     }
 
+    #[bench]
+    fn bench_keypair_generate_sr25519(b: &mut Bencher) {
+        bench_keypair_generate(b, SignatureType::Sr25519_Pure);
+    }
+
     #[test]
     fn test_basic_roundtrip() {
         let seed = b"01234567890123456789012345678901";
         let context = b"test context";
         let message = b"test message";
 
-        for method in 0u8..6u8 {
+        for method in 0u8..=6u8 {
             let sig_type: SignatureType = method.try_into().unwrap();
             if sig_type.is_prehashed() {
                 // Tested in test_basic_roundtrip_prehashed below.
@@ -999,11 +992,6 @@ mod test {
             .expect("call should return something")
             .expect_err("call should fail");
 
-        // Unsupported method.
-        push_all_and_test(Some(6), None, None, None) // sr25519 is not yet supported.
-            .expect("call should return something")
-            .expect_err("call should fail");
-
         // All ok, with context.
         push_all_and_test(None, None, None, None)
             .expect("call should return something")
@@ -1110,6 +1098,26 @@ mod test {
         bench_signer(b, SignatureType::Secp384r1_PrehashedSha384, false, false);
     }
 
+    #[bench]
+    fn bench_sign_sr25519_shortctx_shortmsg(b: &mut Bencher) {
+        bench_signer(b, SignatureType::Sr25519_Pure, false, false);
+    }
+
+    #[bench]
+    fn bench_sign_sr25519_shortctx_longmsg(b: &mut Bencher) {
+        bench_signer(b, SignatureType::Sr25519_Pure, false, true);
+    }
+
+    #[bench]
+    fn bench_sign_sr25519_longctx_shortmsg(b: &mut Bencher) {
+        bench_signer(b, SignatureType::Sr25519_Pure, true, false);
+    }
+
+    #[bench]
+    fn bench_sign_sr25519_longctx_longmsg(b: &mut Bencher) {
+        bench_signer(b, SignatureType::Sr25519_Pure, true, true);
+    }
+
     #[test]
     fn test_verification_params() {
         fn push_all_and_test(
@@ -1153,11 +1161,6 @@ mod test {
 
         // Bogus method.
         push_all_and_test(Some(55), None, None, None, None)
-            .expect("call should return something")
-            .expect_err("call should fail");
-
-        // Unsupported method.
-        push_all_and_test(Some(6), None, None, None, None) // sr25519 is not yet supported.
             .expect("call should return something")
             .expect_err("call should fail");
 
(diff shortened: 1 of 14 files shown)


runtime-sdk/modules/evm/src/precompile/confidential.rs
```diff
@@ -682,7 +685,7 @@ mod test {
 
         // Unsupported method.
         let params = ethabi::encode(&[
-            Token::Uint(6.into()), // sr25519 is not yet supported.
+            Token::Uint(10.into()), // 10 is invalid
```
ptrus, line 688: Consider either using the lowest invalid method number (which is 9 at the moment, I believe) or a larger value that won't be valid anytime soon (like 255)?

ptrus (approved):

runtime-sdk/modules/evm/src/precompile/confidential.rs
kostko: This should actually be removed as an invalid method is tested above this. This was for a valid method but unsupported in the precompiles and with sr25519 added there is currently no such method.

runtime-sdk/src/crypto/signature/sr25519.rs
```diff
@@ -31,7 +32,7 @@ impl PublicKey {
         Ok(PublicKey(bytes.to_vec()))
     }
 
-    /// Verify a signature.
+    /// Verify a signature used in Oasis SDK transactions.
```
peternose, line 35: Should we compute the digest and call `verify_raw` or `verify_transcript/digest`, to avoid code duplication?

runtime-sdk/src/crypto/signature/sr25519.rs
```diff
@@ -60,6 +61,27 @@ impl PublicKey {
             .map_err(|_| Error::VerificationFailed)
     }
 
+    /// Verify a signature.
+    pub fn verify_raw(
```
peternose, line 65: This is not in line with other signers, where you have ` /// Verify signature without using any domain separation scheme.`, so optionally  we could think about renaming this method.

runtime-sdk/src/crypto/signature/sr25519.rs
```diff
@@ -70,3 +92,49 @@ impl From<&'static str> for PublicKey {
...
+impl Signer for MemorySigner {
+    fn random(rng: &mut (impl RngCore + CryptoRng)) -> Result<Self, Error> {
+        Ok(Self {
+            keypair: schnorrkel::Keypair::generate_with(rng),
+        })
+    }
+
+    fn new_from_seed(seed: &[u8]) -> Result<Self, Error> {
+        let seed = sha2::Sha512_256::digest(seed);
+        let mut rng = rand_chacha::ChaChaRng::from_seed(seed.into());
```
peternose, line 110: Where is this defined or taken from?

kostko: Just an ad-hoc deterministic way of generating a key. There is no spec for generating something from a seed AFAIK and we need this in tests. But we could also move some of this to `MemorySigner::new_test` and just forward to `from_bytes` here?

kostko: Actually, we need this (or similar) as otherwise it is harder to quickly generate valid keys. @CedarMist any thoughts on this?

peternose: Once we define this function, we cannot change it as if someone started using this `Signer`, new generation won't be compatible. So the best way would be to return signer in tests and error otherwise.

CedarMist:
Constructing STROBE/Merlin transcript requires access to the underlying KeccakF[1600] permutation function, I'm wary that implementing sr25519 signing alone may not be enough if people end up reverting to Solidity to construct the message to be signed, or if the sr25519 signer makes assumption about the transcript which makes it impossible to use for other things.

kostko:
We can always introduce additional signature modes with different transcripts if needed. This is the easiest way to add into the existing framework, it would be good to know if it is sufficient for the use cases at hand.

runtime-sdk/src/crypto/signature/sr25519.rs
kostko: The thing is that `new_from_seed` should accept some sort of (random) seed without structure. But if we make it forward to `from_bytes` here, it will require a specific structure.

CedarMist: With the `generate` precompile, I look at it as a `Multiply base point by scalar` precompile.

Any modification to the scalar before point multiplication will cause problems.

kostko: Can you provide this in terms of [`schnorrkel`](https://docs.rs/schnorrkel/0.11.4/schnorrkel/)?

runtime-sdk/src/crypto/signature/sr25519.rs
```diff
@@ -55,6 +51,33 @@ impl PublicKey {
...
+        signature: &Signature,
+    ) -> Result<(), Error> {
+        // Convert the context to a Sr25519 SigningContext.
+        let context = schnorrkel::signing_context(context);
+        let transcript = context.bytes(message);
+        self.verify_transcript(transcript, signature)
+    }
+
+    /// Verify a signature using the given transcript.
+    pub fn verify_transcript<T: SigningTranscript>(
```
CedarMist, line 71: Signature verification is Base64 encoded? So... I have to base64 encode it in Solidity before passing into verification function?

kostko: Huh? Where did you see base64 encoding?

CedarMist: Oh my bad, it was from `impl From<&'static str> for PublicKey {
    fn from(s: &'static str) -> PublicKey {
        PublicKey::from_bytes(&BASE64_STANDARD.decode(s).unwrap()).unwrap()
    }
}`

runtime-sdk/src/crypto/signature/sr25519.rs
kostko: E.g. this could use `MiniSecretKey` and expansion, similar to Ed25519? Then the question is which kind of expansion, uniform or ed25519-like?

CedarMist: This does not match Polkadot.

See: https://github.com/paulmillr/micro-sr25519/blob/017d85616c749b3714c3fba3c80b6c458729cc58/index.ts#L238

```typescript
export function secretFromSeed(seed: Uint8Array): Uint8Array {
  abytes('seed', seed, 32);
  const r = sha512(seed);
  // NOTE: different from ed25519
  r[0] &= 248;
  r[31] &= 63;
  r[31] |= 64;
  // this will strip upper 3 bits and lower 3 bits
  const key = encodeScalar(decodeScalar(r.subarray(0, 32)));
  const nonce = r.subarray(32, 64);
  return concatBytes(key, nonce);
}
```

See the test vectors at: https://github.com/polkadot-js/wasm/blob/10010830094e7d033bd11b16c5e3bc01a7045309/packages/wasm-crypto/src/rs/sr25519.rs#L176

PaulMillr's sr25519 library matches the Polkadot wasm-crypto package.

runtime-sdk/src/crypto/signature/sr25519.rs
CedarMist: This matches the polkadot implementation.

CedarMist:
Good news: current implementation can verify test case signature.

Bad news:

 1. Key derivation from seed doesn't match Polkadot implementation
 2. micro-sr25519 can't verify messages it signs with a keypair generated by on-chain implementation, neither can on-chain implementation verify a message signed by micro-sr25519 with keypair generated by on-chain implementation (basic round-trip tests)

```typescript

    // Key derivation from polkadot test cases
    // See: https://github.com/polkadot-js/wasm/blob/10010830094e7d033bd11b16c5e3bc01a7045309/packages/wasm-crypto/src/rs/sr25519.rs#L176
    const secretSeed = getBytes("0xfac7959dbfe72f052e5a0c3c8d6530f202b02fd8f9f5ca3580ec8deb7797479e");
    const secretKey = sr25519.secretFromSeed(secretSeed);
    const publicKey = sr25519.getPublicKey(secretKey);
    expect(hexlify(publicKey)).eq("0x46ebddef8cd9bb167dc30878d7113b7e168e6f0646beffd77d69d39bad76b47a");

    // Known valid signature
    const msg = new TextEncoder().encode("<Bytes>message to sign</Bytes>");
    const sig = getBytes("0x48ce2c90e08651adfc8ecef84e916f6d1bb51ebebd16150ee12df247841a5437951ea0f9d632ca165e6ab391532e75e701be6a1caa88c8a6bcca3511f55b4183");
    const sigSigner = getBytes("0xf84d048da2ddae2d9d8fd6763f469566e8817a26114f39408de15547f6d47805");

    // Verify JS implementation matches polkadot test case signature
    const isValid = sr25519.verify(msg, sig, sigSigner);
    expect(isValid).eq(true);

    const CONTEXT = new TextEncoder().encode('substrate');

    // Verify on-chain implementation also works
    const result = await se.testVerify(6, sigSigner, CONTEXT, msg, sig);
    expect(result).eq(true);

    // Test key generation
    const generatedKey = await se.testKeygen(6, secretSeed);
    // 64 byte secret, appended with 32 byte public key
    expect(getBytes(generatedKey.secretKey).length).eq(96);
    expect(hexlify(getBytes(generatedKey.secretKey).slice(64))).eq(generatedKey.publicKey);

    // JS can verify on-chain signed message
    const onchainSigned = await se.testSign(6, generatedKey.secretKey, CONTEXT, msg);
    const jsVerify = sr25519.verify(msg, getBytes(onchainSigned), getBytes(generatedKey.publicKey));
    expect(jsVerify).eq(true);
    // And on-chain can verify on-chain signed message
    expect(await se.testVerify(6, generatedKey.publicKey, CONTEXT, msg, onchainSigned)).eq(true);

    // JS roundtrip with on-chain generated keypair
    const jsSigned = sr25519.sign(getBytes(generatedKey.secretKey).slice(0, 64), msg);
    expect(sr25519.verify(msg, jsSigned, getBytes(generatedKey.publicKey))).eq(false); // FAIL

    // on-chain verify JS signed message
    const onchainVerify = await se.testVerify(6, generatedKey.publicKey, CONTEXT, msg, jsSigned);
    expect(onchainVerify).eq(false); // FAIL
```

runtime-sdk/src/crypto/signature/sr25519.rs
kostko: Changed derivation to use Ed25519-based expansion as implemented by `schnorrkel` (and also used by the linked WASM implementation).

kostko:
The other thing seems to be that there are two ways to encode a keypair:
* `Keypair::to_bytes`: Serialize Keypair to bytes.
* `Keypair::to_half_ed25519_bytes`: Serialize Keypair to bytes with Ed25519 secret key format.

It looks like we are using `to_bytes` but the Polkadot libraries are using `to_half_ed25519_bytes`? We can change that as well.

CedarMist (approved):
This matches the Polkadot implementation and passes the tests I made in sapphire-contracts: https://github.com/oasisprotocol/sapphire-paratime/pull/469

PR #2108, opened 2024-12-12
Add rofl-containers generic ROFL app runtime for containers

kostko:


rofl-containers/src/main.rs
```diff
@@ -0,0 +1,44 @@
+use std::env;
+
+use base64::prelude::*;
+use oasis_runtime_sdk::{cbor, modules::rofl::app::prelude::*};
+
+/// UNIX socket address where the REST API server will listen on.
+const ROFL_APPD_ADDRESS: &str = "unix:/run/rofl-appd.sock";
+
+/// A generic container-based ROFL application.
```
ptrus, line 9: Am I correct that there's nothing really container specific/related in this app at the moment. It just an app that starts the rofl-appd server (which exposes the kms via a rest api)? 

Are there any docs for future plans / goals of this app?

rofl-appd/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+# Oasis SDK.
+oasis-runtime-sdk = { path = "../runtime-sdk", features = ["tdx"] }
+
+# Third party.
+anyhow = "1.0.86"
+rocket = { git = "https://github.com/rwf2/Rocket", rev = "28891e8072136f4641a33fb8c3f2aafce9d88d5b", features = ["json"] }
```
ptrus, line 12: Please add a short comment why a specific commit is used (and maybe a TODO to update to a release once the required features are released)

runtime-sdk/src/modules/rofl/types.rs
```diff
@@ -16,6 +18,8 @@ pub struct Create {
     pub policy: AppAuthPolicy,
     /// Identifier generation scheme.
     pub scheme: IdentifierScheme,
+    /// Metadata (arbitrary key/value pairs).
+    pub metadata: BTreeMap<String, String>,
```
ptrus, line 22: Create doesn't have Secrets since it cannot be encrypted due to "initial fee payments", right?

Maybe mention this in `AppConfig` docstring, since otherwise it might not be totally clear why we have separate metadata and secrets fields.

kostko: Yes, exactly. The SEK is only generated once the app is registered as it depends on the assigned app ID. Will add a comment.

rofl-containers/src/main.rs
kostko: Will note this in the docs.

rofl-appd/Cargo.toml
kostko: Added.

rofl-containers/src/main.rs
kostko: Added a section to the file documentation comment.

runtime-sdk/src/modules/rofl/types.rs
kostko: Updated, please take a look.

rofl-appd/Cargo.toml
ptrus: Thanks!

rofl-containers/src/main.rs
ptrus: Looks good, thanks! 👍

runtime-sdk/src/modules/rofl/types.rs
ptrus: Thanks!

ptrus (approved):

PR #2068, opened 2024-11-19
docs: Update Draw.io SVG to dark mode

aefhm:
## Description

Supports https://github.com/oasisprotocol/docs/issues/308

PR #2120, opened 2024-12-20
docs: Remove custom SVG font

aefhm:
## Description

During #2068, we found out that font embedding is inconsistent through our SVGs.

We could consolidate by removing embedded ones.


aefhm:
Actually, I go back on this one. "Architects Daughter" looks a lot better? @matevz

matevz:
> Actually, I go back on this one. "Architects Daughter" looks a lot better? @matevz

Let's keep the embedded font, because the system ones, at least on my machine, look inconsistent with the rest of the diagram style (hand-written like lines).

image

Issue #1856, opened 2024-07-03
Support mapping native denominations to ERC-20


kostko:
Needed by DeltaDAO and may also be useful in general. There would be a way to register a complex precompile at a specific address which would expose an ERC-20 ABI that would manipulate the underlying SDK denomination.

There should also be a way to designate an authorized mint/burn address to facilitate supply changes.

PR #2144, opened 2025-02-04
runtime-sdk: Implement ERC-20 compatibility

jberci:
Implements #1856.

### DIFF
--- a/runtime-sdk-macros/src/evm_derive.rs
+++ b/runtime-sdk-macros/src/evm_derive.rs
@@ -1,0 +1,471 @@
+use darling::{
+    util::{Flag, SpannedValue},
+    FromAttributes, FromDeriveInput, FromField, FromVariant,
+};
+use proc_macro2::TokenStream;
+use quote::{format_ident, quote};
+use syn::{DeriveInput, FnArg, Ident, ImplItem, ImplItemFn, ItemImpl, Pat, Path};
+use tiny_keccak::{Hasher as _, Keccak};
+
+const SELECTOR_LENGTH: usize = 4;
+
+#[derive(FromAttributes)]
+#[darling(attributes(evm_method))]
+struct EvmMethod {
+    signature: SpannedValue<String>,
+    convert: Flag,
+}
+
+#[derive(Debug, Clone)]
+struct Signature {
+    selector: [u8; SELECTOR_LENGTH],
+    name: Ident,
+    convert: bool,
+    args: Vec<String>,
+}
+
+fn evm_crate_path() -> Path {
+    let is_internal = std::env::var("CARGO_PKG_NAME")
+        .map(|pkg_name| pkg_name == "oasis-runtime-sdk-evm")
+        .unwrap_or_default();
+    if is_internal {
+        // Doctests are their own crates, but they share the name of the primary crate.
+        // Thus, the primary crate needs to refer to itself. Either that or depend on unstable
+        // rustdoc env vars.
+        syn::parse_quote!(crate::oasis_runtime_sdk_evm)
+    } else {
+        syn::parse_quote!(::oasis_runtime_sdk_evm)
+    }
+}
+
+fn parse_signature(name: Ident, signature: &str) -> Signature {
+    let mut hash = [0u8; 32];
+    let mut keccak = Keccak::v256();
+    keccak.update(signature.as_bytes());
+    keccak.finalize(&mut hash);
+    let mut selector = [0u8; SELECTOR_LENGTH];
+    selector.copy_from_slice(&hash[..SELECTOR_LENGTH]);
+
+    let mut args = Vec::new();
+    if let Some((_, after_paren)) = signature.split_once('(') {
+        if let Some((arg_list, _)) = after_paren.split_once(')') {
+            args = arg_list
+                .split(',')
+                .map(str::to_string)
+                .filter(|s| !s.is_empty())
+                .collect();
+        }
+    }
+
+    Signature {
+        selector,
+        name,
+        convert: false,
+        args,
+    }
+}
+
+fn get_method(f: &ImplItemFn) -> Option<(Signature, EvmMethod, TokenStream)> {
+    let attr = match EvmMethod::from_attributes(&f.attrs) {
+        Ok(a) => a,
+        Err(e) => {
+            e.span().unwrap().error(format!("{}", e)).emit();
+            return None;
+        }
+    };
+
+    let mut signature = parse_signature(f.sig.ident.clone(), attr.signature.as_str());
+    signature.convert = attr.convert.is_present();
+    for inp in f.sig.inputs.iter() {
+        match inp {
+            FnArg::Receiver(recv) => {
+                recv.self_token
+                    .span
+                    .unwrap()
+                    .error("Dispatchable methods can't have a receiver")
+                    .emit();
+                return None;
+            }
+            FnArg::Typed(typed) => match *typed.pat {
+                Pat::Ident(_) => continue,
+                _ => {
+                    typed
+                        .colon_token
+                        .span
+                        .unwrap()
+                        .error("Unsupported function input type")
+                        .emit();
+                    return None;
+                }
+            },
+        }
+    }
+    if signature.convert && signature.args.len() != f.sig.inputs.len() - 1 {
+        signature
+            .name
+            .span()
+            .unwrap()
+            .error("Number of function arguments doesn't match EVM signature")
+            .emit();
+        return None;
+    }
+
+    let byte_tokens = signature.selector.iter().map(|&b| quote!(#b));
+    let byte_list = quote!(#(#byte_tokens ,)*);
+    Some((signature, attr, byte_list))
+}
+
+fn generate_method_call(attr: &EvmMethod, sig: &Signature) -> Option<TokenStream> {
+    let mut decoder_list: Vec<TokenStream> = Vec::new();
+    let mut call_args: Vec<Ident> = Vec::new();
+    let mut arg_decls: Vec<TokenStream> = Vec::new();
+    for (i, evm_ty) in sig.args.iter().enumerate() {
+        let arg_name = format_ident!("arg_{}", i);
+        call_args.push(arg_name.clone());
+        match evm_ty.as_str() {
+            "address" => {
+                decoder_list.push(quote!(::ethabi::ParamType::Address));
+                arg_decls.push(quote! {
+                    let #arg_name = decoded_args[#i].clone().into_address().unwrap();
+                });
+            }
+            "uint256" => {
+                decoder_list.push(quote!(::ethabi::ParamType::Uint(256)));
+                let temp_name = format_ident!("{}_uint", arg_name);
+                arg_decls.push(quote! {
+                    let #temp_name = decoded_args[#i].clone().into_uint().unwrap();
+                    if #temp_name.bits() > (u128::BITS as usize) {
+                        return Some(Err(::evm::executor::stack::PrecompileFailure::Error {
+                            exit_status: ::evm::ExitError::Other("integer overflow".into()),
+                        }));
+                    }
+                    let #arg_name = #temp_name.as_u128();
+                });
+            }
+            ty => {
+                attr.signature
+                    .span()
+                    .unwrap()
+                    .error(format!("Unknown argument type '{}'", ty))
+                    .emit();
+            }
+        }
+    }
+    let method_name = sig.name.clone();
+    Some(quote! { {
+        let decoded_args = match ::ethabi::decode(&[#(#decoder_list),*], &handle.input()[#SELECTOR_LENGTH..]) {
+            Err(e) => return Some(Err(::evm::executor::stack::PrecompileFailure::Error {
+                exit_status: ::evm::ExitError::Other("invalid argument".into()),
+            })),
+            Ok(tokens) => tokens,
+        };
+        #(#arg_decls)*
+        Some(Self::#method_name(handle, #(#call_args),*))
+    } })
+}
+
+pub fn derive_evm_contract(input: ItemImpl) -> TokenStream {
+    let mut methods: Vec<(Signature, EvmMethod, TokenStream)> = Vec::new();
+    let mut address_fn: Option<_> = None;
+
+    for item in input.items.iter() {
+        if let ImplItem::Fn(f) = item {
+            if f.attrs
+                .iter()
+                .any(|attr| attr.path().is_ident("evm_method"))
+            {
+                methods.extend(get_method(f).into_iter());
+            }
+            if f.attrs
+                .iter()
+                .any(|attr| attr.path().is_ident("evm_contract_address"))
+            {
+                address_fn = Some(f.sig.ident.clone());
+            }
+        }
+    }
+    if address_fn.is_none() {
+        input.impl_token
+        .span.unwrap()
+            .error("Missing implementation for address function (use the `evm_contract_address` attribute)")
+            .emit();
+        return TokenStream::new();
+    }
+
+    let method_arms: Vec<TokenStream> = methods
+        .into_iter()
+        .map(|(sig, method, keccak)| {
+            if sig.convert {
+                let decode_body = generate_method_call(&method, &sig);
+                quote! {
+                    &[#keccak] => { #decode_body }
+                }
+            } else {
+                let method_name = sig.name.clone();
+                quote! {
+                    &[#keccak] => Some(Self::#method_name(handle, #SELECTOR_LENGTH)),
+                }
+            }
+        })
+        .collect();
+
+    let impl_generics = &input.generics;
+    let self_ty = input.self_ty.clone();
+    let evm_crate = evm_crate_path();
+    quote! {
+        #input
+
+        #[automatically_derived]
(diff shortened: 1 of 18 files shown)


runtime-sdk/modules/evm/src/precompile/erc20.rs
```diff
@@ -0,0 +1,404 @@
...
+        input: &[u8],
+        _gas_limit: Option<u64>,
+        ctx: &Context,
+        _is_static: bool,
+    ) -> Result<(PrecompileOutput, u64), PrecompileFailure> {
+        let params = Self::decode_params(&[ParamType::Address, ParamType::Uint(256)], input)?;
+        let spender = Self::sdk_address(&params[0]);
+        let amount = Self::sdk_amount(&params[1])?;
+        let owner = address::Address::from_eth(ctx.caller.as_bytes());
+        match T::approve(owner, spender, &amount) {
```
ptrus, line 319: This should also emit an ERC-20 Approval log.

ptrus (commented):
The approach seems sensible to me 👍 . A couple of suggestions regarding the ERC-20 implementation (skipped reviewing the macros part for now, since rust macro code is so unreadable to me 😅)

runtime-sdk/modules/evm/src/precompile/erc20.rs
```diff
@@ -0,0 +1,404 @@
...
+    ) -> Result<bool, Error> {
+        Ok(false) // XXX
+    }
+
+    fn allowance(_owner: address::Address, _spender: address::Address) -> Result<u128, Error> {
+        Ok(0) // XXX
+    }
+
+    fn mint(to: address::Address, amount: &token::BaseUnits) -> Result<(), Error> {
+        T::Accounts::mint(to, amount)
```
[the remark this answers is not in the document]
jberci: Minting and burning specifically are requirements for this feature. People usually seem to implement these two methods on top of the base ERC-20 spec so I added them, but we can decide which particular interface to follow.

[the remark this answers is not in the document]
ptrus: Okay, that makes sense. But I believe that in that case, we need to include a configurable ownership mechanism as well. Since without it, everyone will be able to call this method and mint tokens?

runtime-sdk/modules/evm/src/precompile/erc20.rs
```diff
@@ -0,0 +1,435 @@
...
+    fn approve(
+        _owner: address::Address,
+        _spender: address::Address,
+        _amount: &token::BaseUnits,
+    ) -> Result<bool, Error> {
+        Ok(false) // XXX
+    }
+
+    fn allowance(_owner: address::Address, _spender: address::Address) -> Result<u128, Error> {
+        Ok(0) // XXX
```
ptrus, line 135: Do we plan to support this?  Otherwise, it might it be better to fail on the unsupported methods?

runtime-sdk/modules/evm/src/test.rs
```diff
@@ -1201,3 +1207,527 @@ fn test_return_value_limits() {
     assert_eq!(result[64], 0xFF, "result should be correct");
     assert_eq!(result[1023], 0x42, "result should be correct");
 }
+
+#[derive(Default)]
+pub struct TestErcToken {}
```
ptrus, line 1212: This is maybe more for a follow-up, potentially it should go even in a separate repo ([sapphire-paratime](https://github.com/oasisprotocol/sapphire-paratime)?), but we should consider running some ERC-20 token test cases against our precompile implementation?

e.g. https://github.com/OpenZeppelin/openzeppelin-contracts/tree/master/test/token/ERC20

ptrus (commented):
Updates look good to me. A couple of questions/suggestions about strictly following the ERC-20 spec / Openzeppelins ERC20 token implementation.

runtime-sdk/modules/evm/src/precompile/erc20.rs
```diff
@@ -0,0 +1,491 @@
...
+
+    fn transfer_from(
+        owner: &H160,
+        spender: &H160,
+        recipient: &H160,
+        amount: &token::BaseUnits,
+    ) -> Result<bool, Error> {
+        with_allowance::<Self, _>(owner, spender, |allowance| {
+            let allowance = allowance.unwrap_or_default();
+            if amount.amount() <= allowance {
```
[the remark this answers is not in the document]
jberci: There's an issue here that amounts in EVM are u256, but we use only u128 for them. I can do the equivalent of infinity but for u128, not sure how wrong that would be.

runtime-sdk/modules/evm/src/test.rs
```diff
@@ -41,11 +41,28 @@ static FAUCET_CONTRACT_CODE_HEX: &str =
...
+        struct Token {}
+        impl p::erc20::AccountToken for Token {
+            type Accounts = Accounts;
+            const GAS_COSTS: p::erc20::TokenOperationCosts = p::erc20::TokenOperationCosts::new();
+            const NAME: &str = "Token";
+            const SYMBOL: &str = "TOK";
+            const DECIMALS: u8 = 18;
+        }
+        Some(BTreeMap::from([(
+            primitive_types::H160([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]),
```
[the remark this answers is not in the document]
ptrus: I guess we don't do any checks for existing precompiles either, resolving.

runtime-sdk/modules/evm/src/precompile/erc20.rs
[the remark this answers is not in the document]
ptrus: This was implemetned 👍

runtime-sdk/modules/evm/src/precompile/erc20.rs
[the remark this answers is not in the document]
ptrus: Yeah probably keep it as is then

runtime-sdk/modules/evm/src/precompile/erc20.rs
```diff
@@ -0,0 +1,586 @@
...
+/// Errors emitted by the ERC-20 implementation.
+#[derive(thiserror::Error, Debug, oasis_runtime_sdk_macros::Error)]
+pub enum Error {
+    #[error("forbidden by policy")]
+    #[sdk_error(code = 1)]
+    Forbidden,
+
+    #[error("invalid sender address {0}")]
+    #[sdk_error(code = 2)]
+    InvalidSender(H160),
```
ptrus, line 31: For errors that follow the EIP-6093 it would be great to follow the exact same naming, to make it clearer. Maybe also include a comment. Something like:

```suggestion
    /// EIP-6093 defined error.
    ERC20InvalidReceiver(H160),
```
Maybe the comment is not needed.


Btw how do these errors end up encoded? I guess just as strings "invalid sender address ..."?
Should we make it so that these predefined ERC20 errors end up encoded the same as in ERC-20 smart contracts? So I believe it should be: `keccak256("ERC20InvalidSender(address)")[:4]+<address>` for this one specifically?

runtime-sdk/modules/evm/src/precompile/erc20.rs
```diff
@@ -0,0 +1,552 @@
...
+        }
+        if spender.is_zero() {
+            return Err(Error::ERC20InvalidSpender(*spender));
+        }
+        if recipient.is_zero() {
+            return Err(Error::ERC20InvalidReceiver(*recipient));
+        }
+        with_allowance::<Self, _>(owner, spender, |allowance| {
+            let allowance = allowance.unwrap_or_default();
+            if allowance == u128::MAX {
```
ptrus, line 257: Feel free to remove this, as you pointed out in https://github.com/oasisprotocol/oasis-sdk/pull/2144#discussion_r1971391385 the amounts in EMV are U256, so maybe it's better we don't do this and it also simplifies the code a bit.

ptrus (approved):
Thanks for the updates, looks good to me!

Just a couple of minor comments.

runtime-sdk/modules/evm/src/precompile/contract.rs
```diff
@@ -0,0 +1,26 @@
...
+    fn dispatch_call(handle: &mut impl PrecompileHandle) -> Option<PrecompileResult>;
+}
+
+pub trait EvmEvent {
+    fn emit<C: StaticContract>(&self, handle: &mut impl PrecompileHandle) -> Result<(), ExitError>;
+}
+
+pub trait EvmError {
+    fn encode(&self) -> PrecompileFailure;
+}
```
kostko, line 26: Add documentation comments.

runtime-sdk/modules/evm/src/precompile/erc20.rs
```diff
@@ -0,0 +1,556 @@
...
+    const NAME: &str = T::NAME;
+    const SYMBOL: &str = T::SYMBOL;
+    const DECIMALS: u8 = T::DECIMALS;
+
+    fn address() -> H160 {
+        Self::ADDRESS
+    }
+
+    fn total_supply() -> Result<u128, Error> {
+        Ok(T::Accounts::get_total_supplies()?
```
kostko, line 220: Add `Accounts::get_total_supply(denomination)` to avoid first iterating over all total supplies for no good reason. In storage, each supply is its own storage location so this should be straightforward.

runtime-sdk/modules/evm/src/precompile/erc20.rs
```diff
@@ -0,0 +1,556 @@
...
+    transfer_from: u64,
+    approve: u64,
+    allowance: u64,
+    mint: u64,
+    burn: u64,
+}
+
+impl TokenOperationCosts {
+    pub const fn new() -> Self {
+        Self {
```
kostko, line 119: Why do you need `new` since it looks the same as `Default::default`? I would implement a custom `Default` instead and use non-zero values.

runtime-sdk/modules/evm/src/precompile/erc20.rs
```diff
@@ -0,0 +1,556 @@
...
+
+    fn total_supply() -> Result<u128, Error> {
+        Ok(T::Accounts::get_total_supplies()?
+            .get(&T::denomination())
+            .copied()
+            .unwrap_or_default())
+    }
+
+    fn balance_of(account: &H160) -> Result<u128, Error> {
+        let account = address::Address::from_eth(account.as_bytes());
```
kostko, line 227: I would just import `Address` to avoid repetition.

runtime-sdk/modules/evm/src/precompile/erc20.rs
```diff
@@ -0,0 +1,556 @@
...
+        }
+        with_allowance::<Self, _>(owner, spender, |allowance| {
+            let allowance = allowance.unwrap_or_default();
+            if amount <= allowance {
+                Ok(allowance - amount)
+            } else {
+                Err(Error::ERC20InsufficientAllowance(
+                    *spender, allowance, amount,
+                ))
+            }
```
kostko, line 269: You could use `checked_sub` here, e.g.:
```rust
allowance.checked_sub(amount).ok_or(Error::ERC20InsufficientAllowance(...))
```

runtime-sdk/modules/evm/src/precompile/erc20.rs
```diff
@@ -0,0 +1,552 @@
...
+            exit_status: ExitSucceed::Returned,
+            output: ethabi::encode(&[Token::Uint(T::DECIMALS.into())]),
+        })
+    }
+
+    #[evm_method(signature = "totalSupply()")]
+    fn total_supply(handle: &mut impl PrecompileHandle, _input_offset: usize) -> PrecompileResult {
+        match T::total_supply() {
+            Ok(amount) => {
+                handle.record_cost(T::GAS_COSTS.total_supply)?;
```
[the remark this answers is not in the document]
kostko: Yes, gas should be recorded _before_ doing any execution as if there is not enough gas, execution should not even be attempted.

runtime-sdk/modules/evm/src/precompile/erc20.rs
```diff
@@ -0,0 +1,556 @@
...
+use oasis_runtime_sdk_macros::{evm_contract_address, evm_method, sdk_derive, EvmEvent};
+use primitive_types::H160;
+
+use crate::precompile::{
+    contract::{EvmError as _, EvmEvent as _},
+    PrecompileResult,
+};
+
+const MODULE_NAME: &str = "evm/erc20";
+const STATE_ALLOWANCES: &[u8] = &[0x01];
```
kostko, line 23: Is there a reason why this wouldn't be better suited for the accounts module?

runtime-sdk-macros/src/evm_derive.rs
```diff
@@ -0,0 +1,467 @@
...
+                    let #temp_name = decoded_args[#i].clone().into_uint().unwrap();
+                    if #temp_name.bits() > (u128::BITS as usize) {
+                        return Some(Err(::evm::executor::stack::PrecompileFailure::Error {
+                            exit_status: ::evm::ExitError::Other("integer overflow".into()),
+                        }));
+                    }
+                    let #arg_name = #temp_name.as_u128();
+                });
+            }
+            ty => {
```
kostko, line 145: I guess this is currently enough for ERC20 and we can extend it later if needed.

runtime-sdk-macros/src/evm_derive.rs
```diff
@@ -0,0 +1,467 @@
...
+                    .span()
+                    .unwrap()
+                    .error(format!("Unknown argument type '{}'", ty))
+                    .emit();
+            }
+        }
+    }
+    let method_name = sig.name.clone();
+    Some(quote! { {
+        let decoded_args = match ::ethabi::decode(&[#(#decoder_list),*], &handle.input()[#SELECTOR_LENGTH..]) {
```
kostko, line 156: Should this use `decode_whole` instead? What is standard EVM behavior in this case as we should probably be consistent?

runtime-sdk-macros/src/evm_derive.rs
```diff
@@ -0,0 +1,467 @@
...
+        impl #impl_generics #evm_crate::precompile::contract::StaticContract for #self_ty {
+            fn address() -> ::primitive_types::H160 {
+                Self::#address_fn()
+            }
+
+            fn dispatch_call(
+                handle: &mut impl ::evm::executor::stack::PrecompileHandle,
+            ) -> Option<Result<#evm_crate::precompile::PrecompileOutput, #evm_crate::precompile::PrecompileFailure>> {
+                use #evm_crate::precompile as precompiles;
+                let selector = match handle.input().get(..4) {
```
kostko, line 228: This uses `4` while some places use `SELECTOR_LENGTH`.

kostko (commented):
Nice!

Would be great to fuzz the implementation since it involves decoding to make sure it doesn't crash on method calls.

runtime-sdk/modules/evm/src/precompile/erc20.rs
jberci: I didn't implement `Default` because it can't be `const`, but that's also the only reason. Would it be better to forego constness and change the cost const in the trait to a function?

runtime-sdk-macros/src/evm_derive.rs
jberci: Yup, I implemented only the ones I needed. It should be trivial to add more types as necessary.

runtime-sdk-macros/src/evm_derive.rs
jberci: If I'm reading the specification (https://docs.soliditylang.org/en/v0.8.28/abi-spec.html) right, it doesn't specify decoding any further than "inverse of encoding", so various decoding corner cases are not covered. In practice apparently neither under- nor overflows of call data are handled specially, so various attacks are possible...

runtime-sdk/modules/evm/src/precompile/erc20.rs
```diff
@@ -0,0 +1,549 @@
...
+    F: FnOnce(Option<u128>) -> Result<u128, Error>,
+{
+    let owner = owner.as_bytes();
+    let spender = spender.as_bytes();
+    CurrentState::with_store(|store| {
+        let store = storage::PrefixStore::new(store, &MODULE_NAME);
+        let allowances = storage::PrefixStore::new(store, &STATE_ALLOWANCES);
+        let denomination = T::denomination().to_string();
+        let allowances = storage::PrefixStore::new(allowances, &denomination);
+        let mut owner = storage::TypedStore::new(storage::PrefixStore::new(allowances, &owner));
```
kostko, line 196: So this is per-denomination and multiple ERC-20 contracts share the same allowances? I guess this makes sense, but also makes it seems like it would be even more suited for the accounts module, even if it is initially only available through the internal API (e.g. no transactions/calls for modifying/using allowances, just API methods that currently only this ERC20 implementation would use).

runtime-sdk/src/modules/accounts/mod.rs
```diff
@@ -543,6 +562,33 @@ impl API for Module {
         })
     }
 
+    fn get_allowance(
+        owner: Address,
+        user: Address,
```
kostko, line 567: Could use the term "beneficiary" as we do in Oasis Core.

kostko (approved):

PR #2145, opened 2025-02-05
runtime-sdk/modules/rofl: Add watchdog task

### LINKED ISSUE #2143: rofl: Add watchdog task
There should be a simple watchdog task for ROFL apps which causes the runtime to abort in case no registration refresh has been processed in a while. This ensures that the runtime will stop in case it is not able to re-attest and/or re-register.

kostko:
Fixes #2143 

The watchdog task ensures that the app is successfully re-registering and terminates it if this is not the case.

### DIFF
--- a/rofl-containers/Cargo.toml
+++ b/rofl-containers/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rofl-containers"
-version = "0.3.3"
+version = "0.3.5"
 edition = "2021"
 
 [dependencies]
--- a/runtime-sdk/src/modules/rofl/app/watchdog.rs
+++ b/runtime-sdk/src/modules/rofl/app/watchdog.rs
@@ -1,0 +1,77 @@
+use anyhow::Result;
+use tokio::{sync::mpsc, time};
+
+use crate::core::common::{logger::get_logger, process};
+
+/// Interval in which at least one keep-alive must be delivered to avoid the watchdog from
+/// terminating the application.
+const WATCHDOG_TRIGGER_INTERVAL: u64 = 6 * 3600; // 6 hours
+
+/// Application watchdog task.
+pub(super) struct Task {
+    imp: Option<Impl>,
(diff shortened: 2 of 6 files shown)


runtime-sdk/src/modules/rofl/app/watchdog.rs
```diff
@@ -0,0 +1,77 @@
+use anyhow::Result;
+use tokio::{sync::mpsc, time};
+
+use crate::core::common::{logger::get_logger, process};
+
+/// Interval in which at least one keep-alive must be delivered to avoid the watchdog from
+/// terminating the application.
+const WATCHDOG_TRIGGER_INTERVAL: u64 = 6 * 3600;
```
peternose, line 8: It is not obvious that this is in seconds.

runtime-sdk/src/modules/rofl/app/watchdog.rs
```diff
@@ -0,0 +1,77 @@
...
+
+impl Impl {
+    /// Start the application watchdog task.
+    pub(super) fn start(self) {
+        tokio::task::spawn(self.run());
+    }
+
+    /// Run the application watchdog task.
+    async fn run(mut self) {
+        slog::info!(self.logger, "starting watchdog task");
```
peternose, line 57:
```suggestion
        slog::info!(self.logger, "starting");
```
Could also shorten the text as the logger's name is `modules/rofl/app/watchdog`.

peternose (approved):

rofl-containers/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "rofl-containers"
-version = "0.3.3"
+version = "0.3.5"
```
peternose, line 3: Why 5 and not 4?

kostko: We published 0.3.4 (see tags) but we forgot to bump the crate version.

PR #2148, opened 2025-02-05
runtime-sdk/modules/rofl: Add key scope in rofl.DeriveKey

### LINKED ISSUE #2141: rofl: Add derive key scopes
Currently there is only the "global" scope (e.g. shared between all ROFL app instances under the same app ID). We could also introduce per-instance scopes (e.g. using node ID or entity ID for domain separation).

kostko:
Fixes #2141

### DIFF
--- a/runtime-sdk/src/modules/rofl/mod.rs
+++ b/runtime-sdk/src/modules/rofl/mod.rs
@@ -1,5 +1,5 @@
 //! On-chain coordination for ROFL components.
-use std::collections::BTreeSet;
+use std::collections::{BTreeMap, BTreeSet};
 
 use once_cell::sync::Lazy;
 
@@ -217,9 +217,16 @@ impl<Cfg: Config> Module<Cfg> {
         )?;
 
         // Generate the secret encryption (public) key.
-        let sek = Self::derive_app_key(ctx, &app_id, types::KeyKind::X25519, ROFL_KEY_ID_SEK)?
-            .input_keypair
-            .pk;
+        let sek = Self::derive_app_key(
+            ctx,
+            &app_id,
+            types::KeyKind::X25519,
+            types::KeyScope::Global,
+            ROFL_KEY_ID_SEK,
+            None,
+        )?
+        .input_keypair
+        .pk;
 
         // Register the application.
         let cfg = types::AppConfig {
@@ -263,9 +270,16 @@ impl<Cfg: Config> Module<Cfg> {
 
         // If there is no SEK defined, regenerate it.
         if cfg.sek == Default::default() {
-            cfg.sek = Self::derive_app_key(ctx, &body.id, types::KeyKind::X25519, ROFL_KEY_ID_SEK)?
-                .input_keypair
-                .pk;
+            cfg.sek = Self::derive_app_key(
+                ctx,
+                &body.id,
+                types::KeyKind::X25519,
+                types::KeyScope::Global,
+                ROFL_KEY_ID_SEK,
+                None,
+            )?
+            .input_keypair
+            .pk;
         }
 
         cfg.policy = body.policy;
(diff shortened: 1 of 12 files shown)


runtime-sdk/src/modules/rofl/mod.rs
```diff
@@ -426,14 +445,45 @@ impl<Cfg: Config> Module<Cfg> {
...
-            key_id,
-        ]);
+        // Build the base key identifier.
+        let kind_id = &[kind as u8];
+        let mut key_id = vec![ROFL_DERIVE_KEY_CONTEXT, app.as_ref(), kind_id, key_id];
+
+        // Append scope for non-global scopes.
+        let scope_id = &[scope as u8];
+        if !scope.is_global() {
+            key_id.push(scope_id);
```
peternose, line 464: Not sure if just appending raw values to `key_id` is a good solution. I think that we should somehow tie this to scope, and not just add raw `u8` or `[u8]` value, like context does. Or change `key_id` so that it's structure is obvious, e.g. key id = kind + optional scope id + optional scope extra data + some optional data in the future.

kostko: One thing is that we need to retain backwards compatibility for the global scope as this is what is currently deployed. So currently, the structure of the tuple is:
* Context domain separator.
* App ID.
* Encoded kind.
* Key ID.
* Encoded scope.
* Encoded scope-specific domain separation. If there are other scopes in the future, this scope-specific data would all go into this single element.

What did you have in mind that would work without changing the existing keys?

kostko: We could also introduce a V2 and completely change the structure. But this V2 would only be used for non-global scopes.

peternose: Currently, if I want to add a new field, e.g. `timeout` or `deadline`, I cannot append it to the list, as scope is optional. So I was thinking if we can instead add a new field (e.g. struct `tags`), to which we can latter add extra data and the byte encoding would be unique, as encoded scope would be tied to something, e.g. to a byte string "scope".

> * V1 context domain separator.
> * App ID.
> * Encoded kind.
> * Key ID.
> * Encoded scope.
> * Encoded scope-specific domain separation. If there are other scopes in the future, this scope-specific data would all go into this single element.

It would be nice if the context separator should have this in the comment.

kostko: Changed, please take another look.

runtime-sdk/src/modules/rofl/mod.rs
```diff
@@ -422,18 +446,75 @@ impl<Cfg: Config> Module<Cfg> {
...
+                extra_dom.insert("scope", [scope as u8].to_vec());
+                extra_dom.insert("entity_id", entity_id.as_ref().to_vec());
+            }
+        };
+
+        // Add optional extra domain separation.
+        let extra_dom = if !extra_dom.is_empty() {
+            cbor::to_vec(extra_dom)
+        } else {
+            vec![]
```
peternose, line 499: Can this be removed, i.e. if `extra_dom` is not empty, we cbor the data and push it to `key_id`?

runtime-sdk/src/modules/rofl/test.rs
```diff
@@ -242,6 +242,70 @@ fn test_create_scheme() {
     );
 }
 
+#[test]
+fn test_derive_app_key_id() {
+    // Ensure that app key identifier derivation is stable.
+    let tcs = [
```
peternose, line 248: Named fields are more readable, but this is shorter. I guess both solutions are fine :thinking: 

```rust
    struct TestCase {
        key: &'static str,
        kind: types::KeyKind,
        scope: types::KeyScope,
        data: &'static [u8],
        registration: Option<types::Registration>,
        expected: &'static str,
    }
```

peternose (approved):

PR #2168, opened 2025-03-04
tools/orc: Add support for ephemeral signer key

### LINKED ISSUE #2167: tools/orc: Add support for ephemeral signer key
Add support for the `orc` tool to generate an ephemeral key used to sign enclaves. Given that IAS is deprecated, DCAP allows any signer and we don't actually use MRSIGNER anywhere, it can be ephemeral.

kostko:
Fixes #2167

### DIFF
--- a/tools/orc/main.go
+++ b/tools/orc/main.go
@@ -1,10 +1,13 @@
 package main
 
 import (
+	"crypto/rand"
 	"crypto/rsa"
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
+	"io"
+	"math/big"
 	"os"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


tools/orc/main.go
```diff
@@ -531,6 +556,84 @@ func showTdxComponent(indent string, bnd *bundle.Bundle, comp *bundle.Component)
 	fmt.Printf("%s  Memory:  %d MiB\n", indent, comp.TDX.Resources.Memory)
 }
 
+// sgxGenerateKey generates a 3072-bit RSA key with public exponent 3 as required for SGX.
+//
+// The code below is adopted from the Go standard library as it is otherwise not possible to
+// customize the exponent.
+func sgxGenerateKey(random io.Reader) (*rsa.PrivateKey, error) {
```
peternose, line 563: Should we add some kind of test, e.g. to test that `e == 3`, `bits == 3072`, two primes, ...?

tools/orc/main.go
```diff
@@ -531,6 +556,84 @@ func showTdxComponent(indent string, bnd *bundle.Bundle, comp *bundle.Component)
 	fmt.Printf("%s  Memory:  %d MiB\n", indent, comp.TDX.Resources.Memory)
 }
 
+// sgxGenerateKey generates a 3072-bit RSA key with public exponent 3 as required for SGX.
```
peternose, line 559: I thought that small exponents should be avoided. But if they required 3, then what can you do? :shrug:

peternose (approved):

tools/orc/main.go
kostko: It doesn't really matter as MRSIGNER is not used anywhere in our system (we just compare against a reference value in the policy). So it doesn't really matter even if this key is compromised. Which is why we are adding these ephemeral keys here.

PR #2170, opened 2025-03-07
docs: Add warning to use FQDN URLs for docker images

### LINKED ISSUE #2163: rofl-docs: Mention "docker.io" domain is mandatory in compose.yaml
In https://docs.oasis.io/build/rofl/app#define-containerized-services add admonition add sth like

```
:::warn Always specify full container image URL

Always use FQDN when specifying the image URL, e.g. `docker.io/ollama/ollama`.

:::
```

matevz:
Fixes #2163

### DIFF
--- a/docs/rofl/app.mdx
+++ b/docs/rofl/app.mdx
@@ -226,7 +226,7 @@ a shell script.
 
 The compose file (`compose.yaml`) looks as follows:
 
-```yaml
+```yaml title="compose.yaml"
 services:
   oracle:
     # See the 'docker' subdirectory in demo-rofl for the content of this image.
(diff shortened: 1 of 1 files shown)


abukosek (approved):

docs/rofl/app.mdx
```diff
@@ -241,6 +241,14 @@ services:
       - /run/rofl-appd.sock:/run/rofl-appd.sock
 ```
 
+:::warning Always specify FQDN image URL
+
+In contrast to Docker compose files, always use fully qualified domain name
```
kostko, line 246: I would not say "In contrast to Docker compose files", but rather something like that this is to avoid security issues and ensure determinism, ROFL apps require fully qualified domain names. Because this is an explicit choice for security purposes, not related to Docker vs. Podman.

kostko (approved):

PR #2190, opened 2025-04-10
runtime-sdk/modules/rofl-market: Multiple fixes and improvements

kostko:
Multiple fixes and improvements discovered during the ROFL scheduler development.


runtime-sdk/modules/rofl-market/src/lib.rs
```diff
@@ -331,15 +331,18 @@ impl<Cfg: Config> Module<Cfg> {
...
+    ) -> Result<types::InstanceId, Error> {
         <C::Runtime as Runtime>::Core::use_tx_gas(Cfg::GAS_COST_CALL_INSTANCE_CREATE)?;
 
         if body.term_count == 0 {
             return Err(Error::InvalidArgument);
         }
 
         if !ctx.should_execute_contracts() {
-            return Ok(());
+            return Ok(Default::default());
```
peternose, line 345: It is hard for the caller to know if returned value was received from the happy path or from this branch. Returning error or option seems better.

runtime-sdk/modules/rofl-market/src/lib.rs
```diff
@@ -626,8 +626,10 @@ impl<Cfg: Config> Module<Cfg> {
 
         let mut provider = state::get_provider(body.provider).ok_or(Error::ProviderNotFound)?;
         Self::ensure_caller_is_scheduler_app(&provider)?;
-        let mut instance =
-            state::get_instance(body.provider, body.id).ok_or(Error::InstanceNotFound)?;
+        let mut instance = match state::get_instance(body.provider, body.id) {
+            Some(instance) => instance,
+            None => return Ok(()), // Already removed.
```
peternose, line 697: Why is this better? Again, user has no idea which road was taken. Or is expected for the caller to observe events also?

runtime-sdk/modules/rofl-market/src/error.rs
```diff
@@ -49,6 +49,10 @@ pub enum Error {
     #[sdk_error(code = 11)]
     BadResourceDescriptor(String),
 
+    #[error("invalid instance state")]
```
peternose, line 52: Would it be beneficial to show state? I prefer current solution, though.

client-sdk/go/modules/roflmarket/types.go
```diff
@@ -21,6 +22,63 @@ type (
 	CommandID [8]byte
 )
 
+// String returns a string representation of offer ID.
+func (oid OfferID) String() string {
```
peternose, line 26: I would just use `id`, targeting always for 1-2 letters.

runtime-sdk/modules/rofl-market/src/lib.rs
```diff
@@ -547,20 +557,53 @@ impl<Cfg: Config> Module<Cfg> {
...
+            if let Some(metadata) = update.metadata {
+                instance.metadata = metadata;
+                changed = true;
+            }
+
+            // Complete commands.
+            if let Some(last_completed_cmd) = update.last_completed_cmd {
+                let cmds =
+                    state::get_instance_commands(body.provider, update.id, last_completed_cmd);
+                if !cmds.is_empty() {
```
peternose, line 583: You can remove this `if` if you move `changed = true` to `for` loop.

runtime-sdk/modules/rofl-market/src/lib.rs
```diff
@@ -547,20 +557,53 @@ impl<Cfg: Config> Module<Cfg> {
...
+                        .saturating_sub(cmds.len().try_into().map_err(|_| Error::InvalidArgument)?);
+                    changed = true;
+                }
+
+                for qc in cmds {
+                    state::remove_instance_command(body.provider, update.id, qc.id);
+                }
+            }
+
+            if changed {
```
peternose, line 595:
```suggestion
            if !changed { continue }
```

peternose (approved):

runtime-sdk/modules/rofl-market/src/lib.rs
kostko: This only happens during internal simulation or transaction checks. I agree that it is somewhat awkward, but in these cases this value is _always_ discarded (as it originates from the gas estimator). So the value doesn't matter but it must satisfy the type system.

Changing it to an option doesn't make sense as in any externally visible call, this will never be `None`. We could probably use special errors for these cases, but currently this could maybe interfere with simulation and therefore it should be handled in a separate issue. Other places also handle it similarly.

runtime-sdk/modules/rofl-market/src/lib.rs
kostko: The user is the scheduler and I don't see a case where this would matter, especially since multiple scheduler instances could issue removal for the same instance. The scheduler internally uses randomized intervals to reduce probability of conflicts, but I don't see a reason why a failure here is the right thing in case the instance already does not exist. An instance with the same identifier will never exist again.

The scheduler could also just ignore the error in such cases and it would be fine, so we can also leave it as-is.

kostko: Ok, you're right, it makes sense to just fail here.

peternose: > The scheduler could also just ignore the error in such cases 

I think you should return an error and the scheduler should ignore it.

runtime-sdk/modules/rofl-market/src/lib.rs
```diff
@@ -501,22 +521,32 @@ impl<Cfg: Config> Module<Cfg> {
         Ok(())
     }
 
-    #[handler(call = "roflmarket.InstanceUpdateMetadata")]
-    fn tx_instance_update_metadata<C: Context>(
-        ctx: &C,
-        body: types::InstanceUpdateMetadata,
-    ) -> Result<(), Error> {
-        <C::Runtime as Runtime>::Core::use_tx_gas(Cfg::GAS_COST_CALL_INSTANCE_UPDATE_METADATA)?;
+    #[handler(call = "roflmarket.InstanceUpdate")]
```
ptrus, line 524: Is there a reason why some methods (such as this one, and the removed `"roflmarket.InstanceUpdateMetadata"`) are missing from the go client? 
Having a complete client in Go with all methods available from the runtime module might be nice.

kostko: Oh, it is just because they can only be called by the scheduler and I'm lazy :-) We can add them for completeness, but unless the Go client is used to implement a scheduler (which is entirely possible given TDX), users would never be able to call them.

PR #2198, opened 2025-04-28
runtime-sdk/modules/consensus-accounts: Add SharesToTokens method

### LINKED ISSUE #2166: Expose internal method to convert delegated shares into tokens at current rate
A new internal method should be exposed that converts a given number of delegated shares into corresponding tokens at the current rate.

abukosek:
Closes #2166.  Currently still a WIP.


runtime-sdk/src/modules/consensus_accounts/mod.rs
```diff
@@ -599,6 +602,36 @@ impl<Consensus: modules::consensus::API> Module<Consensus> {
...
+            .shares
+            .checked_mul(
+                u128::try_from(pool.balance.clone()).expect("pool balance should fit into u128"),
+            )
+            .expect("shares * total_amount should not overflow")
+            .checked_div(
+                u128::try_from(pool.total_shares.clone())
+                    .expect("pool total_shares should fit into u128"),
+            )
+            .expect("total_shares should not be zero");
```
kostko, line 630: This should not crash the runtime. What if one provides an account that does not have anything delegated? It should either return zero or an error (an error is probably better).

runtime-sdk/src/modules/consensus_accounts/mod.rs
```diff
@@ -599,6 +602,36 @@ impl<Consensus: modules::consensus::API> Module<Consensus> {
...
+        let pool = match args.pool {
+            types::SharePool::Active => &account.escrow.active,
+            types::SharePool::Debonding => &account.escrow.debonding,
+            _ => return Err(Error::InvalidArgument),
+        };
+
+        let tokens = args
+            .shares
+            .checked_mul(
+                u128::try_from(pool.balance.clone()).expect("pool balance should fit into u128"),
```
kostko, line 623: I would avoid crashing the runtime in these cases and would instead just return an invalid argument error.

kostko (approved):

PR #2200, opened 2025-04-29
chore: add license link

oksanaphmn:


matevz:
I don't think we need another license badget since the license is already part of the standard github UI.
image

oksanaphmn:
Yes, I see you are right, I completely agree with you

Issue #2096, opened 2024-12-05
Consider whitelisting known EVM proxy contract storage locations


kostko:
Specific storage locations (e.g. see [EIP-1967](https://eips.ethereum.org/EIPS/eip-1967)) are used for proxy contract support. Given that allowing access to them should not disclose any sensitive information, we should consider allowing access to them via regular storage accessors.

Potential locations (before required [EIP-1967](https://eips.ethereum.org/EIPS/eip-1967) transformation):
```
eip1967.proxy.beacon
eip1967.proxy.implementation
eip1967.proxy.admin
```

Additional locations from ZeppelinOS (using a [slightly different](https://blog.openzeppelin.com/proxy-patterns) transformation):
```
org.zeppelinos.proxy.implementation
org.zeppelinos.proxy.admin
```

lukaw3d:
fyi, list of magic storage slots in abi playground: https://github.com/oasisprotocol/abi-playground/blob/49facdc2e49ccbbf4d2074662c920ee64fa16805/packages/nextjs/utils/abi-ninja/proxyContracts.ts#L4-L14

PR #2215, opened 2025-05-12
runtime-sdk/evm: Allow query of known magic storage slots

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-sdk/issues/2096

This should be safe to do since:
- State variables of contracts (except for dynamically-sized arrays and mappings) is stored contiguously item after item starting with the first state variable, which is stored in slot 0
- for mappings and dynamic arrays, the data is stored at slots computed by hashing the key (and sometimes an offset) 

So unless a maliciously crafted contract explicitly writes to one of the exposed storage slots (e.g., using inline assembly or manual keccak preimage targeting), nothing of sensitive nature should end up in the exposed slots.

https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html


runtime-sdk/modules/evm/src/test.rs
```diff
@@ -1110,6 +1110,122 @@ fn test_transfer_event() {
...
+                ..Default::default()
+            },
+            ..Default::default()
+        },
+    );
+    let result: Vec<u8> = cbor::from_value(dispatch_result.result.unwrap()).unwrap();
+    assert_eq!(result.len(), 0, "result should be empty");
+
+    println!("gnossi: {:?}", gnosis_safe_proxy_interface);
+
```
kostko, line 1174:
```suggestion
```

tests/e2e/evm/contracts/evm_magic_slots.sol
```diff
@@ -0,0 +1,9 @@
+pragma solidity ^0.8.0;
+
+contract Test {
+    function setSlot(bytes32 slot,bytes32 value) external {
```
kostko, line 4:
```suggestion
    function setSlot(bytes32 slot, bytes32 value) external {
```

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -116,6 +119,34 @@ pub trait Config: 'static {
             EVM_CONFIG.get_or_init(EVMConfig::shanghai)
         }
     }
+
+    fn whitelisted_magic_slots() -> &'static HashSet<H256> {
```
kostko, line 123: Add documentation comment.

kostko (approved):

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -116,6 +119,37 @@ pub trait Config: 'static {
             EVM_CONFIG.get_or_init(EVMConfig::shanghai)
         }
     }
+
+    /// Returns the whitelisted EVM storage slots.
+    ///
```
peternose, line 124: Should this be defined outside of the function and outside of the config, removing the need for this function?

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -353,11 +387,21 @@ impl<Cfg: Config> API for Module<Cfg> {
...
     }
 
-    fn get_storage<C: Context>(_ctx: &C, address: H160, index: H256) -> Result<Vec<u8>, Error> {
-        state::with_public_storage(&address, |store| {
-            let result: H256 = store.get(index).unwrap_or_default();
-            Ok(result.as_bytes().to_vec())
-        })
+    fn get_storage<C: Context>(ctx: &C, address: H160, index: H256) -> Result<Vec<u8>, Error> {
+        let whitelisted_magic_slots = Cfg::whitelisted_magic_slots();
+        // If whitelisted magic slot, use confidential storage.
```
peternose, line 392: You could return the result here, removing a few lines.

ptrus: In this case, I prefer the if/else instead of early returning since it avoids implying that one branch is special or short-circuiting, and makes it clearer that both branches are equal alternatives based on the condition.

But I did save a line by inlining the `Cfg::whitelisted_magic_slots` call :).

runtime-sdk/modules/evm/src/lib.rs
ptrus: I think you are right, maybe we don't need this to be configurable. I guess doing a new release of the runtime SDK, if/when we need to update this, is fine? 

wdyt @kostko ?

kostko: Yeah it should be fine to not make this configurable.

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -295,6 +298,34 @@ pub trait API {
         -> Result<Vec<u8>, Error>;
 }
 
+/// Whitelisted EVM storge slots.
+///
+/// These slots are public and can be queried via `evm.Storage` queries.
+static WHITELISTED_MAGIC_SLOTS: Lazy<HashSet<H256>> = Lazy::new(|| {
+    // https://github.com/oasisprotocol/abi-playground/blob/49facdc2e49ccbbf4d2074662c920ee64fa16805/packages/nextjs/utils/abi-ninja/proxyContracts.ts#L4-L14
+    [
```
ptrus, line 306: @rube-de @matevz Please check this list if this list of slots is ok? (or are there any other slots we should include or if some of these are unneeded).

peternose (approved):

runtime-sdk/modules/evm/src/lib.rs
ptrus: Updated, thx!

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -295,6 +298,34 @@ pub trait API {
...
+        // OpenZeppelin Proxy Implementation Slot.
+        "0x7050c9e0f4ca769c69bd3a8ef740bc37934f8e2c036e5a723fd8ee048ed3f8c3",
+        // EIP-1967: Standard Proxy Implementation Slot.
+        "0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc",
+        // EIP-1967: Beacon Proxy Implementation Slot.
+        "0xa3f0ad74e5423aebfd80d3ef4346578335a9a72aeaee59ff6cb3582b35133d50",
+        // EIP-1822 Proxiable Slot.
+        "0xc5f16f0fcc639fa48a6947836d9850f504798523bf8c9a3a87d5876cf622bcf7",
+        // EIP-1167: Beacon method.
+        "0x5c60da1b00000000000000000000000000000000000000000000000000000000",
```
rube-de, line 316: that is the function selector for `implementation`. 
1167 doesn't have this function, as the minimal proxy doesn't have storage slot for the implemenation. Etherscan detects these contracts by matching the contract's deployed bytecode against the standardized EIP-1167 bytecode pattern.
I don't think we need that.

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -295,6 +298,34 @@ pub trait API {
...
+        // EIP-1167: Beacon method.
+        "0xda52571600000000000000000000000000000000000000000000000000000000",
```
rube-de, line 318: this is the function selector for `childImplementation`.
this seems to be some custom proxy implementation, not related to 1167.
I don't think we need to support a custom proxy function.

(that the git repo with the most starts which reference this selector is https://github.com/ZhangZhuoSJTU/Web3Bugs makes it even worse ;))

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -295,6 +298,34 @@ pub trait API {
...
+        // Gnosis Safe Proxy.
+        "0xa619486e00000000000000000000000000000000000000000000000000000000",
```
rube-de, line 320: the function selector for `masterCopy`.
it's not a storage slot. normally in safe the storage slot is 0 and this is the fallback function.
Do we need to whitelist a function that should be public anyway?

runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -295,6 +298,34 @@ pub trait API {
...
+        // Comptroller Proxy.
+        "0xbb82aa5e00000000000000000000000000000000000000000000000000000000",
```
rube-de, line 322: the function selector for `comptrollerImplementation`
It's not a storage slot. Do we need to whitelist a function that should be public anyway?

ptrus: Thanks for checking. Will update the list accordingly.

ptrus: I have updated the list. Do you think there is also a usecase to include the EIP-1967 Admin slot?

matevz: > I have updated the list. Do you think there is also a usecase to include the EIP-1967 Admin slot?

Yes!

Some smart contracts (ahm, hyperlane, ahm) call getStorageAt() directly as noted in https://docs.openzeppelin.com/contracts/5.x/api/proxy#ERC1967Utils-getAdmin-- to get the admin account.

runtime-sdk/modules/evm/src/lib.rs
matevz: I think `0xc5f16f0fcc639fa48a6947836d9850f504798523bf8c9a3a87d5876cf622bcf7` should also be added from the [EIP-1822](https://eips.ethereum.org/EIPS/eip-1822#functions).

matevz: Then there are some other addresses listed by @lukaw3d 's abi playground fork https://github.com/oasisprotocol/abi-playground/blob/49facdc2e49ccbbf4d2074662c920ee64fa16805/packages/nextjs/utils/abi-ninja/proxyContracts.ts#L4-L14 and our docs https://docs.oasis.io/build/sapphire/develop/deployment, but those are mostly function address slots (ending with a number of `00000...0`). So I think we're good.

ptrus: > I think 0xc5f16f0fcc639fa48a6947836d9850f504798523bf8c9a3a87d5876cf622bcf7 should also be added from the [EIP-1822](https://eips.ethereum.org/EIPS/eip-1822#functions).

Had in in the list before, but @rube-de suggested to not support it, see the reasoning: https://github.com/oasisprotocol/oasis-sdk/pull/2215#discussion_r2093544706

> Then there are some other addresses listed by @lukaw3d 's abi playground fork

See @rube-de comments. Some of them are storage slots, and some of them are apparently not widely used/outdated, so it was suggested to not both supporting those.

runtime-sdk/modules/evm/src/lib.rs
ptrus: 👍  added EIP-1967: Admin Slot address (`0xb53127684a568b3173ae13b9f8a6016e243e63b6e8ee1178d6a717850b5d6103`).

PR #2226, opened 2025-05-23
runtime-sdk/consensus_accounts: Include shares in DelegateEvent

ptrus:
It would be useful if `DelegateEvent` also returns the amount of new shares received by the delegation.

Ensure Nexus deployments are updated first, before a paratime with this change is released.


client-sdk/go/modules/consensusaccounts/types.go
```diff
@@ -199,6 +214,8 @@ type DelegateEvent struct {
 	To     types.Address   `json:"to"`
 	Amount types.BaseUnits `json:"amount"`
 	Error  *ConsensusError `json:"error,omitempty"`
+	// Added in runtime-sdk v0.15.0.
```
peternose, line 217: Will these comments be removed in the future?

peternose (approved):

client-sdk/go/modules/consensusaccounts/types.go
ptrus: Not sure. As far as I'm concerned, I can also remove the comment now. But for someone looking at the code without this comment, it won't be obvious why this field is potentially empty.

PR #2236, opened 2025-06-11
rofl-scheduler: Add API endpoint and route to get logs

### LINKED ISSUE #2229: rofl-scheduler: Add API endpoint for log retrieval
The scheduler should expose an API endpoint which can be used to retrieve logs of deployed machines. Additionally, it can also be used to issue commands in the same manner as via on-chain transactions.

kostko:
Fixes #2229

### DIFF
--- a/rofl-scheduler/src/serverd/auth.rs
+++ b/rofl-scheduler/src/serverd/auth.rs
@@ -1,0 +1,228 @@
+use std::sync::{Arc, LazyLock};
+
+use anyhow::{anyhow, Context};
+use axum::{extract, http, response, RequestPartsExt};
+use axum_extra::{
+    headers::{authorization::Bearer, Authorization},
+    TypedHeader,
+};
+use base64::{prelude::BASE64_STANDARD, Engine};
+use chrono::{prelude::*, Duration};
+use rand::RngCore;
+use rustc_hex::FromHex;
+
+use oasis_runtime_sdk::{
+    crypto::signature::{self, PublicKey},
+    types::address::{Address, SignatureAddressSpec},
+};
+
+use super::{error::Error, State};
+
+/// Global JWT keys instance.
+static JWT_KEYS: LazyLock<Keys> = LazyLock::new(Keys::generate);
+
+/// JWT keys.
+struct Keys {
+    encoding: jsonwebtoken::EncodingKey,
+    decoding: jsonwebtoken::DecodingKey,
+}
+
+impl Keys {
+    /// Generate a new pair of JWT keys.
+    fn generate() -> Self {
+        let mut key = [0; 32];
+        rand::rngs::OsRng.fill_bytes(&mut key);
+
+        Self {
+            encoding: jsonwebtoken::EncodingKey::from_secret(&key),
+            decoding: jsonwebtoken::DecodingKey::from_secret(&key),
+        }
+    }
+}
+
+/// Claims issued after successful authentication.
+#[derive(Debug, Default, Clone, serde::Serialize, serde::Deserialize)]
+pub struct Claims {
+    pub address: String,
+    exp: u64,
+}
+
+impl<S> extract::FromRequestParts<S> for Claims
+where
+    S: Send + Sync,
+{
+    type Rejection = Error;
+
+    async fn from_request_parts(
+        parts: &mut http::request::Parts,
+        _state: &S,
+    ) -> Result<Self, Self::Rejection> {
+        // Extract the token from the authorization header.
+        let TypedHeader(Authorization(bearer)) = parts
+            .extract::<TypedHeader<Authorization<Bearer>>>()
+            .await
+            .map_err(|_| Error::BadAuthToken)?;
+
+        // Validate the token and decode the user data.
+        let token_data = jsonwebtoken::decode::<Claims>(
+            bearer.token(),
+            &JWT_KEYS.decoding,
+            &jsonwebtoken::Validation::default(),
+        )
+        .map_err(|_| Error::BadAuthToken)?;
+
+        Ok(token_data.claims)
+    }
+}
+
+/// Login request.
+#[derive(Debug, Clone, serde::Deserialize)]
+#[serde(tag = "method", content = "data")]
+pub enum AuthLoginRequest {
+    #[serde(rename = "siwe")]
+    Siwe { message: String, signature: String },
+
+    #[serde(rename = "std")]
+    Std { body: String, signature: String },
+}
+
+impl AuthLoginRequest {
+    /// Verify the login request and return the authenticated claims.
+    pub async fn verify(&self, domain: &str, provider: Address) -> Result<Claims, anyhow::Error> {
+        let address = match self {
+            Self::Siwe { message, signature } => {
+                let message: siwe::Message = message.parse().context("malformed SIWE message")?;
+                let signature: Vec<u8> =
+                    signature.from_hex().context("malformed SIWE signature")?;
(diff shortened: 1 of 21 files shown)


rofl-scheduler/src/serverd/auth.rs
```diff
@@ -0,0 +1,229 @@
...
+
+    #[serde(rename = "std")]
+    Std { body: String, signature: String },
+}
+
+impl AuthLoginRequest {
+    /// Verify the login request and return the authenticated claims.
+    pub async fn verify(&self, domain: &str, provider: Address) -> Result<Claims, anyhow::Error> {
+        match self {
+            Self::Siwe { message, signature } => {
```
matevz, line 93: SIWE also contains other useful fields we could enforce for security:
- chain ID (sapphire?)
- contract address (hex-encoded address of the app ID?)
- statement (some meaningful text shown in the Metamask UI when signing)
- resources (also arbitrary text, can be the scheduler endpoint path e.g. `/rofl-scheduler/v1/logs/get`)

kostko: These seem quite useless for this use case, binding to the domain is IMO enough. The non-SIWE variant also has the provider address which we could add somewhere.

kostko: Will add the required statement of `Provider: oasis1...` to bind the message to a specific provider. Although the domain is enough in theory.

runtime-sdk/Cargo.toml
```diff
@@ -7,8 +7,9 @@ license = "Apache-2.0"
 
 [dependencies]
 cbor = { version = "0.5.1", package = "oasis-cbor" }
-oasis-core-runtime = { git = "https://github.com/oasisprotocol/oasis-core", tag = "v25.2" }
-oasis-core-keymanager = { git = "https://github.com/oasisprotocol/oasis-core", tag = "v25.2" }
+# TODO: Update to released version.
```
kostko, line 10: Don't merge until we have a new Oasis Core release which includes the log APIs.

rofl-scheduler/src/serverd/auth.rs
```diff
@@ -0,0 +1,234 @@
...
+        .map_err(|_| Error::BadAuthToken)?;
+
+        Ok(token_data.claims)
+    }
+}
+
+/// Login request.
+#[derive(Debug, Clone, serde::Deserialize)]
+#[serde(tag = "method", content = "data")]
+pub enum AuthLoginRequest {
```
ptrus, line 81: No nonce usage? I gues that would complicate things a bit, as the API would need to provision and temporarily store them? 

For this use-case just relying on expiry might be good enough.

kostko: There are several ways to complicate things with the nonce:
* Introduce a two-step process where the server gets you the nonce that you need to use. This requires an additional API call. You could make it stateless (so it doesn't consume memory) by computing the nonce using `Timestamp || HMAC[NonceKey](Address || Timestamp)` and require that the Timestamp is at most X seconds old. Or just use a small cache.
* Use the on-chain account nonce in some way, so e.g. signing a transaction would invalidate all sign-in messages.

Thoughts?

kostko: > The stateless variant is intriguing, but I don’t think it offers significantly stronger security guarantees

No, it is just about reducing memory usage from the need to store nonces and instead storing the nonce at the client (similar as what JWTs are doing so you don't need to keep state on the server). It also avoids the problem with someone polluting the (small) cache and causing legitimate requests to be evicted before they can be completed.

> requiring a short expiry (not_after field),

The current implementation makes it so that `not_after` (if lower than 6 hours) is respected in the issued JWT, so the whole session then has a shorter validity.

kostko: Actually you are right, it doesn't make much sense to have basically two similar authentication tokens, one for the nonce the other for the session. Maybe we should just change it so that the `not_after` is only taken into account for the initial authentication and the session validity is a separate thing. And then we don't need to deal with the nonce.

rofl-scheduler/src/serverd/auth.rs
```diff
@@ -0,0 +1,228 @@
...
+                body.signer.verify(&ctx, &raw_body, &signature)?;
+
+                Address::from_sigspec(
+                    &SignatureAddressSpec::try_from_pk(&body.signer)
+                        .ok_or(anyhow!("unsupported public key scheme"))?,
+                )
+            }
+        };
+
+        let expiry = Utc::now() + Duration::hours(6);
```
ptrus, line 144: Should this duration be configurable?

rofl-scheduler/src/serverd/auth.rs
```diff
@@ -0,0 +1,228 @@
...
+        parts: &mut http::request::Parts,
+        _state: &S,
+    ) -> Result<Self, Self::Rejection> {
+        // Extract the token from the authorization header.
+        let TypedHeader(Authorization(bearer)) = parts
+            .extract::<TypedHeader<Authorization<Bearer>>>()
+            .await
+            .map_err(|_| Error::BadAuthToken)?;
+
+        // Decode the user data.
```
ptrus, line 66:
```suggestion
        // Validate the token and decode the user data.
```

rofl-scheduler/src/serverd/auth.rs
```diff
@@ -0,0 +1,228 @@
...
+                let ctx = signature::context::get_chain_context_for(STD_AUTH_CONTEXT_BASE);
+                body.signer.verify(&ctx, &raw_body, &signature)?;
+
+                Address::from_sigspec(
+                    &SignatureAddressSpec::try_from_pk(&body.signer)
+                        .ok_or(anyhow!("unsupported public key scheme"))?,
+                )
+            }
+        };
+
```
ptrus, line 143: Should we require that the SIWE/StdAuthBody expiry time is short? (e.g. not_before/not_after times are close around the current time).

rofl-scheduler/src/serverd/tls.rs
```diff
@@ -0,0 +1,304 @@
...
+            self.resolver.set_challenge(Some(Arc::new(certified_key)));
+            challenge.set_ready().await?;
+
+            // Currently we only support a single identifier (domain).
+            break;
+        }
+
+        // Exponentially back off until the order becomes ready or invalid.
+        slog::info!(self.logger, "waiting for order to become ready");
+        let status = order.poll(5, Duration::from_millis(250)).await?;
```
ptrus, line 198: Is this enough in practice? Looking at the impl, it only waits like ~3.75 secs? But it looks like they only wait for this long in their example as well, so might be enough?

rofl-scheduler/src/serverd/auth.rs
ptrus: Yeah

rofl-scheduler/src/serverd/tls.rs
kostko: It seems to be working fine in practice, I used the recommended defaults, but we can increase the number of retries to `6` for an extra 4 seconds before the final attempt?

ptrus (approved):
Looks good to me. Just have the one remaining unaddressed comment.

rofl-scheduler/src/serverd/auth.rs
ptrus: Yeah, that would probably be unusual. I guess it's the client's responsibility not to sign messages with long validity. 👍

PR #2241, opened 2025-06-18
runtime-sdk: Add provider-related endorsement constraints

### LINKED ISSUE #2237: rofl: Add more provider-based endorsement constraints
We should add enclave endorsement constraints which relate to the higher-level concept of providers and machines to allow specifying the following policies:
* Only nodes hosted by ROFL market provider X.
* Only ROFL market machines where X is the admin (this requires https://github.com/oasisprotocol/oasis-core/issues/6158 so nodes can attest that the app is running on a specific machine).

kostko:
Fixes #2237


runtime-sdk/modules/rofl-market/src/policy.rs
```diff
@@ -0,0 +1,123 @@
...
+                }
+                Ok(())
+            }
+            AllowedEndorsement::ProviderInstanceAdmin(expected_admin) => {
+                let pa: ProviderAttestation = cbor::from_slice(
+                    &BASE64_STANDARD
+                        .decode(
+                            metadata
+                                .get(METADATA_KEY_POLICY_PROVIDER_ATTESTATION)
+                                .ok_or(Error::NodeNotAllowed)?,
```
peternose, line 72: Will this be hard to debug since all errors are the same?

runtime-sdk/src/modules/rofl/mod.rs
```diff
@@ -204,6 +199,13 @@ impl<Cfg: Config> Module<Cfg> {
             }
         }
 
+        if body.policy.endorsements.len() > Cfg::MAX_ENDORSEMENT_POLICY_ATOMS {
```
peternose, line 202: Isn't this included in the next check?

runtime-sdk/modules/rofl-market/src/policy.rs
```diff
@@ -0,0 +1,123 @@
...
+        Error,
+    },
+    types::address::Address,
+    Context,
+};
+
+use crate::{state, types::InstanceId};
+
+/// Name of the ROFL app instance metadata key used to store the provider attestation.
+pub const METADATA_KEY_POLICY_PROVIDER_ATTESTATION: &str = "net.oasis.policy.provider";
```
peternose, line 22: Why is this defined twice?
Why is `KEY_POLICY`  in the name?

runtime-sdk/modules/rofl-market/src/policy.rs
```diff
@@ -0,0 +1,123 @@
...
+    /// Address of the provider.
+    pub provider: Address,
+    /// Instance identifier.
+    pub instance: InstanceId,
+}
+
+/// Provider attestation metadata stored in `METADATA_KEY_POLICY_PROVIDER_ATTESTATION` label.
+#[derive(Clone, Debug, Default, cbor::Encode, cbor::Decode)]
+pub struct ProviderAttestation {
+    /// A CBOR-serialized `LabelAttestation`.
```
peternose, line 38: In other structs you used name `attestation` and comment `CBOR-serialized label attestation.`.

runtime-sdk/src/modules/rofl/app/registration.rs
```diff
@@ -191,4 +198,49 @@ where
...
+            .attestation()
+            .attest_labels(AttestLabelsRequest {
+                labels: vec![LABEL_PROVIDER.to_string()],
+            })
+            .await?;
+
+        // Decode the attestation to check if the provider label is set and skip setting
+        // metadata in case it is not.
+        let la: LabelAttestation = cbor::from_slice(&rsp.attestation)?;
+        if la.labels.get(LABEL_PROVIDER) == Some(&String::new()) {
```
peternose, line 218: Why don't we use `contains_key`? Are we sure that the label is always present and cannot be `None`?

runtime-sdk/src/modules/rofl/app/registration.rs
```diff
@@ -143,6 +145,11 @@ where
             }
         };
 
+        // Include provider-specific metadata if available.
+        if let Err(err) = self.collect_provider_metadata(&mut metadata).await {
```
peternose, line 149: [nit] Might be more readable if we renamed this function to `fetch_provider_metadata` and then insert it here, as currently the reader has no idea what is going on with metadata.

runtime-sdk/src/modules/rofl/policy.rs
```diff
@@ -40,6 +65,58 @@ pub enum AllowedEndorsement {
     /// Specific node can endorse the enclave.
     #[cbor(rename = "node")]
     Node(PublicKey),
+    /// Any node from a specific provider can endorse the enclave.
+    #[cbor(rename = "provider")]
+    Provider(Address),
+    /// Any provider instance where the given address is currently the admin.
+    #[cbor(rename = "provider_instance_admin")]
+    ProviderInstanceAdmin(Address),
```
peternose, line 73: Can this be simplified to `Admin` or  `InstanceAdmin`, or are there other admins?

runtime-sdk/src/modules/rofl/policy.rs
```diff
@@ -52,3 +129,190 @@ pub enum FeePolicy {
...
+    #[allow(clippy::too_many_arguments)]
+    fn verify_atoms<C: Context>(
+        ctx: &C,
+        op: EndorsementPolicyOperator,
+        policy: &[Box<AllowedEndorsement>],
+        ect: &EndorsedCapabilityTEE,
+        endorsing_node_id: PublicKey,
+        endorsing_node: &Option<Node>,
+        metadata: &BTreeMap<String, String>,
+        max_depth: usize,
```
peternose, line 189: We already count atoms, so why is this useful? Because of `And` and `Or` I would also check width (e.g. use normal counter).

runtime-sdk/src/modules/rofl/policy.rs
```diff
@@ -52,3 +129,190 @@ pub enum FeePolicy {
...
+            if node.expiration < ctx.epoch() {
+                return Ok(None);
+            }
+
+            Ok(Some(node))
+        }()?;
+
+        Self::verify_atoms(
+            ctx,
+            EndorsementPolicyOperator::Or,
```
peternose, line 167: If it is not important which operator we insert here, I would try to remove it from function signature?

runtime-sdk/src/modules/rofl/policy.rs
```diff
@@ -52,3 +129,190 @@ pub enum FeePolicy {
...
+            policy,
+            ect,
+            endorsing_node_id,
+            &endorsing_node,
+            metadata,
+            4,
```
peternose, line 173: I would move this to a constant.

runtime-sdk/src/modules/rofl/policy.rs
kostko: Why is what useful, depth? Defense in depth against too deep stacks. If atom count is 32 you could create policies which go 32 levels deep which I don't think is desired.

runtime-sdk/src/modules/rofl/policy.rs
kostko: It is not important in this particular invocation (because the default for outer-most atoms is OR), but that function is also called from other places where this can be AND (e.g. when evaluating the And operator).

runtime-sdk/modules/rofl-market/src/policy.rs
kostko: That's a good point, but the problem is that this can be part of a more complex condition and it would only be a single basic atom failing. So even if this fails, the overall policy may still evaluate to success.

runtime-sdk/src/modules/rofl/mod.rs
kostko: Yes, I guess it is just saving some recursion. Now that I think of it, depth could (should?) also be checked here.

peternose: Agree, first check depth, then atoms. Could also improve in such a way that this check stops once depth or counter is exceeded (this way you always count at most 32 atoms).

runtime-sdk/modules/rofl-market/src/policy.rs
kostko: See #2265.

runtime-sdk/src/modules/rofl/app/registration.rs
peternose: If the node is corrupted, it could not attest an empty label and in that case this could be `None`, right?

peternose: You are right, but I would still check, so that it is in line with the comment `// Decode the attestation to check if the provider label is set and skip setting metadata in case it is not.`.

peternose (approved):

runtime-sdk/src/modules/rofl/app/registration.rs
kostko: Fixed.

PR #2265, opened 2025-07-11
rofl-app-core: Extract ROFL app core to its own crate

### LINKED ISSUE #2270: Extract ROFL app core to its own crate
Currently the ROFL app core library is part of the Runtime SDK which prevents it from importing other external modules like the `roflmarket`. It would make more sense to refactor it into a new crate.

kostko:
Fixes #2270


tests/runtimes/components-rofl/src/main.rs
```diff
@@ -1,11 +1,11 @@
-use oasis_runtime_sdk::modules::rofl::app::prelude::*;
+use rofl_app_core::prelude::*;
 
 struct TestApp;
 
 #[async_trait]
 impl App for TestApp {
     /// Application version.
-    const VERSION: Version = sdk::version_from_cargo!();
+    const VERSION: Version = Version::new(0, 2, 0);
```
peternose, line 8: Why 0.2.0?

peternose (approved):

tests/runtimes/components-rofl/src/main.rs
kostko: Because 0.2.0 is set in `Cargo.toml` so this is the same as it was before without the extra import.

peternose: Just wondering as I would expect that version is not important in tests, so 0.0.0 or 0.0.1, 0.1.0, 1.0.0 are expected/normal, but 0.2.0 must mean something.

PR #2266, opened 2025-07-11
rofl-app-core: Retry registration using exponential backoff

### LINKED ISSUE #2232: Add exponential backoff for failing registrations
Currently the ROFL app will try to re-register on each round in case the registration fails. It should instead use exponential backoff.

kostko:
Fixes #2232

### DIFF
--- a/rofl-app-core/src/registration.rs
+++ b/rofl-app-core/src/registration.rs
@@ -39,7 +39,6 @@ where
             env,
             logger: get_logger("modules/rofl/app/registration"),
             notify: rx,
-            last_registration_epoch: None,
         };
 
         Self { imp: Some(imp), tx }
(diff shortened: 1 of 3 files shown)


rofl-app-core/src/registration.rs
```diff
@@ -80,20 +78,39 @@ where
...
+                        "err" => ?err,
+                    );
+                }
+
+                result.map_err(backoff::Error::transient)
+            })
+            .await;
+
+            match result {
+                Ok(epoch) => last_registration_epoch = Some(epoch),
```
peternose, line 99: Do we need to clear the `notify` buffer after successful registration, as otherwise we will register twice in a row?

kostko: Why, the `refresh_registration` method should check before registering again?

peternose: Did know that was implemented in `refresh_registration`. Looks fine this way.

peternose (approved):

PR #2269, opened 2025-07-13
runtime-sdk/modules/rofl-market: Add InstanceChangeAdmin call

### LINKED ISSUE #2233: roflmarket: Support changing the instance admin
Currently the instance admin is fixed at creation time and cannot be changed. We should add a call that allows the existing admin to transfer the instance to a different admin.

kostko:
Fixes #2233

### DIFF
--- a/client-sdk/go/modules/roflmarket/roflmarket.go
+++ b/client-sdk/go/modules/roflmarket/roflmarket.go
@@ -17,6 +17,7 @@ var (
 	methodProviderUpdateOffers = types.NewMethodName("roflmarket.ProviderUpdateOffers", ProviderUpdateOffers{})
 	methodProviderRemove       = types.NewMethodName("roflmarket.ProviderRemove", ProviderRemove{})
 	methodInstanceCreate       = types.NewMethodName("roflmarket.InstanceCreate", InstanceCreate{})
+	methodInstanceChangeAdmin  = types.NewMethodName("roflmarket.InstanceChangeAdmin", InstanceChangeAdmin{})
 	methodInstanceTopUp        = types.NewMethodName("roflmarket.InstanceTopUp", InstanceTopUp{})
 	methodInstanceCancel       = types.NewMethodName("roflmarket.InstanceCancel", InstanceCancel{})
 	methodInstanceExecuteCmds  = types.NewMethodName("roflmarket.InstanceExecuteCmds", InstanceExecuteCmds{})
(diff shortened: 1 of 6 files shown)


runtime-sdk/modules/rofl-market/src/types.rs
```diff
@@ -411,6 +411,17 @@ pub struct InstanceCreate {
     pub term_count: u64,
 }
 
+/// A request by the existing instance admin to change the admin.
+#[derive(Clone, Debug, Default, cbor::Encode, cbor::Decode)]
+pub struct InstanceTransferAdmin {
```
peternose, line 416: Would name `InstanceChangeAdmin` be more appropriate, as you are changing admin, but transferring admin rights?

runtime-sdk/modules/rofl-market/src/test.rs
```diff
@@ -768,6 +768,84 @@ fn test_instance_management() {
     assert_eq!(cmds[1].id, 2.into());
     assert_eq!(&cmds[1].cmd, b"and the last one.");
 
+    // Transfer instance to a new admin and then back again.
```
peternose, line 771: Just mentioning that we should probably avoid having so long tests, for multiple reasons:
- It is simple to verify these changes, but very hard to find latter which part is testing admin transfer functionality.
- One might assume that he needs to know everything that was going on up to this point.
- ...

client-sdk/go/modules/roflmarket/roflmarket.go
```diff
@@ -303,6 +304,11 @@ func NewInstanceCreateTx(fee *types.Fee, body *InstanceCreate) *types.Transactio
 	return types.NewTransaction(fee, methodInstanceCreate, body)
 }
 
+// NewInstanceTransferAdmin generates a new roflmarket.InstanceTransferAdmin transaction.
```
peternose, line 307: How can one figure out which methods must be called by admin?  Should that be in the comments?

peternose (approved):

runtime-sdk/modules/rofl-market/src/types.rs
kostko: Yeah good point, will rename the method.

runtime-sdk/modules/rofl-market/src/test.rs
kostko: I agree and was wondering about the same thing when adding this test. Will move into a separate test.

client-sdk/go/modules/roflmarket/roflmarket.go
kostko: Added.

PR #2278, opened 2025-07-17
ts-web/rt: make typescript warn on extraneous fields

lukaw3d:


pro-wh:
has this been happening? why these specific type mistakes?

lukaw3d:
I mainly care about `callUpdate().setBody({...app, <changes> })` because it wasted my hours. This is the most intuitive way to update something, but it doesn't work. We use it through subcalls in rofl.app and subcalls [currently](https://github.com/oasisprotocol/nexus/pull/1099) don't show when they fail.

Others: likely to be used in subcalls and have decent overlap with another type

kostko:
Omg where is the type safety, this looks like an ugly hack...

kostko:
What do you mean the network makes them fail silently? If they are proper calls in transactions incorrect structures will _not_ fail silently.

If you are using EVM smart contracts, then it follows EVM semantics and the subcall does indicate errors and it is up to the caller to handle them.

I do not see how this is the runtime's fault and what could be done better? I assume you could introduce a precompile that would be more suited to direct calls, but not sure if it is worth it.

lukaw3d:
@kostko I mostly meant that nexus+explorer were not showing an error, nor results data

grpc: It is possible to get status code with runtimeClientGetTransactionsWithResults + abi decoding result
json-rpc: I don't think there's a way

PR #2297, opened 2025-08-17
rofl-containers: Perform additional LUKS2 header validation

kostko:


rofl-containers/src/storage/mod.rs
```diff
@@ -76,21 +81,31 @@ pub async fn init(kms: Arc<dyn services::kms::KmsService>) -> Result<()> {
...
+    run_cmd!(cryptsetup luksHeaderBackup --header-backup-file "/tmp/storage-luks2-header" --disable-locks "/dev/mapper/part-storage")?;
+    // Dump and validate header.
+    let header_json = run_fun!(cryptsetup luksDump --type luks2 --dump-json-metadata "/tmp/storage-luks2-header")?;
+    luks2::validate_header(&header_json).context("failed to validate LUKS2 header")?;
+
+    // Attempt to open the storage device using a validated header.
     run_cmd!(
         echo -n ${storage_key} |
-            cryptsetup open --type luks2 --disable-locks "/dev/mapper/part-storage" storage
+            cryptsetup open --type luks2 --disable-locks --header "/tmp/storage-luks2-header" "/dev/mapper/part-storage" storage
```
peternose, line 93: If the primary metadata is corrupted, can we run into a problem when this tries to fall back to the secondary copy? I guess we only verify primary.

kostko: AFAIK using `--header` only uses the specified detached header and does not load the header from the block device itself.

kostko: Just checked the cryptsetup source code again and this is indeed the case.

PR #2312, opened 2025-09-02
rofl-scheduler: Add support for machine permissions

### LINKED ISSUE #2305: rofl-scheduler: Support delegated authentication
Allow delegated access to machines, where the policy is configured via deployment metadata. To start, one should be able to allow a different account to access logs by setting `net.oasis.scheduler.permissions` to something like (CBOR-encoded):
```json
{
  "logs.view": ["address1", "address2"],
}
```

kostko:
Fixes #2305

### DIFF
--- a/rofl-app-core/src/secrets.rs
+++ b/rofl-app-core/src/secrets.rs
@@ -233,9 +233,6 @@ mod test {
             "a462706b582077e51262274c9505a89db23b4921105d4ea1d8bd49bc99087281f3f92c812144646e616d65533fd8dabddfaf5cf0d1505f9ea0c090d241ed71656e6f6e63654f49f7baede65e2b5e930fb0d3f616fe6576616c756555bd080af9cb24129afafc5c461d231385c22ebb83b5",
         )];
 
-        let k1 = x25519::PrivateKey::generate();
-        let k2 = x25519::PrivateKey::generate();
-
         for (idx, tv) in tvs.iter().enumerate() {
             let sk1: [u8; x25519::PRIVATE_KEY_LENGTH] =
                 hex::decode(tv.0).unwrap().try_into().unwrap();
--- a/rofl-scheduler/src/state.rs
+++ b/rofl-scheduler/src/state.rs
@@ -1,0 +1,124 @@
+use std::{collections::BTreeMap, time::Instant};
+
+use base64::{prelude::BASE64_STANDARD, Engine};
+use oasis_runtime_sdk::types::address::Address;
+use oasis_runtime_sdk_rofl_market::types::{Deployment, Instance};
+
+use crate::types::{Permissions, METADATA_KEY_PERMISSIONS};
+
+/// Instance state.
+#[derive(Debug, Clone)]
+pub struct InstanceState {
+    /// Address of the instance administrator.
+    admin: Address,
+    /// Permissions.
+    permissions: Permissions,
+    /// Last deployment.
+    pub last_deployment: Option<Deployment>,
+    /// Last error message corresponding to deploying `last_deployment`.
+    pub last_error: Option<String>,
+    // Whether to ignore instance start until the given time elapses.
+    pub ignore_start_until: Option<Instant>,
+    /// Backoff associated with ignoring instance start.
+    pub ignore_start_backoff: Option<backoff::ExponentialBackoff>,
+}
+
+impl InstanceState {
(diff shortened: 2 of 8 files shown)


rofl-app-core/src/secrets.rs
```diff
@@ -233,9 +233,6 @@ mod test {
             "a462706b582077e51262274c9505a89db23b4921105d4ea1d8bd49bc99087281f3f92c812144646e616d65533fd8dabddfaf5cf0d1505f9ea0c090d241ed71656e6f6e63654f49f7baede65e2b5e930fb0d3f616fe6576616c756555bd080af9cb24129afafc5c461d231385c22ebb83b5",
         )];
 
-        let k1 = x25519::PrivateKey::generate();
```
ptrus, line 236: Was this variable not used used? Shouldn't `clippy` catch this? Or do we not run it?

rofl-scheduler/src/types.rs
```diff
@@ -39,3 +42,35 @@ pub struct TerminateRequest {
...
 }
+
+/// Name of the metadata key used to store the delegated permissions.
+pub const METADATA_KEY_PERMISSIONS: &str = "net.oasis.scheduler.permissions";
+
+/// Name of the action for viewing machine logs.
+pub const ACTION_LOG_VIEW: &str = "log.view";
+
+/// Instance permissions for different actions.
+pub type Permissions = BTreeMap<String, BTreeSet<Address>>;
```
ptrus, line 53: Should we also introduce an enum type for permission names? Or is there an usecase for these to be arbitrary strings so that rofl apps could also use it for application specific permissions?

If we support arbitrary strings, we should probably have some limits on the key sizes? Maybe also on the number of addresses in the value?

rofl-scheduler/src/state.rs
```diff
@@ -0,0 +1,99 @@
...
+    pub last_error: Option<String>,
+    // Whether to ignore instance start until the given time elapses.
+    pub ignore_start_until: Option<Instant>,
+    /// Backoff associated with ignoring instance start.
+    pub ignore_start_backoff: Option<backoff::ExponentialBackoff>,
+}
+
+impl InstanceState {
+    /// Updates instance state from the given instance and deployment.
+    pub fn update_from(&mut self, instance: &Instance, deployment: Option<&Deployment>) {
```
ptrus, line 28: Could we split this into two methods, `update_admin` and `update_permissions`? Or rename the `update_from` to something less general, since the method name does not clearly indicate what is getting updated.

rofl-scheduler/src/types.rs
ptrus: It's fine then 👍

ptrus (approved):

rofl-scheduler/src/state.rs
```diff
@@ -0,0 +1,103 @@
...
+
+        match self.permissions.get(action) {
+            Some(addresses) => addresses.contains(&address),
+            None => false,
+        }
+    }
+
+    /// Parse permissions from deployment metadata.
+    fn parse_permissions(metadata: &BTreeMap<String, String>) -> Option<Permissions> {
+        let permissions = metadata.get(METADATA_KEY_PERMISSIONS)?;
```
peternose, line 54: I prefer having this somewhere else and change signature to `update_permissions(&mut self, permissions: Vec<u8>)` (maybe 2 methods, one for raw permissions and one for normal), as the state shouldn't care how permissions are stored. This way, the test would also simplify.

rofl-scheduler/src/state.rs
```diff
@@ -0,0 +1,103 @@
...
+use oasis_runtime_sdk::types::address::Address;
+use oasis_runtime_sdk_rofl_market::types::{Deployment, Instance};
+
+use crate::types::{Permissions, METADATA_KEY_PERMISSIONS};
+
+/// Instance state.
+#[derive(Default, Debug, Clone)]
+pub struct InstanceState {
+    /// Address of the instance administrator.
+    admin: Address,
```
peternose, line 13: It is a bit strange to have a default admin address. I prefer having a constructor with admin and permissions as parameters.

rofl-scheduler/src/state.rs
kostko: Changed a bit. I don't want to have permission extraction/handling logic in the caller, this was the main reason for moving state into its own module.

PR #2335, opened 2025-09-19
feat: format known addresses for consensus txes

uniyalabhishek:


client-sdk/go/types/address.go
```diff
@@ -44,6 +44,11 @@ type contextKey string
 // ContextKeyAccountNames is the key to retrieve the public key to account name map from context.
 const ContextKeyAccountNames = contextKey("runtime/account-names")
 
+// ContextKeyAccountEthMap is the key to retrieve a mapping from native (Bech32)
+// Oasis addresses to their corresponding Ethereum hex addresses for pretty printing.
+// When present, Ethereum addresses take precedence inside parentheses.
+const ContextKeyAccountEthMap = contextKey("runtime/account-eth-map")
```
peternose, line 50: Where is this set?

client-sdk/go/types/token.go
```diff
@@ -87,18 +87,38 @@ func NewBaseUnits(amount quantity.Quantity, denomination Denomination) BaseUnits
 	}
 }
 
+// FormatNamedAddress returns a human-friendly representation of an address:
+// it prints the name (if known) followed by the preferred form of the address
+// in parentheses. If an Ethereum hex address mapping is provided for the native
+// address, it is used; otherwise the native Bech32 address is used.
+func FormatNamedAddress(ctx context.Context, addr Address) string {
```
peternose, line 94:
```suggestion

// FormatNamedAddress returns a human-friendly representation of an address.
//
// It prints the name (if known) followed by the preferred form of the address
// in parentheses. If an Ethereum hex address mapping is provided for the native
// address, it is used; otherwise the native Bech32 address is used.
func FormatNamedAddress(ctx context.Context, addr Address) string {
	native := addr.String()

	name, ok := getAccountName(ctx, native)
	if !ok {
		return native
	}

	hex, ok := getEthereumHex(ctx, native)
	if !ok {
		return fmt.Sprintf("%s (%s)", name, native)
	}

	return fmt.Sprintf("%s (%s)", name, hex)
}

func getAccountName(ctx context.Context, addr string) (string, bool) {
	names, ok := ctx.Value(ContextKeyAccountNames).(AccountNames)
	if !ok {
		return "", false
	}
	name, ok := names[addr]
	return name, ok
}

func getEthereumHex(ctx context.Context, addr string) (string, bool) {
	addrs, ok := ctx.Value(ContextKeyAccountEthMap).(map[string]string)
	if !ok {
		return "", false
	}
	hex, ok := addrs[addr]
	return hex, ok
}
```
I prefer always using `ok`, not `exists`, `has`, `found`, ...

peternose, line 94: I think that we cannot get rid of the `ctx` here, but I will still comment that fetching things from a `ctx` should be avoided as much as possible, for multiple reasons:
- it is harder to unit test this function,
- the function signature is misleading as things are hidden in the context.

Ideally, this function should actually accept two maps and an address. Or even better, a struct should accept maps, and a method should accept an address. 

Can we add a unit test for this?

client-sdk/go/types/address.go
uniyalabhishek: Updated the PR, please review again. This is set by callers constructing a pretty-print context (oasis-cli). The SDK only defines the context key + formatting helper, if the map isn’t present, formatting falls back to the native Bech32 address.

matevz (approved):
I think this is good to go now.

PR #2341, opened 2025-09-25
rofl-client-py: Prepare 0.1.4 release

rube-de:
- fix root for hatch, that hatch can see git tags
(was before not seeing tags and falling back to 0.0.0)

(works now correctly on my [fork](https://github.com/rube-de/oasis-sdk/actions/runs/18011169927))


kostko:
Tagged.

Issue #2234, opened 2025-06-09
rofl-containers: Allow containers to easily set instance metadata


kostko:
There should be an easy way for containers to set instance metadata via the appd API.

PR #2356, opened 2025-10-20
rofl-containers: Allow containers to set instance metadata

ptrus:
Closes: https://github.com/oasisprotocol/oasis-sdk/issues/2234

Also see: https://github.com/oasisprotocol/demo-rofl/pull/16

### DIFF
--- a/docs/rofl/features/appd.md
+++ b/docs/rofl/features/appd.md
@@ -147,7 +147,7 @@ For example:
     "data": "a1626f6b40"
   }
   ```
-  
+
   deserialized as `{"ok": ''}`.
 
 - Unsusccessful call result:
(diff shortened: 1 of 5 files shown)


rofl-app-core/src/registration.rs
```diff
@@ -119,9 +127,26 @@ where
...
+        let mut metadata = match self.state.app.clone().get_metadata(self.env.clone()).await {
+            Ok(metadata) => metadata,
+            Err(err) => {
+                slog::error!(self.logger, "failed to get instance metadata"; "err" => ?err);
+                // Do not prevent registration, just clear metadata.
+                Default::default()
+            }
+        };
+        // Include provider-specific metadata if available.
+        // TODO: if this is expensive, we could only check the user-provided metadata.
```
kostko: Not expensive.

rofl-appd/src/services/metadata.rs
```diff
@@ -0,0 +1,106 @@
...
+pub struct OasisMetadataService<A: App> {
+    env: Environment<A>,
+    metadata: RwLock<BTreeMap<String, String>>,
+}
+
+impl<A: App> OasisMetadataService<A> {
+    /// Create a new metadata service, initializing from existing on-chain registration.
+    pub async fn new(env: Environment<A>) -> Result<Self, Error> {
+        // Fetch existing metadata from on-chain registration. This ensures that if the app
+        // restarts, previously set metadata is preserved rather than being reset to empty.
```
kostko: But if the entire app restarts it will have a different RAK, thus the metadata will be empty anyway (since metadata is per app instance, not per app)?

docs/rofl/features/appd.md
```diff
@@ -162,3 +162,43 @@ For example:
   `{"fail": {"code": 8, "module": "evm", "message": "reverted: aJAoLw=="}}`.
 
 [call result]: https://api.docs.oasis.io/rust/oasis_runtime_sdk/types/transaction/enum.CallResult.html
+
+## Instance Metadata
```
matevz: I would avoid term "instance", because it can describe both a ROFL replica or a ROFL app.

```suggestion
## App Metadata
```

docs/rofl/features/appd.md
```diff
@@ -162,3 +162,43 @@ For example:
...
+
+```json
+{
+  "key_fingerprint": "a54027bff15a8726",
+  "version": "1.0.0"
+}
+```
+
+**Note:** Metadata is validated against runtime-configured limits for the
+number of pairs, key size, and value size.
```
matevz: I would also add the "Remove Metadata" endpoint using the HTTP DELETE request. Similar to how the env variables can be set as empty string or not set at all.

kostko (approved):

docs/rofl/features/appd.md
ptrus: The POST metadata overrides existing metadata, so this can be done by POSTing an empty map. Not sure if we need to add a HTTP DELETE request endpoint to `rofl-appd`.  The client can expose a delete method (and just call POST `{}`) to achieve the same thing for the enduser.

rofl-containers/src/main.rs
```diff
@@ -60,6 +64,20 @@ impl App for ContainersApp {
...
             meta.extend(identity.metadata());
         }
+
+        // Get user-provided metadata from the service.
+        if let Some(metadata_service) = self.metadata.get() {
+            metadata_service.get().await.map(|m| {
+                // Namespace user-provided metadata.
+                meta.extend(m.into_iter().map(|(k, v)| {
+                    (
+                        format!("{}.{}", services::metadata::METADATA_NAMESPACE, k),
```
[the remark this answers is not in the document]
ptrus: Value size is limited, so using a map is not really feasible. We (the ROFL developers) can ensure that no other internal ROFL system will ever use this namespace, so there’s no risk of collision in practice.

Issue #2369, opened 2025-11-03
Reuse the same ACME account for ROFL app to avoid rate limiting from Let's Encrypt


anzoman:
Recently, we faced some problems after deploying ROFL apps (e.g., ROFL scheduler) when the app failed to obtain SSL/TLS certificates due to rate limiting from Let's Encrypt.

For example:

```console
{"component":"rofl.<rofl_app_id>","err":"Api(Problem { type: Some(\"urn:ietf:params:acme:error:rateLimited\"), detail: Some(\"too many new registrations (10) from this IP address in the last 3h0m0s, retry after 2025-11-03 10:23:08 UTC: see https://letsencrypt.org/docs/rate-limits/#new-registrations-per-ip-address\"), status: Some(429), subproblems: [] })","level":"error","module":"runtime/serverd/cert-provisioner","msg":"failed to initialize ACME account","provisioner":"tdx-qemu","runtime_id":"000000000000000000000000000000000000000000000000a6d1e3ebf60dff6c","runtime_name":"","ts":"2025-11-03T10:11:05.509202076Z"}
```

where the IP address has created more than 10 new ACME accounts in 3 hours, which exceeds [Let’s Encrypt’s rate limit](https://letsencrypt.org/docs/rate-limits/#new-registrations-per-ip-address).

The temporary workaround is to stop the app, wait for the latest "retry-at" timestamp, and start it again.

To avoid this problem, the app could reuse the same ACME account instead of creating a new ACME registration for new certificate issuance.

ptrus:
The solution is to address this TODO: https://github.com/oasisprotocol/oasis-sdk/blob/5ae2ca36b56a5c1bc222c82c70adbca478f6c725/rofl-proxy/src/http/tls.rs#L163

PR #2370, opened 2025-11-03
rofl-appd: Provision ACME account credentials via KMS

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-sdk/issues/2369

TODO:
- [ ] test

### DIFF
--- a/rofl-containers/src/proxy/mod.rs
+++ b/rofl-containers/src/proxy/mod.rs
@@ -94,7 +94,7 @@ async fn maybe_start<A: App>(
 
     slog::info!(logger, "proxy configuration is available, starting proxy");
     tokio::spawn(async move {
-        if let Err(err) = run(proxy_label, compose).await {
+        if let Err(err) = run(proxy_label, compose, kms).await {
             slog::error!(logger, "failed to start proxy"; "err" => ?err);
         }
     });
@@ -102,7 +102,11 @@ async fn maybe_start<A: App>(
     Ok(())
 }
 
-async fn run(proxy_label: ProxyLabel, compose: ParsedCompose) -> Result<()> {
+async fn run(
+    proxy_label: ProxyLabel,
+    compose: ParsedCompose,
+    kms: Arc<dyn services::kms::KmsService>,
+) -> Result<()> {
     let logger = get_logger("proxy");
 
     // We only need the IP address part of the IP/CIDR format.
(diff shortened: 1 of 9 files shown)


rofl-proxy/src/http/tls.rs
```diff
@@ -207,6 +188,81 @@ impl CertificateProvisioner {
...
+    /// Try to provision a new acme account.
+    async fn try_provision_acme(&self) -> Result<instant_acme::Account> {
+        let (acme, creds) = instant_acme::Account::builder()?
+            .create(
+                &instant_acme::NewAccount {
+                    contact: &[],
+                    terms_of_service_agreed: true,
+                    only_return_existing: false,
+                },
+                instant_acme::LetsEncrypt::Production.url().to_owned(),
```
kostko: Maybe we should allow one to configure "staging" on Testnet or something? That has much lower rate limits.

kostko (commented):
Mostly looks good to me.

Haven't checked, but is it possible to initialize an ACME account using just a key? That way we could use the ROFL KMS to generate a deterministic key for an ACME account without the need for separate persistent storage. But that introduces additional dependencies so let's keep it simple for now.

rofl-containers/src/proxy/mod.rs
```diff
@@ -146,13 +150,36 @@ async fn run(proxy_label: ProxyLabel, compose: ParsedCompose) -> Result<()> {
...
-        listen_port: PROXY_HTTPS_LISTEN_PORT,
-        ..Default::default()
-    })
+    let mut http = http::Proxy::new(
+        http::Config {
+            listen_address: listen_address.clone(),
+            listen_port: PROXY_HTTPS_LISTEN_PORT,
+            ..Default::default()
+        },
+        acme,
```
kostko: Could put it into config?

rofl-scheduler/src/serverd/mod.rs
```diff
@@ -41,7 +41,30 @@ pub struct State {
 
 /// Start the server endpoint in a background task.
 pub async fn serve(cfg: Config<'_>) -> Result<()> {
-    let tls_provisioner = tls::CertificateProvisioner::new();
+    // Create ACME account from ROFL KMS.
```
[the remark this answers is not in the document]
ptrus: >  I guess the reason is that we don't initialize the ROFL KMS service here.

It would also simplify the `init_acme_account` since it could accept in the KMS directly and wouldn't need the `get_raw_key` closure.  But if want the Node scope that's currently not supported in the KMS service right?

> Would it be possible to generate the key only once to avoid an extra transaction

Yes of-course, fixed. Previous version was silly.

> And also we should use the node scope (scope: KeyScope::Node)

Added.

PR #2371, opened 2025-11-04
feat(rofl-client/rs): add Rust ROFL client 

ZigaMr:
This PR adds a new Rust crate providing an async client for ROFL appd over Unix domain sockets. Includes core APIs for app id, key generation, and transaction sign+submit, along with docs and an example.

### DIFF
--- a/rofl-client/rs/src/lib.rs
+++ b/rofl-client/rs/src/lib.rs
@@ -176,14 +176,14 @@ impl RoflClient {
         &self,
         gas_limit: u64,
         to: &str,
-        value: u64,
+        value: &str,
         data_hex: &str,
         encrypt: Option<bool>,
     ) -> Result<String, Box<dyn std::error::Error>> {
         let eth = EthCall {
             gas_limit,
             to: to.to_string(),
-            value,
+            value: value.to_string(),
             data: data_hex.to_string(),
         };
         self.sign_submit(Tx::Eth(eth), encrypt).await
@@ -312,7 +312,7 @@ pub enum Tx {
 pub struct EthCall {
     pub gas_limit: u64,
     pub to: String,
-    pub value: u64,
+    pub value: String,
     pub data: String, // hex string without 0x prefix
 }
 


rofl-client/rs/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+    let mut buf = [0u8; 8192];
+    loop {
+        let n = stream.read(&mut buf)?;
+        if n == 0 {
+            break;
+        }
+        resp.extend_from_slice(&buf[..n]);
+    }
+
+    let header_end = resp
```
njelich: Would it be a bit easier to maintain if we used something like https://crates.io/crates/httparse?

rofl-client/rs/README.md
```diff
@@ -0,0 +1,32 @@
...
+# #[tokio::main]
+# async fn main() -> Result<(), Box<dyn std::error::Error>> {
+let client = RoflClient::new()?; // or with_socket_path("/custom.sock")
+
+let app_id = client.get_app_id().await?;
+let key = client.generate_key("my-key", KeyKind::Secp256k1).await?;
+
+// ETH-style call
+let result = client
+    .sign_submit_eth(21_000, "0xdeadbeef...", 0, "a9059cbb...", None)
```
matevz: Ideally the quickstart snippet here should be just copy&pasted from `examples/quickstart.rs` so it's tested.

```suggestion
    .sign_submit(21_000, "0xdeadbeef...", 0, "a9059cbb...", None)
```

rofl-client/rs/src/lib.rs
```diff
@@ -0,0 +1,449 @@
...
+    Eth(EthCall),
+    #[serde(rename = "std")]
+    Std(String), // CBOR-serialized hex-encoded Transaction
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct EthCall {
+    pub gas_limit: u64,
+    pub to: String,
+    pub value: u64,
```
uniyalabhishek: Consider changing to `String` for consistency with Python/TS clients, which normalize all values to decimal strings to support amounts > u64::MAX. For context https://github.com/oasisprotocol/oasis-sdk/pull/2416#discussion_r2627631517

Issue #2375, opened 2025-11-05
Fix codecov rate limiting


ptrus:
https://github.com/oasisprotocol/oasis-sdk/actions/runs/19103003299/job/54579411086?pr=2370

> [2025-11-05T13:19:33.060Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 429 - {"message":"Rate limit reached. Please upload with the Codecov repository upload token to resolve issue. Expected time to availability: 2172s."}

PR #2380, opened 2025-11-06
ci: Login to codecov to fix rate-limiting

lukaw3d:
Does this fix https://github.com/oasisprotocol/oasis-sdk/issues/2375?


ptrus:
Thanks, merging. I'll try to fix the disk space issue in a separate PR.

PR #2383, opened 2025-11-06
rofl-scheduler: Bump memory to 1GB

ptrus:
Also includes id's from https://github.com/oasisprotocol/oasis-sdk/pull/2374

Please update the testnet policies @kostko when available


ptrus:
👍

Issue #2394, opened 2025-11-10
rofl-client/ts: Fix sign-submit example


rube-de:
The current sign-submit example makes an empty call with no value.

```
        // Sign & submit (ETH style)
        const result = await client.signAndSubmit({
            kind: 'eth',
            gas_limit: 200_000,
            to: '', // empty => contract creation
            value: 0,
            data: '0x', // no-op calldata
        });
```

A basic example would be just sending native currency (with `to` and `value`).
Advanced examples like contract calls and contract creation could be in a demo example.

PR #2393, opened 2025-11-10
feat: update rofl-client ts to include queries endpoint

uniyalabhishek:
This PR addresses 3 follow-up feedback from the initial `rofl-client/ts` sdk [PR](https://github.com/oasisprotocol/oasis-sdk/pull/2365)

1. add support for the recently added queries endpoint and use the types for the arguments and responses by @ptrus https://github.com/oasisprotocol/oasis-sdk/pull/2365#discussion_r2491114558
2. update transaction handling by @rube-de https://github.com/oasisprotocol/oasis-sdk/pull/2365#discussion_r2494702956
3. support ESM and target ES2022 by @rube-de https://github.com/oasisprotocol/oasis-sdk/pull/2365#discussion_r2494750912

PR #2398, opened 2025-11-12
fix: update txn examples README for ETH-style transfers of ROSE

uniyalabhishek:
closes https://github.com/oasisprotocol/oasis-sdk/issues/2394

### DIFF
--- a/rofl-client/ts/README.md
+++ b/rofl-client/ts/README.md
@@ -28,10 +28,10 @@ const appId = await client.getAppId(); // bech32 string (helper)
 // Sign & submit an authenticated transaction (ETH-style)
 const callResultBytes = await client.signAndSubmit({
   kind: 'eth',
-  gas_limit: 200_000,
-  to: '',                    // empty => contract creation
-  value: '0',                // decimal or 0x hex string
-  data: '0x',                // hex calldata (0x optional)
+  gas_limit: 21_000,
+  to: '0x1234845aaB7b6CD88c7fAd9E9E1cf07638805b20',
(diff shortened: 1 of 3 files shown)


rofl-client/ts/README.md
```diff
@@ -88,7 +89,10 @@ type EthTx = {
   gas_limit: number;
   /** Hex address (0x optional). Empty string => contract creation. */
   to: string;
-  /** JSON number; must fit JS number range (backend expects u128). */
+  /**
+   * JSON number; must fit JS number range (backend expects u128).
```
lukaw3d: what does this mean "must fit JS number range"?

rofl-client/ts/README.md
```diff
@@ -25,13 +25,14 @@ const metadata = await client.getMetadata();
 
 const appId = await client.getAppId(); // bech32 string (helper)
 
-// Sign & submit an authenticated transaction (ETH-style)
+// Sign & submit an authenticated transaction (ETH-style): transfer ROSE
+// NOTE: 'value' uses 18-decimals on EVM ParaTimes (wei-like). Example sends 0.001 ROSE.
```
rube-de: You have below a commend at value, I don't think you need to repeat this for general sign-submit. Also because most of the time sign-submit will be a contract call without value.

rofl-client/ts/README.md
uniyalabhishek: Yes, have the fix in this PR https://github.com/oasisprotocol/oasis-sdk/pull/2393. will rebase and update this PR once that's merged

PR #2400, opened 2025-11-13
rofl-client-py: Make methods sync

matevz:
This simplifies usage of our ROFL python client by adding non-async version of `RoflClient`:
```python
from oasis_rofl_client import RoflClient

client = RoflClient()

key = client.generate_key("my-key-id")
print(f"Generated SECP256K1 key: {key}")
```

Async version was moved to `AsyncRoflClient`:
```python
import asyncio
from oasis_rofl_client import AsyncRoflClient

async def main():
    client = AsyncRoflClient()

    key = await client.generate_key("my-key-id")
    print(f"Generated SECP256K1 key: {key}")

asyncio.run(main())
```

**This is a breaking change!**

### DIFF
--- a/rofl-client/py/src/oasis_rofl_client/__init__.py
+++ b/rofl-client/py/src/oasis_rofl_client/__init__.py
@@ -5,10 +5,13 @@ Oasis ROFL services, including key generation and transaction submission.
 """
 
 __all__ = [
+    "AsyncRoflClient",
     "KeyKind",
     "RoflClient",
     "__version__",
 ]
 
 from .__about__ import __version__
-from .rofl_client import KeyKind, RoflClient
+from .async_rofl_client import AsyncRoflClient
+from .common import KeyKind
+from .rofl_client import RoflClient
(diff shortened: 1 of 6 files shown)


rofl-client/py/tests/test_client.py
matevz: While I don't like redundant code, I thought it's a good idea to have both `test_async_client.py` and `test_client.py`  as long as we also have separate implementations for `RoflClient` and `AsyncRoflClient`. e.g. in case someone forgets to implement a sync version around a new endpoint.

matevz:
Alternative approach: https://github.com/bitcart/universalasync

rofl-client/py/src/oasis_rofl_client/__init__.py
```diff
@@ -5,10 +5,12 @@
...
+    "AsyncRoflClient",
     "KeyKind",
     "RoflClient",
     "__version__",
 ]
 
 from .__about__ import __version__
-from .rofl_client import KeyKind, RoflClient
+from .async_rofl_client import KeyKind
+from .rofl_client import AsyncRoflClient, RoflClient
```
ptrus: Does this work? isn't `AsyncRoflClient` defined in `.async_rofl_client`?

PR #2402, opened 2025-11-14
Update rofl-scheduler to use CLI 0.17.1

ptrus:


ptrus:
Both mainnet and testnet Identities match what I get building locally 👍

PR #2414, opened 2025-12-08
docs: Explicitly use "encrypt" field in appd

lukaw3d:
This field is currently required by mistake


matevz (approved):
We can temporarily add this until rofl-containers 0.8.6 is out.

PR #2416, opened 2025-12-10
fix(rofl-appd): improve transaction field deserialization flexibility

uniyalabhishek:
Fix transaction deserialization issues causing 422 errors in ROFL appd
- Fix value field: Change `u128` to `u64` in `TransactionValue` enum (serde_json doesn't support `u128` in untagged enums). Large values must use string format.
- Fix hex field flexibility: to, data, and gas_limit now accept values with or without 0x prefix, and gas_limit accepts both numbers and strings.

### DIFF
--- a/rofl-appd/src/routes/tx.rs
+++ b/rofl-appd/src/routes/tx.rs
@@ -1,6 +1,7 @@
 use std::{collections::BTreeSet, sync::Arc};
 
 use rocket::{http::Status, serde::json::Json, State};
+use serde::Deserialize;
 use serde_with::serde_as;
 
 use oasis_runtime_sdk::types::transaction;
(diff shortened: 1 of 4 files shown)


rofl-appd/src/routes/tx.rs
```diff
@@ -98,6 +121,68 @@ fn parse_u256_string(value: String) -> Result<evm::types::U256, String> {
...
+    let mut bytes = Vec::with_capacity(bytes_slice.len() / 2);
+    for chunk in bytes_slice.chunks_exact(2) {
+        let hi =
+            hex_nibble(chunk[0]).ok_or_else(|| serde::de::Error::custom("invalid hex string"))?;
+        let lo =
+            hex_nibble(chunk[1]).ok_or_else(|| serde::de::Error::custom("invalid hex string"))?;
+        bytes.push((hi << 4) | lo);
+    }
+
+    Ok(bytes)
```
matevz: I'd rather trim 0x and then use serde's or hex crate to decode to reduce the amount of code.

rofl-appd/src/routes/tx.rs
```diff
@@ -53,21 +54,43 @@ pub enum Transaction {
     /// Ethereum transaction.
     #[serde(rename = "eth")]
     Eth {
-        gas_limit: u64,
-        #[serde_as(as = "serde_with::hex::Hex")]
+        gas_limit: QuantityU64,
```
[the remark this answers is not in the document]
uniyalabhishek: didn't consider it earlier. checked now and `primitive-types` (what we have via evm module) only exposes U128/U256/U512, no U64. also its serde expects hex strings only, not JSON numbers like 21000 which existing callers send

rofl-appd/src/routes/tx.rs
```diff
@@ -53,21 +54,43 @@ pub enum Transaction {
...
+///
+/// Note: The `Number` variant uses `u64` because serde_json does not support `u128` in untagged
+/// enums. For values larger than `u64::MAX`, use the `String` variant with a decimal or hex
+/// representation.
 #[derive(Clone, Debug, serde::Deserialize)]
 #[serde(untagged)]
 pub enum TransactionValue {
     String(String),
-    Number(u128),
+    Number(u64),
```
matevz: Transaction value and gas price must be at least `u128`. You can't fit the circulating supply of the tokens into 64 bits (in wei). e.g. 20B ROSE requires 95 bits.

uniyalabhishek: @matevz String variant already supports values up to `u256`, which handles 20B ROSE in wei, added a test to verify. Number(u64) variant is limited because `serde_json` can't deserialize JSON integers > `u64` into Rust `u128` in untagged enums, clients should use string format for large values (which is also safer since JavaScript's `Number.MAX_SAFE_INTEGER` is only 2^53-1)

rofl-appd/src/routes/tx.rs
```diff
@@ -53,21 +54,43 @@ pub enum Transaction {
...
+        #[serde(deserialize_with = "deserialize_hex_bytes")]
         data: Vec<u8>,
     },
 }
 
+/// A quantity that can be provided either as a JSON number or as a string
+/// (decimal or 0x-prefixed hex) and normalized into a `u64`.
+#[derive(Clone, Debug, serde::Deserialize)]
+#[serde(untagged)]
+pub enum QuantityU64 {
```
ptrus: I would rename this to something more specific, like `GasLimitInput`, since it's only used for that at the moment. Unless we see this being useful elsewhere as well.

rofl-appd/src/routes/tx.rs
```diff
@@ -98,6 +121,45 @@ fn parse_u256_string(value: String) -> Result<evm::types::U256, String> {
...
+{
+    let s = String::deserialize(deserializer)?;
+    let trimmed = s.trim();
+    let without_prefix = trimmed
+        .strip_prefix("0x")
+        .or_else(|| trimmed.strip_prefix("0X"))
+        .unwrap_or(trimmed);
+
+    if without_prefix.is_empty() {
+        return Ok(Vec::new());
```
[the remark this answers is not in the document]
uniyalabhishek: yes now we reject any whitespace in hex fields and also reject prefix-only "0x"/"0X" (use "" for empty bytes)

PR #2452, opened 2026-04-16
runtime-sdk/modules/evm/src/precompile/subcall: Support read-only mode

peternose:


runtime-sdk/modules/evm/src/precompile/confidential.rs
```diff
@@ -396,7 +396,7 @@ mod test {
 
     #[test]
     fn test_x25519_derive() {
-        let mut rng = OsRng {};
+        let rng = OsRng {};
         let static_secret = x25519_dalek::StaticSecret::random_from_rng(rng);
```
kostko: I think you can now also do something like:
```suggestion
        let static_secret = x25519_dalek::StaticSecret::random_from_rng(OsRng);
```

kostko (approved):

PR #2461, opened 2026-05-13
runtime-sdk/modules/rofl-market: Refund overpayment to admin

peternose:


runtime-sdk/modules/rofl-market/src/payment.rs
```diff
@@ -158,78 +148,22 @@ impl PaymentMethod for Payment {
...
-                // Refund.
-                <C::Runtime as Runtime>::Accounts::transfer(
-                    payment_address,
-                    refund_address,
-                    &token::BaseUnits::new(amount, denomination.clone()),
-                )?;
-
-                Ok(())
-            }
-            Self::EvmContract { address, data } => {
```
kostko, line 183: Hm, so now refund only works for non-EVM payment methods?

runtime-sdk/modules/rofl-market/src/payment.rs
```diff
@@ -158,78 +148,22 @@ impl PaymentMethod for Payment {
...
-                    ForbidReentrancy,
-                )?;
+    fn refund<C: Context>(&self, _ctx: &C, instance: &Instance) -> Result<(), Error> {
+        let refund_address = instance.admin;
+        let payment_address = Address::from_eth(&instance.payment_address);
 
-                <C::Runtime as Runtime>::Core::use_tx_gas(result.gas_used)?;
+        // Determine refund amount.
+        let amount =
+            <C::Runtime as Runtime>::Accounts::get_balance(payment_address, Denomination::NATIVE)?;
```
kostko: You can't just use the "native" denomination always as the payment method defines the denomination (if the payment method is "native").

runtime-sdk/modules/rofl-market/src/payment.rs
peternose: Fixed. Now, the contract handles refunds, but doesn't receive the address of the last depositor. This will break already deployed contracts, but I think no-one is using this currently.

runtime-sdk/modules/rofl-market/src/payment.rs
```diff
@@ -52,12 +52,10 @@ const EVM_CONTRACT_PAY: solabi::FunctionEncoder<
 > = solabi::FunctionEncoder::new(solabi::selector!("rmpPay(uint8,uint64,address,bytes)"));
 
 /// Type of the method invoked for the `refund` action for `Payment::EvmContract`.
-/// rmpRefund(to, data)
+/// rmpRefund(data)
```
kostko: Should we also pass the instance ID? Also in other methods.

peternose: Added `instanceId` to all 3 methods.

kostko (approved):

PR #2465, opened 2026-05-25
rofl-scheduler: Per-offer allowed_creators and allowed_artifacts

anzoman:
Adds per-offer overrides for `allowed_creators` and `allowed_artifacts`, making it possible to mix public and internal workloads on the same scheduler instance without repeating the global allowlist for every offer.

## What changed

- `offers` is now a `BTreeMap<String, OfferConfig>` internally. Each entry carries optional `allowed_creators` and `allowed_artifacts` overrides; when absent, the top-level value is used as a fallback. Empty list = allow all.
- `offer_key_by_id` is built once per scheduling round in `discover()` and reused in `process_pending()`, avoiding a redundant on-chain fetch.
- **Fully backwards compatible** — the existing plain-string array format is still accepted. New entries can optionally carry per-offer config using a map with an `id` field.

## Config

```yaml
# Global fallback — applies to all offers that don't override it
allowed_creators: []

offers:
  - playground_short            # plain string — global defaults apply
  - playground_short_sgx
  - id: oasis_internal          # map entry — per-offer overrides
    allowed_creators:
      - "oasis1abc..."
```

### DIFF
--- a/rofl-scheduler/src/config.rs
+++ b/rofl-scheduler/src/config.rs
@@ -20,8 +20,21 @@ use oasis_runtime_sdk_rofl_market as market;
 const ROFL_SCHEDULER_CONFIG_KEY: &str = "rofl_scheduler";
 
 /// Raw per-offer configuration as serialized.
-#[derive(Clone, Debug, Default, cbor::Decode)]
+///
+/// Each element of the `offers` sequence is either a plain string (old format, no overrides) or
+/// a map with an `id` field and optional per-offer config overrides:
+///
+/// ```yaml
+/// offers:
+///   - public               # plain string — global defaults apply
+///   - id: internal         # map entry — per-offer overrides
+///     allowed_creators:
+///       - "oasis1..."
+/// ```
+#[derive(Clone, Debug, Default)]
 pub struct RawOfferConfig {
+    /// Offer identifier (value of the `net.oasis.scheduler.offer` metadata key).
+    pub id: String,
     /// Allowed instance creator addresses for this offer.
     ///
     /// When set, overrides the global `allowed_creators`. An empty list means all creators are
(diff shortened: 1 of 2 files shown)


rofl-scheduler/src/config.rs
```diff
@@ -19,15 +19,79 @@ use oasis_runtime_sdk_rofl_market as market;
...
+///
+/// Each element of the sequence is either a plain string (old format, no overrides) or a
+/// single-key map whose key is the offer name and whose value is a [`RawOfferConfig`]:
+///
+/// ```yaml
+/// offers:
+///   - playground_short             # plain string — global defaults apply
+///   - oasis_internal:              # map entry — per-offer overrides
+///       allowed_creators:
+///         - "oasis1..."
```
matevz: This is probably more in line with what we have for ROFLs and runtimes config.

```suggestion
///   - id: oasis_internal              # map entry — per-offer overrides
///     allowed_creators:
///       - "oasis1..."
```

anzoman: Good point, implemented.

rofl-scheduler/src/config.rs
```diff
@@ -19,15 +19,90 @@ use oasis_runtime_sdk_rofl_market as market;
...
+    /// Allowed artifact hashes for this offer.
+    ///
+    /// When set, overrides the global `allowed_artifacts` entirely. When not set (`None`), the
+    /// global `allowed_artifacts` is used as a fallback.
+    pub allowed_artifacts: Option<BTreeMap<String, Vec<String>>>,
+}
+
+/// A map entry with an explicit `id` field and optional per-offer config overrides.
+#[derive(Clone, Debug, Default, cbor::Decode)]
+struct RawOfferEntry {
```
kostko: Why have both `RawOfferEntry` and `RawOfferConfig`? Wouldn't it be possible to just have `RawOfferConfig` with `id` included and then implement `cbor::Decode` directly for `RawOfferConfig` (supporting both string and map)?

And then in `RawLocalConfig` could just be `Vec<RawOfferConfig>` directly instead of having this extra `RawOffersField`?

anzoman: Makes sense, I have now simplified the offers to be of type `Vec<RawOfferConfig>`.

rofl-scheduler/src/config.rs
```diff
@@ -229,23 +368,38 @@ impl LocalConfig {
...
+        let artifacts = self
+            .offers
+            .get(offer_key)
+            .and_then(|cfg| cfg.allowed_artifacts.as_ref())
+            .unwrap_or(&self.allowed_artifacts);
 
-        if !allowed_hashes.contains(hash) {
-            return Err(anyhow!("{kind} artifact not allowed"));
+        match artifacts.get(kind) {
+            None => Ok(()), // all artifacts of this kind are allowed
```
kostko:
```suggestion
            None => Ok(()), // All artifacts of this kind are allowed.
```

kostko (commented):
Looks good, have you tested it anywhere?

rofl-scheduler/src/config.rs
anzoman: Addressed.
I have tested the changes on my local ROFL node.

kostko (approved):

PR #2470, opened 2026-06-02
rofl-scheduler: Release 0.8.0

anzoman:
Let's release a new minor version of ROFL Scheduler that supports per-offer `allowed_creators` and `allowed_artifacts`.


matevz (approved):
I verified the enclave IDs locally and they match.

kostko (approved):

PR #2471, opened 2026-06-04
runtime-sdk/modules/evm: Increase create_contract_limit to 64k

matevz:
This PR increases `create_contract_limit` to 64 KiB and `max_initcode_size` to 128 KiB. These values are defined in EIP-7907 and EIP-3860 respectively.

No new gas charging mechanism was introduced in this PR (EIP-7907 proposes 2 gas units per 32 bytes exceeding 24 KiB) and init code (EIP-3860 proposes 2 gas units per 32 bytes of init code).

Fixes https://github.com/oasisprotocol/sapphire-paratime/issues/712

I also tried a fork of https://github.com/oasisprotocol/privana/pull/153 and successfully deployed a 26k big production Accounting contract.

### DIFF
--- a/runtime-sdk/modules/evm/src/lib.rs
+++ b/runtime-sdk/modules/evm/src/lib.rs
@@ -77,6 +77,12 @@ pub trait Config: 'static {
     /// Maximum result size in bytes.
     const MAX_RESULT_SIZE: usize = 1024;
 
+    /// Maximum contract code size in bytes.
+    const CREATE_CONTRACT_LIMIT: usize = 0x10000; // 64 KiB (see EIP-7907).
+
+    /// Maximum size of the init code in bytes.
+    const MAX_INITCODE_SIZE: usize = 0x20000; // 2*CREATE_CONTRACT_LIMIT (see EIP-3860).
+
     /// Maps an Ethereum address into an SDK account address.
     fn map_address(address: primitive_types::H160) -> Address {
         Address::new(
@@ -117,7 +123,11 @@ pub trait Config: 'static {
                 }
             })
         } else {
-            EVM_CONFIG.get_or_init(EVMConfig::shanghai)
+            EVM_CONFIG.get_or_init(|| EVMConfig {
+                create_contract_limit: Some(Self::CREATE_CONTRACT_LIMIT),
+                max_initcode_size: Some(Self::MAX_INITCODE_SIZE),
+                ..EVMConfig::shanghai()
+            })
         }
     }
 }
--- a/runtime-sdk/modules/evm/src/test.rs
+++ b/runtime-sdk/modules/evm/src/test.rs
@@ -78,6 +78,51 @@ fn load_erc20() -> Vec<u8> {
     .expect("compiled ERC20 contract should be a valid hex string")
 }
 
+/// Constructs an EVM init code that, once executed, deploys `code_len` bytes of runtime code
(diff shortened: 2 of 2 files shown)


runtime-sdk/modules/evm/src/lib.rs
```diff
@@ -117,7 +117,11 @@ pub trait Config: 'static {
                 }
             })
         } else {
-            EVM_CONFIG.get_or_init(EVMConfig::shanghai)
+            EVM_CONFIG.get_or_init(|| EVMConfig {
+                create_contract_limit: Some(0x10000),
```
peternose: You should put `0x10000` in a constant, so that we know that this is `64 KiB`.

peternose (approved):

kostko (approved):

runtime-sdk/modules/evm/src/test.rs
```diff
@@ -1006,6 +1056,105 @@ fn test_fee_refunds() {
...
 
+#[test]
+fn test_large_contract_deploy() {
+    let mut mock = mock::Mock::default();
+    let ctx = mock.create_ctx_for_runtime::<EVMRuntime<EVMConfig>>(true);
+    let mut signer = EvmSigner::new(0, keys::dave::sigspec());
+
+    EVMRuntime::<EVMConfig>::migrate(&ctx);
+
+    // Try 32 KiB code size. Should pass since `create_contract_limit` is greater than this.
```
peternose: This test should now rely on constants only, and should never mention 32 KiB neither 64 KiB. So just use `CREATE_CONTRACT_LIMIT / 2` or `CREATE_CONTRACT_LIMIT` or `CREATE_CONTRACT_LIMIT - 1`.

runtime-sdk/modules/evm/src/test.rs
```diff
@@ -1006,6 +1056,105 @@ fn test_fee_refunds() {
...
+    let code_len = 0x8000;
```
peternose: Should be computed from the constant, so that if we change the constant (e.g. to 1KiB), this test still passes.

runtime-sdk/modules/evm/src/test.rs
```diff
@@ -1006,6 +1056,105 @@ fn test_fee_refunds() {
...
+    // Try 65 KiB total init size. Should pass since `max_initcode_size` is greater than this.
```
peternose: Remove fixed values in the comments.

runtime-sdk/modules/evm/src/test.rs
```diff
@@ -1006,6 +1056,105 @@ fn test_fee_refunds() {
...
+        } else {
+            panic!(
+                "deploy exceeding the {} should fail: {:?}",
+                test_name, dispatch_result.result
+            );
+        }
+    };
+
+    // 64 KiB + 1 byte code size. Should fail since `create_contract_limit` is only 0x10000 (64 KiB).
+    test_deploy_failure(0x10000 + 1, 0, "contract size limit");
```
peternose: Use constant now.

peternose:
Just checking if the EVM, which is still in review, supports setting this limit also?

peternose (approved):

kostko:
> Just checking if the EVM, which is still in review, supports setting this limit also?

Yes, should be doable in a similar way (via config).

PR #2478, opened 2026-06-12
rofl-appd: Add upsert and delete metadata endpoint

### LINKED ISSUE #2405: rofl/appd: Implement add metadata method
Currently, appd metadata only supports `set` and `get`. The `set` method is declarative and replaces the existing metadata state.

For use cases where multiple containers/applications in the same deployment need to update metadata, an `add` (or similar) method would be useful. This method should only add new keys or override existing ones, without resetting the rest of the metadata.

matevz:
Implements #2405


matevz:
Tests seem to be failing due to regression in time `0.3.48` crate released today. :angry: 

Discussion: https://github.com/time-rs/time/issues/783

rofl-appd/src/services/metadata.rs
```diff
@@ -90,10 +99,63 @@ impl<A: App> MetadataService for OasisMetadataService<A> {
...
+    async fn upsert(&self, metadata: BTreeMap<String, String>) -> Result<(), Error> {
+        let mut map = self.metadata.write().await;
+
+        let mut updated = map.clone();
+        let mut changed = false;
+        for (key, value) in metadata {
+            if updated.get(&key) != Some(&value) {
+                changed = true;
+            }
+            updated.insert(key, value);
```
kostko: Since `insert` returns the old value you could do this with a single operation, something like:
```rust
if updated.insert(key, value) != Some(&value) {
    changed = true;
}
```

matevz: The problem is that `insert()` consumes the value so I'd need to clone it for `Some(&value)` to work afterwards. I moved `update.insert()` inside `if` block now to save some cycles.

matevz:
> Tests seem to be failing due to regression in time `0.3.48` crate released today. 😠
> 
> Discussion: [time-rs/time#783](https://github.com/time-rs/time/issues/783)

Time crate `0.3.51` compiles fine. Unblocking.

matevz:
I tested the new PUT and DELETE methods and work as expected.

kostko (approved):

PR #2479, opened 2026-06-19
fix(rofl-scheduler): derive non-empty domain verification token

PrazwalR:
Fixes a custom-domain ownership-verification bypass in the ROFL scheduler proxy.

`domain_verification_token` (`rofl-scheduler/src/types.rs`) finalized a TupleHash
into an empty `Vec`, producing an empty token for all inputs. The proxy's DNS TXT
verification (`proxy/domain.rs`) therefore stopped binding a domain to a specific
instance, enabling cross-tenant custom-domain takeover and TLS MITM (valid ACME
cert for a hijacked domain).

Finalize the digest into a 32-byte buffer (256-bit, matching `keymanager.rs`).

Added 8 regression tests (shape, determinism, binding to each tuple field, no
substring collisions, independent-derivation anti-stub check). Verified on
linux/amd64 — `cargo test -p rofl-scheduler domain_verification_token` → 8 passed.
The crate does not build on macOS/arm64 (deoxysii x86 asm); use CI / a linux/amd64
container. rustfmt clean.

### DIFF
--- a/rofl-scheduler/src/types.rs
+++ b/rofl-scheduler/src/types.rs
@@ -87,7 +87,7 @@ pub fn domain_verification_token(
     hasher.update(instance.id.as_ref());
     hasher.update(domain.as_bytes());
 
-    let mut output = Vec::new();
+    let mut output = [0u8; 32]; // TupleHash::v256 → 256-bit token (matches keymanager.rs).
     hasher.finalize(&mut output);
     BASE64_STANDARD.encode(output)
 }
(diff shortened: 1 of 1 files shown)


PrazwalR:
The 4 failing checks (e2e-ts-web-core, e2e-ts-web-rt, e2e-rt, benchmark-rt) all fail at ./download-artifacts.sh with: Need Buildkite artifacts, but BUILDKITE_API_TOKEN environment variable is not set. This is the fork-PR secret limitation: GitHub does not expose secrets.BUILDKITE_API_TOKEN to PRs from forks, so the artifact download step exits before any test runs. It is unrelated to this change (a one-line fix in rofl-scheduler). The checks that exercise this change pass: test-rust, test-rust-sgx, lint-rust, ci-audit, and codecov/patch (100% of the diff covered). Could a maintainer re-run the e2e jobs with secrets, or confirm they are expected to fail on fork PRs? Thanks.

peternose (approved):

peternose:
I dropped the comment and will merge your changes in the following PR:
- https://github.com/oasisprotocol/oasis-sdk/pull/2483

Thank you for your contribution.

Issue #2248, opened 2025-06-24
Support non-encrypted variables in ROFL manifest yaml


matevz:
Currently, there are two ways to configure your ROFL:
1. define secrets in rofl.yaml which are fed to your docker image
2. define publicly visible variables in `compose.yaml`, `Dockerfile`, or the image itself

Step 2. will produce a different enclave ID for different configuration every build which is annoying since it's still the same (audited) app. Add support for defining publicly exposed variables to `rofl.yaml`. They should work exactly the same as secrets, just that they are unencrypted.

kostko:
We could simply make it so that any metadata prefixed with `env.` are set as environment variables (e.g. `env.MY_VAR`) with `rofl-containers` doing the interpretation.

kostko:
Of course it is deployment-specific, what do you mean?

martintomazic:
This functionality would be nice to implement as using pattern **2.** forces a new app release every time a config changes (assuming strict release policy). e.g. - see https://github.com/oasisprotocol/Honoroll/issues/416.

martintomazic:
>We could simply make it so that any metadata prefixed with env. are set as environment variables (e.g. env.MY_VAR) with rofl-containers doing the interpretation.

## Simple solution

https://github.com/oasisprotocol/oasis-sdk/issues/2248#issuecomment-3009497433
* Pros:
    - Almost trivial.
* Cons:
    1. Services inside docker compose need to interpolate every variable, which becomes boilerplate as the number of vars grow.
    2. From the extremely puristic view, this might be breaking, as environment is now sourced with something that previously might not be (don't think this will affect anyone in practice).


## Alternative Solution

* For every `env.[service_name].*`, we write those vars into `/run/podman/env/[service_name].env`
* Optionally we would have `common_env.*`, writting shared confing into `/run/podman/env/common.env`.

Then the app developer, in order to enable this convention writes:
```yaml
services:
  my-service:
    image: my-image:latest

    env_file:
      - path: /run/podman/env/common.env
        required: false
      - path: /run/podman/env/my-service.env
        required: false
```
Explicit acknowledge of the convention also mitigates _Simple_Solution:Cons:2_. This is also inline with existing [container-secrets](https://docs.oasis.io/build/rofl/features/secrets/#container-secrets) pattern.

[...]

PR #2481, opened 2026-06-28
Support non-encrypted variables in ROFL manifest yaml

martintomazic:
Part of #2248.

## What
Keys that are part of the rofl instance metadata and start with `.env` prefix are passed as part of the env file during rofl containers init, like is currently done for the rofl secrets.

## E2E testing
Created simple tdx instance that printed container secrets, normal secrets, and "public" variable using `.env` metadata convention above. Used local build of rofl containers, and tested on the Sapphire testnet.

## Consideration
Like with secrets we expect the admin to not shoot itself by e.g.:
- Setting values that contains new line character.
- Non posix compliant variables.

## Follow-up
_Alternative solution_ from https://github.com/oasisprotocol/oasis-sdk/issues/2248#issuecomment-4826995846 can be done as follow-up without being breaking, and will be in fact complementary.

### DIFF
--- a/rofl-containers/src/main.rs
+++ b/rofl-containers/src/main.rs
@@ -158,9 +158,25 @@ impl App for ContainersApp {
             }
         }
 
-        // Initialize secrets.
+        // Fetch app config.
+        let app_cfg = match env.client().app_cfg().await {
+            Ok(cfg) => cfg,
+            Err(err) => {
+                slog::error!(logger, "failed to fetch app config"; "err" => ?err);
+                process::abort();
+            }
+        };
+
+        // Initialize environment variables from deployment metadata (env.* keys).
+        slog::info!(logger, "initializing container environment variables");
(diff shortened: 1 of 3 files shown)


rofl-containers/src/main.rs
```diff
@@ -158,9 +158,28 @@ impl App for ContainersApp {
...
+            }
+        };
+
+        // Initialize environment variables from deployment metadata (env.* keys).
+        slog::info!(logger, "initializing container environment variables");
+        for (key, value) in app_cfg.metadata {
+            if let Some(name) = key.strip_prefix("env.") {
+                let name_upper = name.to_uppercase().replace(' ', "_");
+                containers::env().set(&name_upper, &value);
+                slog::info!(logger, "provisioned environment variable"; "key" => key);
```
martintomazic: NIT: I would maybe allow only alphanumerics for the key? 

Probably we can log config value here as well.

kostko (approved):

PR #2483, opened 2026-06-30
fix(rofl-scheduler): derive non-empty domain verification token

peternose:


rofl-scheduler/src/types.rs
```diff
@@ -87,7 +87,7 @@ pub fn domain_verification_token(
     hasher.update(instance.id.as_ref());
     hasher.update(domain.as_bytes());
 
-    let mut output = Vec::new();
+    let mut output = [0u8; 32]; // TupleHash::v256 → 256-bit token (matches keymanager.rs).
```
kostko: Drop this comment.

kostko (approved):

Issue #2475, opened 2026-06-10
Move `allowed_creators` and `allowed_artifacts` to Offers metadata on-chain


matevz:
Followup to https://github.com/oasisprotocol/oasis-sdk/pull/2465 let's extend the ROFL-node config level configuration on-chain so there is no need to restart machine each time and that we also hide private offers upfront from users:
- put `allowed_creators` and `allowed_artifacts` inside offer metadata
- add CLI support for that
- add private field (is_private?) so these offers are hidden from normal users when listing offers (unless `-a` provided)
- add rofl-scheduler to read these and acknowledge them

PR #2485, opened 2026-07-01
rofl-scheduler: Read offer access policy from on-chain metadata

anzoman:
Moves per-offer `allowed_creators` and `allowed_artifacts` from the scheduler's local node config to on-chain offer metadata, so providers can change access policy with a transaction instead of editing the node config and restarting the machine. The scheduler parses an `OfferPolicy` from each offer's metadata every round (keys `allowed_creators`, `allowed_artifacts.<kind>`, and a `private` listing hint); matching constants are added to the Go SDK for the CLI. The global local-config fields are removed with no fallback.

Implements #2475.

### DIFF
--- a/client-sdk/go/modules/roflmarket/types.go
+++ b/client-sdk/go/modules/roflmarket/types.go
@@ -124,6 +124,29 @@ type Offer struct {
 	Metadata map[string]string `json:"metadata"`
 }
 
+// Well-known offer metadata keys recognized by the ROFL scheduler.
+const (
+	// MetadataKeyOffer is the offer metadata key holding the offer identifier (the value used by
+	// the scheduler's local config to select which offers it serves).
+	MetadataKeyOffer = "net.oasis.scheduler.offer"
+	// MetadataKeyOfferAllowedCreators is the offer metadata key holding a comma-separated list of
+	// allowed instance creator addresses. When unset or empty, all creators are allowed.
+	MetadataKeyOfferAllowedCreators = "net.oasis.scheduler.offer.allowed_creators"
+	// MetadataKeyOfferAllowedArtifactsPrefix is the prefix of the offer metadata keys holding the
+	// allowed artifact hashes. The artifact kind is the suffix following the prefix (e.g.
+	// "net.oasis.scheduler.offer.allowed_artifacts.firmware") and the value is a comma-separated
+	// list of allowed SHA256 hashes. If no key exists for a kind, all artifacts of that kind are
+	// allowed.
+	MetadataKeyOfferAllowedArtifactsPrefix = "net.oasis.scheduler.offer.allowed_artifacts."
+	// MetadataKeyOfferPrivate is the offer metadata key suggesting that
+	// the offer should be hidden from public facing UI.
+	MetadataKeyOfferPrivate = "net.oasis.scheduler.offer.private"
+
+	// MetadataValueTrue is the offer metadata value that enables a boolean flag such as
+	// MetadataKeyOfferPrivate.
+	MetadataValueTrue = "1"
+)
+
 // Term is the reservation term.
 type Term uint8
 
--- a/rofl-scheduler/src/config.rs
+++ b/rofl-scheduler/src/config.rs
@@ -1,100 +1,17 @@
-use std::{
-    collections::{BTreeMap, BTreeSet},
-    sync::Arc,
-};
+use std::{collections::BTreeSet, sync::Arc};
 
 use anyhow::{anyhow, Result};
 use base64::prelude::*;
 
 use oasis_runtime_sdk::{
     cbor,
-    core::{
-        common::crypto::{hash::Hash, signature::PublicKey},
-        Protocol,
-    },
+    core::{common::crypto::signature::PublicKey, Protocol},
     types::address::Address,
 };
 use oasis_runtime_sdk_rofl_market as market;
 
 /// Local configuration key that contains the ROFL scheduler configuration.
 const ROFL_SCHEDULER_CONFIG_KEY: &str = "rofl_scheduler";
-
-/// Raw per-offer configuration as serialized.
-///
-/// Each element of the `offers` sequence is either a plain string (old format, no overrides) or
-/// a map with an `id` field and optional per-offer config overrides:
-///
-/// ```yaml
-/// offers:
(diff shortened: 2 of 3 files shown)


kostko (commented):
This makes sense to me, have you done any tests with an actual scheduler? Before this can be deployed all existing offers need to be updated to use this metadata, right?

rofl-scheduler/src/manager.rs
```diff
@@ -102,6 +117,93 @@ impl InstanceUpdates {
...
+            }
+            let hashes = parse_csv(raw)
+                .filter_map(|item| item.parse::<Hash>().ok())
+                .collect();
+            allowed_artifacts.insert(kind.to_string(), hashes);
+        }
+
+        let private = metadata
+            .get(METADATA_KEY_OFFER_PRIVATE)
+            .map(|v| is_truthy(v))
```
matevz: `v` should be clearly defined by the metadata format and then implemented in CLI accordingly. I'd use `"1"`.

matevz (commented):
Do I remember correctly the current offer is copied over to machine once you deploy your app? And this copied offer is read in the new rofl-scheduler code instead of the latest on-chain offer? So if someone changes offer Metadata afterwards, those won't be reflected in the existing ROFL replica until it's redeployed.

rofl-scheduler/src/manager.rs
anzoman: I added "1" as the single canonical value: the scheduler only enables the flag on an exact "1" match (dropped the loose true/yes parsing), and I added a `MetadataValueTrue = "1"` constant to the Go SDK plus doc comments on both sides so the CLI can use the same value.

rofl-scheduler/src/config.rs
```diff
@@ -1,120 +1,32 @@
...
-        }
-    }
-}
-
 /// Raw local configuration as serialized.
+///
+/// Unknown fields are ignored so that configs still carrying the removed `allowed_creators` and
+/// `allowed_artifacts` keys (now sourced from on-chain offer metadata) continue to load.
 #[derive(Clone, Debug, Default, cbor::Decode)]
+#[cbor(allow_unknown)]
```
matevz: Is node-specific filtering by `allowed_creators` still supported? If not, then this should not be allowed anymore. Also from the docs https://docs.oasis.io/node/run-your-node/rofl-node#limit-rofl-node-to-specific-rofl-creators.

```suggestion
```

client-sdk/go/modules/roflmarket/types.go
```diff
@@ -124,6 +124,31 @@ type Offer struct {
...
+	// MetadataKeyOfferAllowedArtifactsPrefix is the prefix of the offer metadata keys holding the
+	// allowed artifact hashes. The artifact kind is the suffix following the prefix (e.g.
+	// "net.oasis.scheduler.offer.allowed_artifacts.firmware") and the value is a comma-separated
+	// list of allowed SHA256 hashes. If no key exists for a kind, all artifacts of that kind are
+	// allowed.
+	MetadataKeyOfferAllowedArtifactsPrefix = "net.oasis.scheduler.offer.allowed_artifacts."
+	// MetadataKeyOfferPrivate is the offer metadata key marking an offer as private. Private
+	// offers are hidden from normal offer listings unless explicitly requested. The only value
+	// that enables the flag is MetadataValueTrue; any other value (or an absent key) leaves the
+	// offer public.
```
matevz:
```suggestion
	// MetadataKeyOfferPrivate is the offer metadata key suggesting that
	// the offer should be hidden from public facing UI.
```

anzoman: Applied.

rofl-scheduler/src/config.rs
anzoman: Removed; will also remove from docs after we release the new ROFL Scheduler.

rofl-scheduler/src/manager.rs
```diff
@@ -563,15 +666,13 @@ impl Manager {
     ) -> Result<()> {
         let instances = local_state.client.instances().await?;
 
-        // Build a map from offer ID to offer key for offers this scheduler accepts.
-        // `offer_key_by_id` was populated by discover(); no second network call needed.
-        let acceptable_offers: BTreeMap<market::types::OfferId, String> = local_state
+        // Build the set of offer IDs this scheduler accepts. `offer_key_by_id` was populated by
```
matevz, line 690: Leftover?

```suggestion
        // Build the set of offer IDs this scheduler accepts. `offer_policy_by_id` was populated by
```

rofl-scheduler/src/manager.rs
```diff
@@ -102,6 +120,84 @@ impl InstanceUpdates {
...
+
+impl OfferPolicy {
+    /// Parse the offer access policy from on-chain offer metadata. Malformed list entries are
+    /// silently skipped.
+    fn from_metadata(metadata: &BTreeMap<String, String>) -> Self {
+        let allowed_creators = metadata
+            .get(METADATA_KEY_OFFER_ALLOWED_CREATORS)
+            .map(|raw| {
+                parse_csv(raw)
+                    .filter_map(|item| Address::from_bech32(item).ok())
```
matevz: This will ignore any bad-formatted addresses and allow anyone to use the offer. I would at least put a warning to the log that a malformed offer exists.

matevz (approved):
Two nits, otherwise looks good IMO.

rofl-scheduler/src/manager.rs
anzoman: Added a warning log now. Malformed creator addresses (and artifact hashes, for consistency) are skipped as before but logged with the offer ID and the offending value, so a misconfigured offer is visible in the logs.

rofl-scheduler/src/manager.rs
anzoman: Not a leftover — this block iterates `offer_key_by_id` (not `offer_policy_by_id`): it needs the `net.oasis.scheduler.offer` metadata-key string to match against `cfg.offers`, and that string only lives in `offer_key_by_id`. offer_policy_by_id stores the parsed `OfferPolicy`, which doesn't carry the key. So the comment is accurate as written. Kept it as-is.

kostko (approved):

PR #2487, opened 2026-07-15
docs/rofl/features: Document public variables

### LINKED ISSUE #2486: Docs: Document rofl public variables
If we want to follow secrets section pattern, we need to first release cli reference and then reference it here: https://github.com/oasisprotocol/cli/pull/705

martintomazic:
Closes https://github.com/oasisprotocol/oasis-sdk/issues/2486.

Followed the style of secrets 1:1 (instructed codex to generate text for me and reviewed myself).

E.g. see https://explorer.oasis.io/testnet/sapphire/tx/0c788cbf3be706f6641e6d475d5af126c482e05e6c7dcd5e05e0930364a03db7 for real deployment.

### DIFF
--- a/docs/rofl/features/public_variables.md
+++ b/docs/rofl/features/public_variables.md
@@ -1,0 +1,53 @@
+# Public Variables
+
+Sometimes containers need access to non-sensitive configuration, API endpoints,
+contract addresses, feature flags, and other deliberately exposed information.
+Such data can be passed to containers running in ROFL via _public variables_.
+Public variables are arbitrary key-value pairs that are exposed to containers
+as environment variables. For confidential values, use [secrets] instead.
+
+Public variables can be easily managed via the Oasis CLI, for example to create
+a public variable called `API_URL` you can use:
+
+```sh
+echo -n "https://api.example.com" | oasis rofl public-var set API_URL -
+```
+
+Note that this only updates the local app manifest file, but the public variable
+is not propagated to the app just yet. This allows you to easily configure as
+many public variables as you want without the need to constantly update the
+on-chain app configuration.
+
+Updating the on-chain configuration can be performed via the usual `update`
+command as follows:
+
+```sh
+oasis rofl update
+```
+
+:::info Detailed CLI Reference
+
+For comprehensive documentation on public variable management commands including
+importing from `.env` files, removing public variables, and other advanced
+features, consult the [Oasis CLI] documentation.
+
+:::
+
+Inside containers public variables can be passed via environment variables.
+
+## Environment Variables
+
+Each public variable is automatically exposed in the Compose environment and can
+be trivially used in the Compose file.
+
+```yaml
+services:
+  test:
+    image: docker.io/library/alpine:3.21.2@sha256:f3240395711384fc3c07daa46cbc8d73aa5ba25ad1deb97424992760f8cb2b94
+    command: echo "API URL is $API_URL"
+    environment:
+      - API_URL=${API_URL}
+```
+
+[secrets]: ./secrets.md
+[Oasis CLI]: https://github.com/oasisprotocol/cli/blob/master/docs/rofl.md#public-var
--- a/docs/rofl/features/secrets.md
+++ b/docs/rofl/features/secrets.md
@@ -4,7 +4,7 @@ Sometimes containers need access to data that should not be disclosed publicly,
 for example API keys to access certain services. This data can be passed to
 containers running in ROFL via _secrets_. Secrets are arbitrary key-value pairs
 which are end-to-end encrypted so that they can only be decrypted inside a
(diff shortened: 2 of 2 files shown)


docs/rofl/features/public_var.md
```diff
@@ -0,0 +1,63 @@
+# Public Variables
+
+Sometimes containers need access to non-sensitive configuration, for example
+API endpoints or public feature flags. This data can be passed to containers
```
matevz:
```suggestion
API endpoints, contract addresses, feature flags, and other deliberately exposed information.
Such data can be passed to containers
```

docs/rofl/features/public_var.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+:::info Detailed CLI Reference
+
+For comprehensive documentation on public variable management commands including
+importing from `.env` files, removing public variables, and other advanced
+features, consult the [Oasis CLI] documentation.
+
+:::
+
```
matevz: I'd move this info box somewhere at the end. Do the same in the Secrets chapter.

docs/rofl/features/public_var.md
```diff
@@ -0,0 +1,63 @@
...
+Note that this only updates the local app manifest file, but the public variable
+is not propagated to the app just yet. This allows you to easily configure as
+many public variables as you want without the need to constantly update the
+on-chain app configuration.
+
+:::info
+
+Public variables are stored unencrypted in the local app manifest and in
+on-chain app metadata. This means that they can be read by anyone who can access
+the manifest or inspect the on-chain app configuration.
```
matevz:
```suggestion
on-chain app metadata. Anyone who can access the manifest or inspect the on-chain app
configuration can read them.
```

docs/rofl/features/public_var.md
```diff
@@ -0,0 +1,63 @@
...
+features, consult the [Oasis CLI] documentation.
+
+:::
+
+Note that this only updates the local app manifest file, but the public variable
+is not propagated to the app just yet. This allows you to easily configure as
+many public variables as you want without the need to constantly update the
+on-chain app configuration.
+
+:::info
```
matevz:
```suggestion
:::info Public variables vs Secrets
```

Also update Secrets chapter to include reference to Public variables chapter.

docs/rofl/features/public_variables.md
```diff
@@ -0,0 +1,63 @@
...
+
+```yaml
+services:
+  test:
+    image: docker.io/library/alpine:3.21.2@sha256:f3240395711384fc3c07daa46cbc8d73aa5ba25ad1deb97424992760f8cb2b94
+    command: echo "API URL is $API_URL"
+    environment:
+      - API_URL=${API_URL}
+```
+
```
matevz, line 51: Maybe add some alternatives such as Compose file and `.env`, for example:

````suggestion

## Public variables vs Compose file and `.env`

Apart from public variables ROFL developers can expose environment variables directly in the Compose file as literals. For example:

```yaml
services:
  test:
    image: docker.io/library/alpine:3.21.2@sha256:f3240395711384fc3c07daa46cbc8d73aa5ba25ad1deb97424992760f8cb2b94
    command: echo "API URL is $API_URL"
    environment:
      - API_URL="https://api.example.com/v1"
```

The downside of this approach is that each time the literal is updated, the hash of the Compose file changes and thus the ROFL needs to be rebuilt and redeployed.

Another approach is storing public environment variables inside the `.env` file. In this case, users can investigate the container image to examine environment variables. This approach comes handy during development, because you can feed the same `.env` file locally and then inside production ROFL. However, similarly to the approach above, the container hash will change each time you change the `.env` file and consequently the container hash changes in your Compose file, which requires you to rebuild and redeploy ROFL.
````

docs/rofl/features/public_var.md
martintomazic: Removed this entirely as this is repeating the obvious once more, secrets/public var are references as alternative at the top of the chapter in both files.

docs/rofl/features/public_variables.md
martintomazic: Does this add value?
1. Compose file literals probably one can infer from the existing compose syntax.
2. Alternative doesn't have much to do with public variables, as one can bake arbitrary stuff into docker image, doesn't have to be .env, can be `config.toml` also. 

>This approach comes handy during development, because you can feed the same `.env` file locally and then inside production ROFL.

you can do the same by adding explicit --env-file flag to compose command.

matevz (approved):

docs/rofl/features/public_variables.md
matevz, line 1: Maybe rename to `public_variables.md` for consistency? The chapters here are named more SEO-like and not what CLI names the commands.

PR #2514, opened 2026-08-26
rofl-proxy: Log exact cert-provider errors

matevz:
Currently, only a generic `unexpected order status: Invalid` is reported when there are issues with DNS setting, service availability or the client cache. This PR logs exact reason(s) why certificate order failed.


anzoman (approved):
Nice to have some more extensive logs.

kostko (approved):

